PR #7, opened 2026-03-02
fix: auto-install JS parser npm deps on first run

### LINKED ISSUE #6: JS/TS parser fails on fresh install: missing npm dependencies
## Description

When scanning a JavaScript or TypeScript repository after a fresh install of OpenAnt, the `typescript_analyzer` stage fails with:

```
Error: Cannot find module 'ts-morph'
Require stack:
- /tmp/OpenAnt/libs/openant-core/parsers/javascript/typescript_analyzer.js
```

## Root Cause

The JS parser at `libs/openant-core/parsers/javascript/` has a `package.json` declaring dependencies (`ts-morph`, `typescript`, `@typescript/vfs`), but `npm install` is never run automatically. The Go CLI's `EnsureRuntime()` bootstraps Python (creates venv, runs `pip install -e`) but has no equivalent step for npm.

## Reproduction

```bash
# Fresh install
git clone https://github.com/knostic/OpenAnt.git
cd OpenAnt/apps/openant-cli && make build
./bin/openant set-api-key <key>
./bin/openant init /path/to/typescript-project -l javascript
./bin/openant parse
# → typescript_analyzer fails with MODULE_NOT_FOUND
```

## Workaround

```bash
cd libs/openant-core/parsers/javascript && npm install
```

## Suggested Fix

Add an `ensureNodeDeps()` check to `EnsureRuntime()` in `apps/openant-cli/internal/python/runtime.go` that:
1. Checks for `parsers/javascript/node_modules/ts-morph` as a fast-path marker
2. Runs `npm install --prefix <parser-dir>` if missing
3. Treats failure as non-fatal (warning only) so non-JS workflows are unaffected

This mirrors the existing Python auto-install pattern. I have a fix ready and will open a PR.

## Environment

- OpenAnt: commit 1d9c8ca (master)
- macOS 15.5 / darwin arm64
- Node.js v25.6.1
- Go 1.26.0

cbchhaya:
## Summary

Fixes #6 — the JavaScript/TypeScript parser fails on fresh installs because `npm install` is never run for `parsers/javascript/`.

- Adds `ensureNodeDeps()` to `EnsureRuntime()` in the Go CLI, mirroring the existing Python auto-install pattern
- Uses `node_modules/ts-morph` as a fast-path existence check (zero overhead on subsequent runs)
- Non-fatal: prints a warning if npm is unavailable, so Python/Go/C workflows are unaffected

## Context

Found while running OpenAnt against a TypeScript codebase. The `typescript_analyzer.js` requires `ts-morph` (declared in `parsers/javascript/package.json`), but unlike the Python deps which are auto-installed via `pip install -e` during venv setup, the npm deps had no equivalent bootstrap step.

## Test plan

- [x] `go build` and `go vet` pass
- [x] Fresh run (deleted `node_modules/`): prints "Installing JS parser dependencies..." and succeeds
- [x] Second run: skips install silently (fast path)
- [x] Verified on a 57-file TypeScript project — full parse pipeline completes

### DIFF
--- a/apps/openant-cli/internal/python/runtime.go
+++ b/apps/openant-cli/internal/python/runtime.go
@@ -174,8 +174,9 @@ func CheckOpenantInstalled(pythonPath string) error {
 }
 
 // EnsureRuntime is a convenience that detects a runtime, ensures openant
-// is installed (creating a venv if necessary), and returns the final
-// RuntimeInfo pointing to the correct Python binary.
+// is installed (creating a venv if necessary), installs JS parser npm
+// dependencies if needed, and returns the final RuntimeInfo pointing to
+// the correct Python binary.
 func EnsureRuntime() (*RuntimeInfo, error) {
 	rt, err := DetectRuntime()
 	if err != nil {
@@ -184,6 +185,12 @@ func EnsureRuntime() (*RuntimeInfo, error) {
 
 	if err := CheckOpenantInstalled(rt.Path); err != nil {
 		return nil, err
+	}
+
+	// Ensure JS parser npm dependencies are installed.
+	if err := ensureNodeDeps(); err != nil {
+		// Non-fatal: only affects JavaScript/TypeScript scanning.
+		fmt.Fprintf(os.Stderr, "Warning: JS parser deps not installed: %v\n", err)
 	}
 
 	// After CheckOpenantInstalled, the venv may have been created.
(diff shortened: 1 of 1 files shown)


ar7casper:
Thanks for catching this @cbchhaya, and
  apologies for closing the underlying issue prematurely back in March — the bug is real and
  Josh hit it independently while working through his fork contributions in #16 (item 21). His
   #37 addresses the same problem with a few extras: lazy invocation (only runs when JS
  parsing actually happens, so Python/Go-only users don't need npm),
  `node_modules/.package-lock.json` as the completion sentinel (catches Ctrl+C-interrupted
  installs), a cross-platform file lock for concurrent invocations, fail-fast errors with a
  reproduction command, and unit tests. We're going to merge that one. Closing this PR as
  superseded — really appreciate you originally surfacing the bug.

PR #34, opened 2026-05-04
feat: add SARIF 2.1.0 output (-f sarif)

ChrisJr404:
Adds `-f sarif` to `openant report`. Emits a SARIF 2.1.0 log so a scan can be uploaded to GitHub Code Scanning or GitLab SAST without writing a converter, which is the format every other SAST in this category supports.

Implementation mirrors `-f html`: Python's `report-data` returns the same pre-computed JSON we already use for HTML, and the SARIF emitter is purely Go-side (no LLM round-trip, deterministic). Findings become SARIF results, verdicts get hoisted into a synthesized `rules` list. `vulnerable` and `bypassable` surface as level `error`, `inconclusive` and `unclear` as `warning`, everything else as `note`.

Couple of choices worth calling out:

- No `startLine` in the location region. The current `Finding` struct only carries file + function, so I omit the region rather than emit a synthetic line 1 (which would anchor alerts to the wrong row in Code Scanning). When line numbers eventually land in `ReportData`, the region payload in `sarifLocationFor` is the only thing that needs to grow.
- Each result carries a `partialFingerprints` entry keyed `openant/file/function/verdict/v1`, so successive scans dedupe cleanly in Code Scanning instead of opening fresh alerts each run.
- Rules are synthesized per verdict since findings aren't keyed by a stable per-rule taxonomy yet. Categories from `ReportData.Categories` supply the descriptions, falling back to a generic blurb if a verdict has no category entry.
- `versionControlProvenance` is emitted when `ReportData.RepoURL` is set, including `revisionId` from `CommitSHA` if available. Skipped entirely otherwise so we don't write a half-empty object.

Default output path is `<resultsDir>/final-reports/report.sarif`, matching the HTML default's directory layout.

15 unit tests in `internal/report/sarif_test.go` cover envelope shape, rule dedup, level mapping, path normalization (`./` strip + backslash conversion), logical location, dynamic test property propagation, fingerprint stability, VCS provenance gating, the empty-AttackVector fallback that keeps `message.text` non-empty per spec, the 4 KiB message cap, and an end-to-end round-trip through `json.Unmarshal` to confirm the emitted log is valid JSON.

`go test ./...` is green; `gofmt` and `go vet` are clean.

Open to a different flag shape if you'd rather. Could be a flag on `scan` so it lands alongside `pipeline_output.json` automatically, or the rule synthesis can be split into something more granular if per-verdict rules feel too coarse for Code Scanning's per-rule severity controls.

### DIFF
--- a/apps/openant-cli/cmd/report.go
+++ b/apps/openant-cli/cmd/report.go
@@ -26,6 +26,7 @@ Formats:
   summary      Narrative security overview (uses LLM)
   html         Interactive HTML report with charts and filters
   csv          Spreadsheet export of all findings
+  sarif        SARIF 2.1.0 log for GitHub Code Scanning / GitLab SAST upload
 
 If no results path is given, the active project's results_verified.json is used.
 Python owns default output paths — you only need -o to override.
@@ -51,7 +52,7 @@ var (
 func init() {
 	reportCmd.Flags().StringVarP(&reportOutput, "output", "o", "", "Output path (default: derived from format)")
 	reportCmd.Flags().StringVar(&reportDataset, "dataset", "", "Path to dataset JSON (for html/csv)")
-	reportCmd.Flags().StringVarP(&reportFormat, "format", "f", "", "Report format: disclosure, summary, html, csv")
+	reportCmd.Flags().StringVarP(&reportFormat, "format", "f", "", "Report format: disclosure, summary, html, csv, sarif")
 	reportCmd.Flags().StringVar(&reportPipelineOutput, "pipeline-output", "", "Path to pipeline_output.json (for summary/disclosure)")
 	reportCmd.Flags().StringVar(&reportRepoName, "repo-name", "", "Repository name (used when auto-building pipeline_output)")
 	reportCmd.Flags().StringVar(&reportExtraDest, "copy-to", "", "Copy reports to an additional location")
@@ -215,6 +216,31 @@ func runReport(cmd *cobra.Command, args []string) {
 				output.PrintReportSummary(data)
 			}
 			allResults = append(allResults, data)
+		} else if fmt == "sarif" {
+			// SARIF reports use the Go renderer for the same reason HTML
+			// does: it's a deterministic data transformation, not an
+			// LLM-generated narrative, so there's no need to round-trip
+			// through Python.
+			outputPath := reportOutput
+			if outputPath == "" {
+				resultsDir := filepath.Dir(resultsPath)
+				outputPath = filepath.Join(resultsDir, "final-reports", "report.sarif")
+			}
+
+			if err := runSARIFReport(rt, resultsPath, outputPath); err != nil {
+				output.PrintError("sarif: " + err.Error())
+				exitCode = 2
+				continue
+			}
+
+			data := map[string]any{
+				"output_path": outputPath,
+				"format":      "sarif",
+			}
+			if !jsonOutput {
+				output.PrintReportSummary(data)
+			}
+			allResults = append(allResults, data)
 		} else {
 			// Other formats delegate to Python
 			pyArgs := buildReportArgs(resultsPath, fmt)
@@ -264,6 +290,7 @@ func promptFormats() ([]string, error) {
 					huh.NewOption("Summary — narrative security overview written by AI ($)", "summary"),
 					huh.NewOption("HTML — interactive report with charts and filters", "html"),
 					huh.NewOption("CSV — spreadsheet export of all findings", "csv"),
+					huh.NewOption("SARIF — GitHub Code Scanning / GitLab SAST upload", "sarif"),
 				).
 				Value(&selected),
 		),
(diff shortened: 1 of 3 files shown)


gadievron:
Thanks for this @ChrisJr404 — a SARIF 2.1.0 emitter is a genuinely useful addition and the implementation is clean and well-tested. One correctness fix worth folding in before merge (happy to open it as a commit against your branch if you'd prefer):

**`error` verdict is silently filtered from Code Scanning.** In `sarifLevelForVerdict`, an `error` verdict falls through `default → "note"`, which GitHub Code Scanning hides from the default view. But `error` is a real producer verdict in **`DISCLOSURE_ELIGIBLE`** (`core/verdict_taxonomy.py`, "surfaced so it stays on the manual-triage radar rather than vanishing"), so an errored, disclosure-eligible unit disappears from SARIF triage. It should map to `warning` (visible, without over-claiming exploitability — that's reserved for `error`-level = vulnerable/bypassable). Master's own HTML report already surfaces errored units for manual review, so this just brings SARIF to parity.

Also: the `"unclear"` arm is dead — `unclear` isn't a producer verdict anywhere in `core/` (only a local variable name in the Go output formatter). A drift token would fall to `default → note` like any other; the special-case was misleading.

**Reachability note (why `error`-only and not `confirmed`/`agreed`/`unverified`):** the SARIF sink reads `result['finding']` (via `cli.py` report-data), whose namespace is closed to `{vulnerable, bypassable, inconclusive, safe, protected, error}` (`finding_verifier.py` `correct_finding` enum). The stage-2 verdicts live in a separate `stage2_verdict` field the SARIF path never reads — so adding arms for them would be dead code.

**Verified:** RED→GREEN unit test on the full mapping table; a real emitted+validated `.sarif` (2.1.0, `error → warning`); `go vet` + `go test ./...` green; adversarial + reachability review.

```diff
diff --git a/apps/openant-cli/internal/report/sarif.go b/apps/openant-cli/internal/report/sarif.go
index 88c903b..1183e57 100644
--- a/apps/openant-cli/internal/report/sarif.go
+++ b/apps/openant-cli/internal/report/sarif.go
@@ -236,14 +236,20 @@ func fingerprintFor(f Finding, verdict string) string {
 }
 
 // sarifLevelForVerdict maps an OpenAnt verdict to a SARIF result.level.
-// Vulnerable + bypassable surface as `error`; inconclusive/unclear as
-// `warning`; everything else (safe, protected, etc.) as `note` so they
-// don't pollute Code-Scanning alert lists.
+// Vulnerable + bypassable surface as `error`; inconclusive + error (a
+// disclosure-eligible analysis failure — see core/verdict_taxonomy.py) as
+// `warning` so they stay visible; everything else (safe, protected, etc.)
+// as `note` so they don't pollute Code-Scanning alert lists.
 func sarifLevelForVerdict(v string) string {
 	switch v {
 	case "vulnerable", "bypassable":
 		return "error"
-	case "inconclusive", "unclear":
+	// "error" is a real producer verdict (core/verdict_taxonomy.py: PRODUCER_VERDICTS
+	// and DISCLOSURE_ELIGIBLE) meaning a unit whose analysis errored — surfaced so it
+	// stays on the manual-triage radar. It must be VISIBLE, not "note" (which GitHub
+	// Code Scanning filters from the default view). "unclear" is not a producer verdict,
+	// so it has no dedicated arm.
+	case "inconclusive", "error":
 		return "warning"
 	default:
 		return "note"
```

+ a regression test `TestSarifLevelForVerdict_ErrorAndTaxonomy` asserting the full mapping. Full credit for the feature is yours — this is just a one-arm hardening. (One more thing worth a rebase: the Go+Python CI matrix hasn't run on this branch since it predates the current `test.yaml` — a rebase onto master would trigger it.)

gadievron:
Thanks again @ChrisJr404 — I've opened #258, which carries your SARIF 2.1.0 emitter **verbatim** (your original commit preserved, authorship intact) rebased onto current `master`, plus one additive correctness fix: `sarifLevelForVerdict` was routing the `error` verdict to SARIF `note`, which GitHub Code Scanning hides — but `error` is disclosure-eligible (`core/verdict_taxonomy.py`), so a failed-analysis unit was silently vanishing. #258 maps it to the visible `warning` level (no other verdict changes). Full credit to you for the feature. Recommend landing #258 and closing this in its favor.

ChrisJr404:
Thanks Gadi, #258 looks right and I'm glad the attribution carried over. The error→warning mapping is the correct call — errored disclosure-eligible units shouldn't drop out of Code Scanning. Happy to close this in favor of #258.

gadievron:
Reopening this. #258 (a recreation of your SARIF work) was reverted in #261 so that **your original PR lands under your authorship** instead. `master` is now back to a SARIF-free state, so this branch merges cleanly. Thanks for the feature and sorry for the detour, @ChrisJr404 — the goal is for SARIF to ship as your contribution. Once this merges, a small follow-on PR will add one improvement on top (mapping the `error` verdict to a visible SARIF level), explicitly building on your work.

ChrisJr404:
Thanks for reverting to keep this under my authorship, Gadi — appreciate it. Rebased the branch onto the current SARIF-free master (clean replay, no conflicts) and folded in your error->warning correctness fix, so errored disclosure-eligible units now map to a visible `warning` in Code Scanning instead of dropping to `note`. Dropped the dead `unclear` arm too since it isn't a producer verdict, and added a regression test asserting the full mapping table. `go build ./...`, `go vet ./...`, `gofmt`, and `go test ./...` are all green. Should be good to go.

Issue #16, opened 2026-03-23
Fork contributions: bug fixes, Windows compatibility, pipeline resilience, and new features


joshbouncesecurity:
I've been using OpenAnt (primarily on Windows) and have accumulated some changes in [my fork](https://github.com/joshbouncesecurity/OpenAnt) that I'd like to contribute back. I've already opened PR #15 (test suite + CI). Before submitting the rest as individual PRs, I wanted to check which changes you'd be interested in.

I have working implementations for all of these — happy to open PRs for whichever ones you'd like. Check the ones you're interested in, or let me know if you have questions about any of them.

### Testing & CI

- [ ] **1. Test suite + CI** *(already open as PR #15)* — 60 pytest tests covering parsers, token tracking, language detection, and Go CLI integration. GitHub Actions CI on Linux, macOS, and Windows.
- [ ] **2. Add ruff lint to CI** — Adds ruff with two rules: `F821` (undefined name) and `F811` (redefined unused name). Unlike compiled languages, Python won't report an undefined name until that code path executes at runtime, so a missing import can ship undetected and only crash when a user hits that branch. These two rules catch that statically with zero false positives and no style noise.

### Bug Fixes

These exist in the current codebase.

- [x] ~~**3. Findings count shows 0 in build-output** — `PrintBuildOutputSummary` in the Go CLI expects `findings_count` in the JSON response, but the Python CLI returns only `{"pipeline_output_path": path}` — so "Findings included: 0" is always displayed regardless of actual results.~~ *Addressed in upstream PR #23 — `build_pipeline_output` now returns `findings_count` and `cmd_build_output` includes it in the JSON response.*
- [ ] **4. Parse defaults to `--level all` instead of `reachable`** — The Go CLI `parse` command defaults `--level` to `"all"`, while both `scan` and the Python CLI default to `"reachable"`. This means `openant parse` standalone produces a different (larger, noisier) dataset than `openant scan` with no indication to the user.
- [x] ~~**5. Analyze summary missing verdict categories** — The Python analyzer tracks 6 verdict categories (vulnerable, bypassable, inconclusive, protected, safe, errors) but `PrintAnalyzeSummary` only displays Vulnerable and Safe — so the totals don't add up. `PrintScanSummaryV2` already displays all categories correctly; analyze should match.~~ *Addressed in upstream PR #23 — `PrintAnalyzeSummary` now displays all verdict categories (vulnerable, bypassable, protected, safe, inconclusive, errors).*
[...]

PR #35, opened 2026-05-04
fix: default parse --level to reachable to match scan and Python CLI

joshbouncesecurity:
@
## Summary

The Go CLI `parse` command defaults `--level` to `"all"`, while both `scan` and the Python CLI default to `"reachable"`. This means standalone `openant parse` produces a different (larger, noisier) dataset than `openant scan`, with no indication to the user.

This change flips parse's default to `"reachable"` so the three entry points stay consistent.

Addresses item 4 from #16 (does not close the issue).

## Test plan

- [ ] `openant parse <repo>` writes the same set of nodes as `openant scan <repo> --steps parse`.
- [ ] `openant parse <repo> --level all` still produces the larger dataset.
@

### DIFF
--- a/apps/openant-cli/cmd/parse.go
+++ b/apps/openant-cli/cmd/parse.go
@@ -99,7 +99,7 @@ func runParse(cmd *cobra.Command, args []string) {
 	if parseLanguage != "auto" {
 		pyArgs = append(pyArgs, "--language", parseLanguage)
 	}
-	if parseLevel != "all" {
+	if parseLevel != "reachable" {
 		pyArgs = append(pyArgs, "--level", parseLevel)
 	}
 	if manifestPath != "" {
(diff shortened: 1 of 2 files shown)


joshbouncesecurity:
## Manual verification

In addition to the new automated tests:

- [ ] `openant parse <repo>` (no `--level` flag): output matches `openant scan <repo> --steps parse` (same node count). Previously parse produced a larger dataset.
- [ ] `openant parse <repo> --level all`: dataset is larger (compare `wc -l dataset.json` against the no-flag run).
- [ ] `openant parse --help`: confirm `--level` shows `(default "reachable")`.
- [ ] `openant parse <repo> --level codeql` and `--level exploitable`: the explicit flag is forwarded to the Python CLI (was silently dropped in an earlier draft of this PR; now fixed).

joshbouncesecurity:
## Local test results

Built the Go CLI from this branch and verified the help text on Windows.

Commands run:

```
cd .worktrees/fix/issue16-04-parse-level/apps/openant-cli
go build -o openant.exe ./
./openant.exe parse --help
```

Result (relevant flag):

```
      --level string        Processing level: all, reachable, codeql, exploitable (default "reachable")
```

Outcome:

- [x] `openant parse --help` shows `(default "reachable")` ✅

Did not run the full parse-vs-scan parity comparison in this pass (that's well-covered by the automated test in the diff), but the user-facing default is now correctly flipped.

ar7casper (approved):
Good catch

PR #36, opened 2026-05-04
feat: auto-detect dependency changes and reinstall openant

joshbouncesecurity:
## Summary

The Go CLI now hashes `pyproject.toml` after `pip install -e` and re-runs install automatically when dependencies change. Prevents stale venv issues after `git pull` introduces new dependencies.

Addresses item 14 from #16 (does not close the issue).

## Test plan

- [ ] First run after a fresh checkout installs the venv as before.
- [ ] Subsequent runs with unchanged `pyproject.toml` skip the install.
- [ ] Modifying `pyproject.toml` (e.g., adding a dep) causes the next CLI invocation to reinstall automatically.
- [ ] Deleting `~/.openant/venv/.deps-hash` and re-running causes a reinstall and recreates the hash.

### DIFF
--- a/apps/openant-cli/internal/python/runtime.go
+++ b/apps/openant-cli/internal/python/runtime.go
@@ -322,7 +322,13 @@ func depsStaleness(corePath string) (stale bool, currentHash string, err error) 
 // If stale, it re-runs pip install -e and updates the stored hash.
 // Returns nil if deps are up-to-date or were successfully refreshed.
 func CheckDepsStale(pythonPath string) error {
-	corePath, err := findOpenantCore()
+	return checkDepsStaleWith(pythonPath, findOpenantCore)
+}
+
+// checkDepsStaleWith is the testable core of CheckDepsStale; coreFinder is
+// injected so tests can avoid os.Chdir to simulate a missing source tree.
+func checkDepsStaleWith(pythonPath string, coreFinder func() (string, error)) error {
+	corePath, err := coreFinder()
 	if err != nil {
 		// Can't find source — skip staleness check
 		return nil
@@ -338,6 +344,9 @@ func CheckDepsStale(pythonPath string) error {
 	}
 
 	fmt.Fprintln(os.Stderr, "Dependencies changed, updating openant installation...")
+	// Known limitation: concurrent invocations that both detect stale deps
+	// will race to pip-install into the same venv. pip does not support
+	// concurrent writes; an OS-level lock would be needed to close this gap.
 	if err := installOpenant(pythonPath, corePath); err != nil {
 		return fmt.Errorf(
 			"failed to update openant dependencies: %w\n"+
--- a/apps/openant-cli/internal/python/runtime.go
+++ b/apps/openant-cli/internal/python/runtime.go
@@ -2,6 +2,8 @@
 package python
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"os/exec"
@@ -198,6 +200,12 @@ func CheckOpenantInstalled(pythonPath string) error {
 		)
 	}
 
+	// Save dependency hash so CheckDepsStale knows this is the baseline.
+	pyprojectPath := filepath.Join(corePath, "pyproject.toml")
+	if h, err := hashFile(pyprojectPath); err == nil {
+		_ = writeStoredHash(h)
+	}
+
 	fmt.Fprintln(os.Stderr, "openant installed successfully.")
 	return nil
 }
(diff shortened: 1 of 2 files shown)


joshbouncesecurity:
## Manual verification

After installing this branch:

- [x] **First run**: `openant --version` triggers initial pip install. Confirm `~/.openant/venv/.deps-hash` is created.
- [x] **Second run, no changes**: `openant --version` skips install — should be near-instant with no install message.
- [x] **Pyproject changed**: edit `libs/openant-core/pyproject.toml` (e.g., add a comment), re-run: prints "Dependencies changed, updating..." and reinstalls.
- [x] **Hash file deleted**: `rm ~/.openant/venv/.deps-hash` then re-run: reinstalls and recreates the hash.
- [ ] **Read-only venv parent**: if `~/.openant/venv/` doesn't exist (e.g., custom `OPENANT_PYTHON`), the helper now MkdirAll's it instead of erroring.

joshbouncesecurity:
## Local test results

Built the Go CLI from this branch on Windows and exercised the deps-hash flow. (Note: I needed `OPENANT_PYTHON` pointing at `Scripts/python.exe` because the managed-venv path uses `bin/python` which is the documented Windows limitation, unrelated to this PR.)

Commands run:

```
go build -o openant.exe ./
# Hash exists from prior install
ls ~/.openant/venv/.deps-hash      # present, sha256 = f251e66e...

# Delete hash to simulate "first run / hash missing"
rm ~/.openant/venv/.deps-hash
OPENANT_PYTHON=...Scripts/python.exe ./openant.exe parse <fixture>
# -> ran install, recreated .deps-hash

# Run again with hash present and no pyproject change
./openant.exe parse <fixture>
# -> NO install message, jumped straight to parser
```

Outcome:

- [x] First run / hash missing: install ran, `.deps-hash` (re)created ✅
- [x] Second run, no changes: skipped install, near-instant ✅
- [ ] Pyproject-changed and hash-deleted paths: I tested the hash-deleted path (covers the same code path as pyproject-changed since both produce a hash mismatch). Did not edit pyproject.toml to avoid touching the worktree.
- (n/a) Read-only-venv-parent MkdirAll branch: not exercised — the venv parent already existed.

Sample output of the post-install run for transparency:

```
Creating managed Python environment at ~/.openant/venv/...
Installing openant from ...libs/openant-core...
[Reachability Filter] ...  -> 5 units written
```

The `.deps-hash` was correctly written after the install completed.

apps/openant-cli/internal/python/runtime_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+func TestCheckDepsStale_SkipsWhenCoreNotFound(t *testing.T) {
+	// CheckDepsStale should silently return nil when it can't find
+	// openant-core. We chdir to a temp dir so findOpenantCore fails.
+	origDir, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	tmpDir := t.TempDir()
+	if err := os.Chdir(tmpDir); err != nil {
```
ar7casper: `os.Chdir` mutates process state, restored via `t.Cleanup`. Works today because Go runs same-package tests sequentially, but if anyone adds `t.Parallel()` to a sibling test in this package you'd get racy CWD reads. Cleaner pattern: take a search-root parameter on `findOpenantCore` and inject the temp dir, mirroring what you did with `depsStalenessAt(corePath, hashPath)`. Not a blocker — flagging for future-proofing.

apps/openant-cli/internal/python/runtime.go
```diff
@@ -216,13 +242,116 @@ func EnsureRuntime() (*RuntimeInfo, error) {
...
+	if err != nil {
+		// Can't read pyproject.toml — skip check
+		return nil
+	}
+	if !stale {
+		return nil // deps are up-to-date
+	}
+
+	fmt.Fprintln(os.Stderr, "Dependencies changed, updating openant installation...")
+	if err := installOpenant(pythonPath, corePath); err != nil {
```
ar7casper, line 336: Concurrent invocations could double-install: two terminals running `openant parse` and `openant analyze` simultaneously right after a `git pull` would both detect stale deps and both fire `pip install -e` against the same venv. pip is not safe under concurrent installs to the same environment. A file lock around this branch (e.g. an OS-level lock on `<venvDir>/.deps-install.lock`) would fix it. Rare in practice but worth either fixing here or noting as a known limitation.

apps/openant-cli/internal/python/runtime_test.go
joshbouncesecurity: Good catch. Extracted `checkDepsStaleWith(pythonPath string, coreFinder func() (string, error)) error` and made `CheckDepsStale` a thin wrapper that passes `findOpenantCore`. `TestCheckDepsStale_SkipsWhenCoreNotFound` now injects a stub finder that returns an error directly — no `os.Chdir` needed. Pushed in 9eae267.

apps/openant-cli/internal/python/runtime.go
joshbouncesecurity: Opting to document rather than fix — a correct cross-platform file lock without adding a dependency (no stdlib `flock` on Windows) is out of scope here, and the race requires two simultaneous invocations right after a `git pull`. Added a "Known limitation" comment above the `installOpenant` call in `checkDepsStaleWith` (9eae267). Happy to revisit in a follow-up if it becomes a real-world issue.

PR #37, opened 2026-05-04
fix: lazy-install JS parser npm deps on first use

joshbouncesecurity:
## Summary

`openant parse` on a JS/TS repo fails out of the box with `Cannot find module 'ts-morph'` because nothing in the install flow populates `parsers/javascript/node_modules/`.

This adds a lazy bootstrap in `_parse_javascript` that runs `npm install` once if `node_modules/` is missing, mirroring the Go CLI's existing venv bootstrap pattern. Lazy (on first JS parse) rather than eager (at CLI startup) so users who only scan Python/Go repos don't need Node/npm installed.

If `npm` is missing or the install fails, the helper raises with a clear message pointing at the parser directory.

Addresses item 21 from #16 (does not close the issue).

## Test plan

- [ ] First JS parse on a clean checkout: bootstrap runs, npm install completes, parse proceeds.
- [ ] Subsequent JS parses skip the bootstrap (node_modules already present).
- [ ] Python/Go scans on a machine without Node installed: no npm invocation, no failure.
- [ ] `npm` missing on PATH: clear error message names the parser directory.
- [ ] `tests/test_js_parser_bootstrap.py` passes (5 unit tests covering the four branches plus the integration surface).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ __pycache__/
 node_modules/
 apps/openant-cli/bin/
 libs/openant-core/parsers/go/go_parser/go_parser
+libs/openant-core/parsers/javascript/.openant-npm-install.lock
 # docs/
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -11,6 +11,7 @@ sys.path hacks in the original code.
 
 import json
 import os
+import shutil
 import subprocess
 import sys
 from pathlib import Path
@@ -19,6 +20,9 @@ from core.schemas import ParseResult
 
 # Root of openant-core (where parsers/ lives)
 _CORE_ROOT = Path(__file__).parent.parent
+
+# JS parser directory (holds its own package.json / node_modules)
+_JS_PARSER_DIR = _CORE_ROOT / "parsers" / "javascript"
 
 
 def detect_language(repo_path: str) -> str:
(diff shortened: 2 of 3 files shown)


joshbouncesecurity:
## Manual verification

- [ ] **Fresh checkout** (no `parsers/javascript/node_modules/`): `openant parse <ts-repo>` prints "installing deps (first run, this may take a minute)..." then proceeds; `node_modules/` populated.
- [ ] **Second JS parse**: skips the bootstrap (no install message).
- [ ] **Partial install**: `rm parsers/javascript/node_modules/.package-lock.json` to simulate a Ctrl+C mid-install. Next parse re-runs `npm install` (sentinel detection).
- [ ] **Concurrent first-parse**: start two `openant parse <js-repo>` simultaneously. Lockfile serializes them; only one runs install; second waits then proceeds. No `node_modules/` corruption.
- [ ] **Missing npm**: PATH without `npm`: clear error message includes "you can reproduce with: npm install (from <dir>)".
- [ ] **Missing package.json**: rare but covered — early error rather than opaque npm failure.
- [ ] **Python-only / Go-only scan** on a machine without Node installed: no npm invocation, no error.

joshbouncesecurity:
## Local test results

Tested the lazy bootstrap on Windows by importing `core.parser_adapter` directly from this branch's worktree (which had no `parsers/javascript/node_modules/`).

Commands run:

```
# Branch worktree shipped without node_modules
ls libs/openant-core/parsers/javascript/node_modules
# -> No such file or directory

python -c "
from core import parser_adapter as pa
print('Before:', pa._js_deps_installed())
pa._ensure_js_parser_dependencies()
print('After:', pa._js_deps_installed())
"
```

First-call output:

```
[Parser] Installing JS parser dependencies (first run, this may take a minute)...
added 14 packages, and audited 15 packages in 2s
Before: False
Calling _ensure_js_parser_dependencies...
After: True
```

Second-call output (run immediately after):

```
Already installed: True
After (no-op): True
```

Outcome:

- [x] Fresh checkout (no `node_modules/`): bootstrap printed `Installing JS parser dependencies (first run, this may take a minute)...` and ran `npm install` ✅
- [x] `node_modules/` populated after the call (ts-morph, typescript, etc. present) ✅
- [x] Second invocation skipped silently — no install message, sentinel returns True ✅
- [ ] Did not exercise missing-npm, partial-install (`.package-lock.json` removed), concurrent-first-parse, or missing-`package.json` paths — those are covered by `tests/test_js_parser_bootstrap.py` in the diff.

ar7casper (approved):

Issue #21, opened 2026-03-29
[bug] JavaScript parser misses Express.js anonymous route handler callbacks


joshbouncesecurity:
## Problem

The JavaScript/TypeScript parser pipeline (`typescript_analyzer.js` → `unit_generator.js`) does not extract anonymous arrow function callbacks used as Express.js route handlers. This means the majority of application code in a typical Express app is invisible to analysis.

## Example

Given a standard Express route definition:

```js
router.post('/orders', authenticateToken, async (req, res) => {
  const { productId, quantity } = req.body;
  // ... route handler logic ...
});
```

The parser extracts `authenticateToken` (a named function) but **not** the anonymous `async (req, res) => { ... }` callback that contains the actual business logic and potential vulnerabilities.

## Impact

In a test codebase with 9 JS files, ~20 route handlers, and 7 known security vulnerabilities:

- **typescript_analyzer** found only **8 functions** (5 named functions + 3 route_handlers)
- **unit_generator** produced **8 units**
- **reachability_filter** reduced to **2 units** (only `authenticateToken` and `authenticateApiKey` matched `req.headers` input patterns)
- All 7 vulnerabilities live in the anonymous route handler callbacks that were never extracted

The 2 surviving units are middleware functions — none of the actual route handler code was analyzed.

## Pipeline output

```
STAGE: typescript_analyzer — 8 functions (function: 5, route_handler: 3)
STAGE: unit_generator — 8 units, call graph: 7 edges
STAGE: reachability_filter — Entry points: 2, Units: 8 → 2 (75% reduction)
```

## Expected behavior

The parser should recognize Express route handler patterns and extract the callback functions as units, ideally with:
- The HTTP method and path as metadata (e.g., `POST /orders`)
- `is_entry_point: true` since they directly receive HTTP request data (`req.body`, `req.params`, `req.query`)
- Correct call graph edges to middleware functions used in the same route

## Workaround

Refactoring route handlers to use named functions works:

```js
async function createOrder(req, res) { ... }
router.post('/orders', authenticateToken, createOrder);
```

But this shouldn't be necessary — the anonymous callback pattern is idiomatic Express.js.

## Environment

- OpenAnt CLI (Windows)
- JavaScript parser pipeline
- Express.js 4.x codebase with CommonJS modules

PR #38, opened 2026-05-04
feat: add --fresh flag to parse for forced reparse

joshbouncesecurity:
## Summary

Adds `--fresh` to `openant parse` to force a full reparse from scratch without manually deleting `dataset.json`. Useful when parser improvements are deployed and the existing dataset needs to be regenerated.

The JS parser also now logs a hint pointing at `--fresh` when existing units are reused, so users discover the flag when they need it.

Addresses item 19 from #16 (does not close the issue).

## Test plan

- [ ] `openant parse <repo>` reuses existing units when `dataset.json` already exists (default).
- [ ] `openant parse <repo> --fresh` deletes the cached dataset and reparses from scratch.
- [ ] After running with `--fresh`, the JS parser hint about reused units no longer fires for that run.

### DIFF
--- a/apps/openant-cli/cmd/parse.go
+++ b/apps/openant-cli/cmd/parse.go
@@ -29,6 +29,7 @@ var (
 	parseDiffBase  string
 	parsePR        int
 	parseDiffScope string
+	parseFresh     bool
 )
 
 func init() {
@@ -38,6 +39,7 @@ func init() {
 	parseCmd.Flags().StringVar(&parseDiffBase, "diff-base", "", "Incremental mode: tag units overlapping diff vs this ref")
 	parseCmd.Flags().IntVar(&parsePR, "pr", 0, "Incremental mode against a GitHub PR number (mutex with --diff-base)")
 	parseCmd.Flags().StringVar(&parseDiffScope, "diff-scope", "changed_functions", "Diff scope: changed_files, changed_functions, callers")
+	parseCmd.Flags().BoolVar(&parseFresh, "fresh", false, "Delete existing dataset and reparse from scratch")
 }
 
 // buildParsePyArgs assembles the argv passed to the Python `openant parse`
@@ -114,6 +116,9 @@ func runParse(cmd *cobra.Command, args []string) {
 	}
 
 	pyArgs := buildParsePyArgs(repoPath, parseOutput, datasetName, parseLanguage, parseLevel, manifestPath)
+	if parseFresh {
+		pyArgs = append(pyArgs, "--fresh")
+	}
 
 	result, err := python.Invoke(rt.Path, pyArgs, "", quiet, resolvedAPIKey())
 	if err != nil {
--- a/apps/openant-cli/cmd/parse.go
+++ b/apps/openant-cli/cmd/parse.go
@@ -39,15 +39,14 @@ func init() {
 	parseCmd.Flags().StringVar(&parseDiffBase, "diff-base", "", "Incremental mode: tag units overlapping diff vs this ref")
 	parseCmd.Flags().IntVar(&parsePR, "pr", 0, "Incremental mode against a GitHub PR number (mutex with --diff-base)")
 	parseCmd.Flags().StringVar(&parseDiffScope, "diff-scope", "changed_functions", "Diff scope: changed_files, changed_functions, callers")
-	parseCmd.Flags().BoolVar(&parseFresh, "fresh", false, "Delete existing dataset and reparse from scratch")
+	parseCmd.Flags().BoolVar(&parseFresh, "fresh", false, "Delete existing dataset.json and reparse from scratch (other artifacts preserved)")
 }
 
-// buildParsePyArgs assembles the argv passed to the Python `openant parse`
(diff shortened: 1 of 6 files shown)


joshbouncesecurity:
## Manual verification

- [ ] `openant parse <repo>` twice in a row: second run is fast (cached `dataset.json` reused).
- [ ] `openant parse <repo> --fresh`: deletes existing `dataset.json`, runs full reparse (longer).
- [ ] `openant parse <missing-repo> --fresh`: doesn't crash when `dataset.json` doesn't pre-exist (no-op).
- [ ] `openant parse --help`: `--fresh` listed; help text mentions "only deletes dataset.json; other artifacts in the output dir are preserved".
- [ ] **Race**: two `openant parse --fresh` simultaneously on the same repo: no `FileNotFoundError` from racing `os.remove` (catch added).

joshbouncesecurity:
## Local test results

Built the Go CLI from this branch and exercised `--fresh` end-to-end on Windows using `libs/openant-core/tests/fixtures/sample_python_repo`.

Commands run:

```
go build -o openant.exe ./
./openant.exe parse <fixture> --output _out          # run 1: fresh output dir
./openant.exe parse <fixture> --output _out --fresh  # run 2: --fresh on existing dataset
./openant.exe parse <fixture> --output _empty --fresh  # run 3: --fresh with no pre-existing dataset
```

Outcome:

- [x] `--fresh` listed in `parse --help` with description "Delete existing dataset.json and reparse from scratch (other artifacts preserved)" ✅
- [x] Run with `--fresh` against existing dataset prints `[Parser] --fresh: deleted existing dataset.json` and rebuilds (mtime advanced) ✅
- [x] Run with `--fresh` and no pre-existing dataset.json: no crash, no "deleted" message, parse runs cleanly ✅
- [ ] Race-condition catch and JS-parser hint: not exercised in this manual pass (covered by automated tests in the diff).

ar7casper:
Hey @joshbouncesecurity — running through your draft batch, this one's the easiest to land. Sharing a quick review now since it looks essentially mergeable:

**Things done quietly well:**

- **TOCTOU-safe deletion** in `parser_adapter.py:108-116` — `try/except FileNotFoundError` instead of `if exists() then remove()`, with a comment explaining the race. Easy to miss; you didn't.
- **Test pyramid is right**: Python (149 LOC, 5 cases) covers the actual deletion semantics with a stubbed parser; Go (114 LOC, 5 cases) covers flag registration + arg pass-through. No overlap, no gaps.
- **Refactor commit `eb8ce74` extracts `buildParsePyArgs`** specifically so the Go side becomes testable without invoking the Python runtime — good engineering.
- **Help text accurately scopes the deletion**: "Delete existing dataset.json … other artifacts preserved." Important — `--fresh` reads broader than what it does.
- **Discoverability via JS hint**: surfaces the flag exactly when duplicates are skipped.
- **Order is correct**: `--fresh` deletion runs before `_maybe_apply_diff_filter`, so `--fresh + --diff-manifest` composes cleanly.

**Three small things before flipping out of draft:**

1. **Wrong issue reference in commit subject `819ad20`** — the message reads `feat: add --fresh flag to parse command (#21)`. `#21` is the Express anonymous handler bug. This PR addresses item 19 from #16. Worth a `git commit --amend` or fixup before the squash so it doesn't end up wrong in history forever.

2. **JS hint references a flag the JS parser doesn't have** — `unit_generator.js:420`:

   ```js
   console.error(`  Note: ${duplicateCount} existing units kept as-is (use --fresh to regenerate all units)`);
   ```

   `unit_generator.js` is also runnable standalone; `--fresh` only exists on the `openant parse` wrapper. Tighten to:

   ```js
   console.error(`  Note: ${duplicateCount} existing units kept as-is (use 'openant parse --fresh' to regenerate all units)`);
   ```

3. **`--fresh` + `--diff-manifest` interaction not covered by tests.** The composition is mechanically correct (verified by reading the order), but locking it down with one extra Python test would close the gap.

**Nit (optional):** `cli.py:124` uses `fresh=getattr(args, "fresh", False)`. `argparse` guarantees `args.fresh` exists when the flag has a `default=`, so `args.fresh` would work. Matches the surrounding style (`name`, `diff_manifest` are the same), so probably right to leave consistent.

Address #1-#3 and this is good to flip out of draft. Nice work.

joshbouncesecurity:
Thanks @ar7casper, addressed in 56ce18b:

1. **Wrong issue ref in `819ad20`** — agreed it's wrong, but since this will be squash-merged the PR title (which is correct) is what'll land in history. Happy to fixup the commit if you'd prefer it clean regardless.

2. **JS hint** — fixed to `'openant parse --fresh'`.

3. **`--fresh` + `--diff-manifest` test** — added `test_fresh_and_diff_manifest_compose_correctly`.

Also fixed a regression I'd introduced relative to upstream: the branch had reverted the `--level` default back to `"all"` (diverged from the fix in #35). Restored to `"reachable"` and merged the level tests from `knostic/master` into `parse_test.go` alongside the new `--fresh` tests.

ar7casper (approved):

PR #40, opened 2026-05-04
feat: auto-detect language in init

joshbouncesecurity:
## Summary

Makes `--language` optional in `openant init` by auto-detecting the project language from file extensions in the target path. Also drops the assumption that the path is a git repository for local-only mode. New users can run `openant init <path>` and have it just work.

Addresses item 13 from #16 (does not close the issue).

## Test plan

- [ ] `openant init <python-repo>` — language detected as Python.
- [ ] `openant init <ts-repo>` — language detected as TypeScript.
- [ ] `openant init <repo> --language go` — explicit flag wins.
- [ ] Path that isn't a git repo: init still proceeds.
- [ ] Unit tests cover detection for Python/JS/TS/Go and explicit-flag override.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ apps/openant-cli/bin/
 libs/openant-core/parsers/go/go_parser/go_parser
 libs/openant-core/parsers/javascript/.openant-npm-install.lock
 _docs/
-
+docs/
+.worktrees/
--- a/apps/openant-cli/cmd/init.go
+++ b/apps/openant-cli/cmd/init.go
@@ -50,7 +50,7 @@ var (
 )
 
 func init() {
-	initCmd.Flags().StringVarP(&initLanguage, "language", "l", "", "Language to analyze: python, javascript, go, c, ruby, php, auto (default: auto-detect)")
+	initCmd.Flags().StringVarP(&initLanguage, "language", "l", "", "Language to analyze: python, javascript, go, c, ruby, php, zig, auto (default: auto-detect)")
 	initCmd.Flags().StringVar(&initCommit, "commit", "", "Specific commit SHA (default: HEAD)")
 	initCmd.Flags().StringVar(&initName, "name", "", "Override project name (default: derived from URL/path)")
 	initCmd.Flags().BoolVar(&initFull, "full", false, "Force full scan (rejects --incremental/--diff-base/--pr)")
@@ -363,7 +363,7 @@ func detectLanguage(repoPath string) (string, error) {
 	if bestLang == "" {
 		return "", fmt.Errorf(
 			"no supported source files found in %s. "+
-				"Supported languages: Python, JavaScript/TypeScript, Go, C/C++, Ruby, PHP",
+				"Supported languages: Python, JavaScript/TypeScript, Go, C/C++, Ruby, PHP, Zig",
 			repoPath,
 		)
 	}
--- a/apps/openant-cli/cmd/init.go
+++ b/apps/openant-cli/cmd/init.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -21,12 +23,16 @@ var initCmd = &cobra.Command{
 For remote URLs, the repo is cloned into ~/.openant/projects/{org}/{repo}/repo/.
 For local paths, the existing directory is referenced in place (no cloning).
 
+If --language is not specified, the dominant language is auto-detected by
+counting source files in the repository.
+
 After init, all commands (parse, scan, etc.) work without path arguments.
 
 Examples:
+  openant init https://github.com/grafana/grafana
   openant init https://github.com/grafana/grafana -l go
   openant init https://github.com/grafana/grafana -l go --commit 591ceb2eec0
-  openant init ./repos/grafana -l go
+  openant init ./repos/grafana
   openant init ./repos/grafana -l go --name myorg/grafana`,
 	Args: cobra.ExactArgs(1),
 	Run:  runInit,
@@ -44,7 +50,7 @@ var (
 )
 
 func init() {
-	initCmd.Flags().StringVarP(&initLanguage, "language", "l", "", "Language to analyze: python, javascript, go, c, ruby, php (required)")
+	initCmd.Flags().StringVarP(&initLanguage, "language", "l", "", "Language to analyze: python, javascript, go, c, ruby, php, auto (default: auto-detect)")
 	initCmd.Flags().StringVar(&initCommit, "commit", "", "Specific commit SHA (default: HEAD)")
 	initCmd.Flags().StringVar(&initName, "name", "", "Override project name (default: derived from URL/path)")
 	initCmd.Flags().BoolVar(&initFull, "full", false, "Force full scan (rejects --incremental/--diff-base/--pr)")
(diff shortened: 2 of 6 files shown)


joshbouncesecurity:
## Manual verification

- [ ] `openant init <python-repo>` (no `-l`): detected as `python`.
- [ ] `openant init <ts-repo>`: detected as `javascript` (TS dispatched through the JS parser; matches existing semantics).
- [ ] `openant init <go-repo>`: detected as `go`.
- [ ] `openant init <repo> -l python` on a JS repo: explicit flag wins.
- [ ] `openant init <non-git-dir>`: succeeds (no longer requires `.git/`); commit_sha reported as `nogit`.
- [ ] `openant init <non-git-dir> --commit <sha>`: warning printed, `nogit` still used.
- [ ] `openant init <empty-dir>`: clean error message listing supported languages, suggests `-l/--language`.
- [ ] **Mixed-language repo** (more .py than .ts at root): detected as `python`.
- [ ] **Repo containing skip_dirs only** (e.g. only `node_modules/` populated): empty-dir error path.

joshbouncesecurity:
## Local test results

Built the Go CLI from this branch on Windows and exercised auto-detect on the in-repo fixtures (`libs/openant-core/tests/fixtures/sample_python_repo` and `sample_js_repo`).

Commands run:

```
go build -o openant.exe ./
./openant.exe init <sample_python_repo>
./openant.exe init <sample_js_repo>
```

Outcome:

- [x] Python fixture: `Auto-detecting language... Detected language: python` ✅
- [x] JS fixture: `Auto-detecting language... Detected language: javascript` ✅
- [x] Both succeed against a non-git directory (commit reported as `nogit`) ✅
- [ ] Did not separately test explicit `-l` override, mixed-language tie-break, empty-dir error message, or `--commit` warning on non-git — those are covered by the unit tests in the diff.

Help text reads:

```
If --language is not specified, the dominant language is auto-detected by
counting source files in the repository.
```

ar7casper (commented):
Lots of value here — let me split the response into two parts.

**CI fix** (already visible from the failing checks): `core/parser_adapter.py:30`'s bare `open(_LANGUAGES_CONFIG)` trips the UTF-8 regression scanner that landed in #56 (which you just rebased onto). One-line fix:

```python
from utilities.file_io import read_json

def _load_language_config() -> dict:
    return read_json(_LANGUAGES_CONFIG)
```

That's the only thing keeping CI red.

**Structural ask: could you drop the `MarkFlagRequired("language")` removal from this PR?**

We've opened #61 to track the validation work needed before we're comfortable flipping `-l` from required to default-auto. The concern isn't your implementation — it's that the underlying dominance heuristic is byte-for-byte the same as what's on master, and we previously made `-l` required because real-world reliability wasn't where we wanted it. We'd like to keep `-l` required for now, expose `-l auto` as an opt-in users can try, and flip the default once #61's three pieces (E2E fixtures + polyglot regression suite + OSS calibration corpus) give us evidence the algorithm holds up.

That keeps the parts of this PR with the most concrete value for end users — the shared `config/languages.json` (eliminates Go↔Python drift), the non-git path (real UX win for tarball-style installs), the test infrastructure, the Zig alignment — and just defers the default-flip until #61 is closed.

Concrete asks for this PR:

1. Fix the bare `open()` (CI green).
2. Restore `_ = initCmd.MarkFlagRequired("language")` in `init.go`.
3. Keep `-l auto` working as an opt-in — your existing `if initLanguage == "" || initLanguage == "auto"` block stays, just the user has to explicitly write `-l auto` to trigger it.
4. Update the `-l` help text to flag `auto` as experimental, e.g. `Language to analyze: python, javascript, go, c, ruby, php, zig, auto (auto = experimental dominance heuristic; see #61)`.

Happy to move the validation discussion to #61 if you want to dig into the testing plan there.

Smaller nits for whenever (not blockers):

- The `_LANGUAGES_CONFIG` path resolution does four `.parent`s — fragile if openant-core ever ships standalone or moves. `importlib.resources` reading from inside the package would be more robust.
- `detect_language`'s docstring still says `Returns: One of: "python", "javascript", "go", "c", "ruby", "php"` — missing `"zig"`. Your commit `5ac863c3` aligned the help text and error messages but the docstring slipped through.

joshbouncesecurity:
Thanks for the detailed review. All four concrete asks are addressed in the latest push:

1. **Bare `open()` → `read_json`** — `_load_language_config()` now delegates to `read_json(_LANGUAGES_CONFIG)` from `utilities.file_io`. CI should be green.
2. **`MarkFlagRequired("language")` restored** — `-l` is required again; auto-detect is opt-in via `-l auto`.
3. **`-l auto` still works** — the `if initLanguage == "" || initLanguage == "auto"` block is unchanged, `auto` is just no longer the implicit default.
4. **Help text updated** — `auto = experimental dominance heuristic; see #61`.

Also addressed the two nits:
- Added `"zig"` to the `detect_language` docstring.
- `_JS_PARSER_DIR` (from the upstream base) is kept alongside `_LANGUAGES_CONFIG` — both are preserved in the resolved conflict.

The `_LANGUAGES_CONFIG` path resolution comment (four `.parent`s) is noted — happy to follow up in a separate PR once this lands, since `importlib.resources` would be a refactor with its own test surface.

Rebased onto `upstream/release/2026-05-12` and force-pushed.

joshbouncesecurity:
@ar7casper responded here

ar7casper (commented):
Round-2 review — thanks for the quick turnaround on the four round-1 asks.

**Verified each fix in code:**

| Round-1 ask | Status |
|---|---|
| Fix bare `open()` (CI green) | ✅ `parser_adapter.py:33-34` uses `read_json(_LANGUAGES_CONFIG)` from `utilities.file_io` |
| Restore `MarkFlagRequired("language")` | ✅ `init.go:58` |
| Keep `-l auto` working as opt-in | ✅ `init.go:137` triggers detection on `-l auto` (or empty) |
| Help text marks `auto` as experimental + points at #61 | ✅ Exactly the wording I'd hoped for |
| Bonus: `zig` added to `detect_language` docstring | ✅ |

CI went from 6/9 (Python tests red on the bare `open()`) → 9/9 green. Test suite updated to pass `-l auto` explicitly in the 7 `TestInit` cases that exercise auto-detection. No new bugs introduced by the rework.

One small housekeeping nit:

## 🟢 Low — PR body is stale

The summary still describes round-1's behavior:

> Makes `--language` optional in `openant init` by auto-detecting the project language from file extensions in the target path. New users can run `openant init <path>` and have it just work.

But the latest commits restored `MarkFlagRequired("language")` — `-l` is required again, users now have to explicitly pass `-l auto` to opt into detection. The body's example list (`openant init <url> -l auto`) is correct, but the prose summary contradicts the actual behavior.

**Suggestion:** update the Summary to reflect the current scope (`-l` required + `-l auto` opt-in + forward reference to #61 explaining why default-on is gated).

---

Non-blocking carry-overs (already noted in round-1):

- `_LANGUAGES_CONFIG = Path(__file__).parent.parent.parent.parent / "config" / ...` — the four-`.parent` walk still works in the current monorepo layout but breaks if openant-core is ever vendored or pip-installed standalone. `importlib.resources` from inside the package would be more robust. Follow-up cleanup.
- `init.go:137` does `if initLanguage == "" || initLanguage == "auto"` — with `MarkFlagRequired`, the empty branch is only reachable on an explicit `-l ""`. Consider simplifying to `if initLanguage == "auto"` and letting the empty case error out naturally. Truly minor.

**Verdict:** ✅ ready to merge once the body is refreshed (or even as-is if you're fine fixing the body in-flight). Approving via UI.

PR #43, opened 2026-05-04
fix: atomic JSON writes for pipeline outputs

joshbouncesecurity:
## Summary

Introduces a new `atomic_write_json()` helper and applies it to all final
pipeline output files. The helper writes JSON to a temp file in the **same
directory** as the target, fsyncs, then `os.replace`s it onto the target path.
If a crash or power loss occurs mid-write, the previous output file is
preserved intact rather than being left truncated.

Upstream uses plain `write_json()` (a `json.dump` wrapper), which can corrupt
multi-hour scan results on interrupt.

Same-directory temp is load-bearing on Windows: `os.replace` is only atomic
when source and target sit on the same volume; cross-volume falls back to
copy+delete and loses atomicity.

**Call sites covered** (one write each, at the very end of a pipeline stage):
- `core/analyzer.py` — `results.json`
- `core/enhancer.py` — `enhanced_dataset.json`
- `core/verifier.py` — `results_verified.json`
- `core/reporter.py` — `pipeline_output.json` (final report)
- `core/parser_adapter.py` — `dataset.json` + `analyzer_output.json`
- `experiment.py` — experiment results (legacy direct-run path)

**Not changed:** checkpoint files (designed for incremental writes), scanner
intermediate state, and the diff-filter sidecar report (cheap to regenerate).

**Performance:** one `fsync` + one rename per output file, all of which occur
once at the end of a stage that runs for minutes. Negligible overhead.
`fsync` is silently skipped on filesystems where it fails (network mounts etc.)
so atomicity degrades gracefully to rename-only in those environments.

Addresses item 20 from #16 (does not close the issue).

## Test plan

- [x] Helper round-trip preserves data (incl. `ensure_ascii=False` Unicode).
- [x] Mid-write failure leaves the existing file intact.
- [x] Mid-write failure does not create a target if none existed.
- [x] No stray `.tmp-` files left on failure.
- [x] Temp file is in the same directory as the target.
- [x] POSIX permissions match umask (not mkstemp's 0600 default).
- [x] Existing tests pass (113 collected, 91 passed, 22 skipped — go/js parser binaries unavailable in this env).

### DIFF
--- a/libs/openant-core/core/analyzer.py
+++ b/libs/openant-core/core/analyzer.py
@@ -27,6 +27,7 @@ from core.progress import ProgressReporter
 
 # Import existing analysis machinery
 from utilities.llm_client import AnthropicClient, get_global_tracker
+from utilities.atomic_io import atomic_write_json
 from utilities.file_io import read_json, write_json
 from utilities.json_corrector import JSONCorrector
 from utilities.rate_limiter import get_rate_limiter, is_rate_limit_error, is_retryable_error
@@ -512,7 +513,7 @@ def run_analysis(
         "code_by_route": code_by_route,
     }
 
-    write_json(results_path, experiment_result)
+    atomic_write_json(results_path, experiment_result)
     print(f"\n[Analyze] Results written to {results_path}", file=sys.stderr)
 
     # Checkpoints are preserved as a permanent artifact alongside results.
--- a/libs/openant-core/core/enhancer.py
+++ b/libs/openant-core/core/enhancer.py
@@ -17,6 +17,7 @@ from core.schemas import EnhanceResult, UsageInfo
 from core import tracking
 from core.progress import ProgressReporter
 from utilities.rate_limiter import configure_rate_limiter
+from utilities.atomic_io import atomic_write_json
 from utilities.file_io import read_json, write_json
 
 
@@ -137,7 +138,7 @@ def enhance_dataset(
 
     # Write enhanced dataset
     os.makedirs(os.path.dirname(os.path.abspath(output_path)), exist_ok=True)
-    write_json(output_path, enhanced)
+    atomic_write_json(output_path, enhanced)
     print(f"[Enhance] Enhanced dataset: {output_path}", file=sys.stderr)
     print(f"[Enhance] Classifications: {classifications}", file=sys.stderr)
     if error_count:
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -18,6 +18,7 @@ import sys
 from pathlib import Path
 
 from core.schemas import ParseResult
+from utilities.atomic_io import atomic_write_json
 from utilities.file_io import open_utf8, read_json, write_json
 
 # Root of openant-core (where parsers/ lives)
@@ -173,7 +174,7 @@ def _maybe_apply_diff_filter(
 
     stats = apply_diff_filter(units, manifest)
 
-    write_json(result.dataset_path, dataset)
+    atomic_write_json(result.dataset_path, dataset)
     # Expose stats on the ParseResult via a side-channel file; the parse
     # step_context reads this when assembling parse.report.json.
     diff_report_path = os.path.join(output_dir, "diff_filter.report.json")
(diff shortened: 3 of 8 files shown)


joshbouncesecurity:
## Manual verification

- [ ] Run a small `openant scan` to completion: `results.json`, `enhanced_dataset.json`, `results_verified.json` all exist and parse as valid JSON.
- [ ] **Atomicity smoke test**: start a scan, kill the process (Ctrl+C) during the analyze stage. Re-inspect `results.json` — should be either the previous good version OR non-existent. Never truncated/corrupt.
- [ ] **Permissions (POSIX)**: confirm output file mode matches umask (typically 0644), not 0600.
- [ ] **Same-dir temp**: while a write is happening, a hidden `*.tmp` file appears in the SAME directory as the target (use `watch ls -la <scan-dir>` or `lsof`). Cross-device renames aren't atomic.
- [ ] **Unicode**: a finding containing non-ASCII text (e.g., a Japanese string) round-trips through `results_verified.json` without escape sequences (`ensure_ascii=False` preserved).

joshbouncesecurity:
## Local test results

Exercised `utilities.atomic_io.atomic_write_json` directly on Windows from this branch.

Test script (round-trip, non-ASCII, no-stray-tempfile, preserve-on-failure, implementation-shape):

```python
from utilities.atomic_io import atomic_write_json
# 1. Round-trip dict round-trips identically (assert equality)
# 2. Non-ASCII payload {"jp": "<JP characters>"} written with ensure_ascii=False:
#    raw bytes contain UTF-8 sequences, no \uXXXX escapes
# 3. After successful write, target dir contains only out.json (no stray .tmp-*)
# 4. Trigger TypeError mid-write (non-serializable value): existing target byte-for-byte
#    unchanged, no stray temp file left behind
# 5. inspect.getsource confirms tempfile.mkstemp(dir=directory) + os.replace pattern
```

Output:

```
OK round-trip
OK non-ASCII preserved as UTF-8 (no \uXXXX escapes)
Files in target dir: ['out.json']
OK no stray temp files
OK target preserved on failure, no stray temp file
OK helper uses tempfile.mkstemp(dir=...) + os.replace
```

Outcome:

- [x] Round-trip preserves data ✅
- [x] `ensure_ascii=False` non-ASCII round-trips as UTF-8 (no `\uXXXX` escapes) ✅
- [x] Mid-write failure leaves existing target intact, no stray `.tmp-*` files ✅
- [x] Same-directory temp + `os.replace` pattern verified by reading the helper source ✅
- [ ] Did not run a real `openant scan` and Ctrl+C-mid-write — would have required a paid LLM call. The unit-level checks above + the in-tree tests (`tests/test_atomic_io.py` per the diff) cover the same code path.
- [ ] POSIX umask path: I'm on Windows so the `os.name == "posix"` chmod branch wasn't exercised; no `os.chmod` is invoked on Windows by design (the helper notes this).

ar7casper:
Hey @joshbouncesecurity — quick one before going deeper on this.

Don't we already get this for free from the checkpoint/resume system? Per the [CHANGELOG `[2026-04-14]`](../blob/master/CHANGELOG.md#2026-04-14--initial-public-release), every phase persists per-unit progress, and as of current master those checkpoints are **preserved** rather than cleaned up (see comments in `core/analyzer.py:518` and `core/verifier.py:196`). So a torn write to `results.json` is recoverable — rerun the step, checkpoints get reloaded, file gets re-emitted, no LLM cost lost.

The one place where atomic writes would clearly still matter is the un-checkpointed outputs (`pipeline_output.json` in `core/reporter.py`, parser outputs like `dataset.json`) — those are one-shot writes with no backstop. But this PR doesn't touch those.

What's your take — close this, or is there a scenario I'm missing?

joshbouncesecurity:
Good analysis. To address both parts:

**Checkpointed outputs** (`results.json`, `enhanced_dataset.json`, `results_verified.json`): you're right that checkpoints prevent cost loss. The only remaining value here is UX — a torn write produces an opaque JSON decode error on the next run rather than a clean resume from the intact previous file. Minor, but real.

**Un-checkpointed outputs**: commit 3 (pushed after your comment) covers exactly the gap you identified — `dataset.json`, `analyzer_output.json`, and `pipeline_output.json` are now all atomic. Those are the strong case and they're closed.

For reference, the relevant diffs:
- Commits 1–2 (checkpointed outputs + helper): https://github.com/knostic/OpenAnt/compare/42fbb32~1...eefafbd
- Commit 3 (un-checkpointed outputs): https://github.com/knostic/OpenAnt/commit/f482e09

Happy to drop commits 1–2 and keep only commit 3 if you'd prefer a tighter diff with no ambiguity about justification. Let me know.

ar7casper:
Since much of the goals of this PR are already on master, it was decided with @joshbouncesecurity to close this PR, feel the current solution on master, and open a new PR in the future with proper adjustments.

PR #45, opened 2026-05-04
fix: centralize UTF-8 file I/O for Windows compatibility

joshbouncesecurity:
## Summary

Bare `open()` calls use the system encoding (cp1252 on Windows), causing `charmap codec can't decode` errors on any target codebase containing non-ASCII characters (e.g. curly quotes U+2019, accented characters, CJK).

Adds `utilities/file_io.py` with centralized UTF-8 helpers (`open_utf8`, `read_json`, `write_json`, `run_utf8`) and migrates ~190 bare `open()` call sites across `libs/openant-core/` (core, parsers, utilities, openant CLI, top-level scripts) to specify `encoding="utf-8"` explicitly. Also sets `encoding`/`errors` on the docker_executor `subprocess.run` that captures container stdout/stderr as text.

A regression test scans non-test Python files for any bare `open()` call without an `encoding=` argument and fails if a regression reappears.

Addresses item 9 from #16 (does not close the issue).

## Test plan

- [x] Helper unit tests: non-ASCII round-trip via `open_utf8`, `read_json`, `write_json`.
- [x] `run_utf8` captures non-ASCII subprocess output (incl. `universal_newlines=True` alias).
- [x] `run_utf8` only injects `encoding`/`errors` when caller asks for text mode; explicit `encoding=` is respected.
- [x] Regression: no bare `open(` in non-test code (test scans `libs/openant-core/` excluding `tests/` and the helper).
- [x] Existing tests pass: `pytest tests/` -> 96 passed, 22 skipped (env-dependent).
- [ ] Manual: run a parse against a repo with non-ASCII content on a cp1252 console -- no `charmap` errors.

### DIFF
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -341,15 +341,15 @@ def check_manual_override(repo_path: Path) -> ApplicationContext | None:
             continue
 
         try:
-            content = filepath.read_text(encoding="utf-8")
-
             if filename.endswith('.json'):
                 # Direct JSON format
-                data = json.loads(content)
+                data = read_json(filepath)
                 data['source'] = 'manual'
                 return ApplicationContext(**data)
 
-            elif filename.endswith('.md'):
+            content = filepath.read_text(encoding="utf-8")
+
+            if filename.endswith('.md'):
                 # Markdown format - check for JSON code block
                 json_match = re.search(r'```json\s*(.*?)\s*```', content, re.DOTALL)
                 if json_match:
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -208,7 +208,7 @@ def gather_context_sources(repo_path: Path) -> dict[str, str]:
         filepath = repo_path / filename
         if filepath.exists():
             try:
-                content = filepath.read_text(errors="ignore")
+                content = filepath.read_text(encoding="utf-8", errors="ignore")
                 # Limit size to avoid token overflow
                 if len(content) > 10000:
                     content = content[:10000] + "\n\n[... truncated ...]"
@@ -289,7 +289,7 @@ def detect_entry_points(repo_path: Path) -> str:
             continue
 
         try:
-            content = py_file.read_text(errors="ignore")
+            content = py_file.read_text(encoding="utf-8", errors="ignore")
             rel_path = py_file.relative_to(repo_path)
 
             for category, patterns in ENTRY_POINT_PATTERNS.items():
@@ -308,7 +308,7 @@ def detect_entry_points(repo_path: Path) -> str:
             continue
 
         try:
-            content = js_file.read_text(errors="ignore")
+            content = js_file.read_text(encoding="utf-8", errors="ignore")
             rel_path = js_file.relative_to(repo_path)
 
             if re.search(r"express\(\)|require\(['\"]express['\"]\)", content):
@@ -340,7 +340,7 @@ def check_manual_override(repo_path: Path) -> ApplicationContext | None:
             continue
 
         try:
-            content = filepath.read_text()
+            content = filepath.read_text(encoding="utf-8")
 
             if filename.endswith('.json'):
                 # Direct JSON format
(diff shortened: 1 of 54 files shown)


joshbouncesecurity:
## Manual verification

- [ ] **Windows cp1252 console**: scan a target repo containing non-ASCII content (e.g., Japanese strings, accented chars in comments). Confirm no `UnicodeDecodeError` / `charmap codec can't decode` errors.
- [ ] **Linux/macOS**: same scan succeeds (no regression).
- [ ] **Subprocess paths**: docker_executor capturing non-ASCII output: no decode crash.
- [ ] **Path-method I/O**: any `Path.read_text()` / `write_text()` calls covered by the round-2 follow-up still work cross-platform.
- [ ] `git grep 'open(' libs/openant-core --include='*.py'` (excluding tests + the helper module): no bare `open(` calls remain. The new regression test enforces this in CI.
- [ ] **JSON round-trip**: `utilities/file_io.py write_json` then `read_json` of a dict with `{"name": "héllo 日本語"}` returns the same dict (not escaped sequences).

joshbouncesecurity:
## Local test results

Verified the migration on Windows by static-grepping the worktree and round-tripping non-ASCII content through `utilities.file_io`.

Commands run:

```
# 1. No bare open() in non-test code
git grep -nE "open\([^)]*\)" -- 'libs/openant-core/**/*.py' \
    ':!libs/openant-core/tests/**' ':!libs/openant-core/utilities/file_io.py' \
    | grep -v "encoding=" | grep -v "open_utf8"
# Two hits, both legitimate:
#   parsers/zig/function_extractor.py:43:  with open(full_path, "rb") as f:   <- binary mode, no encoding needed
#   utilities/agentic_enhancer/prompts.py:68:  - Files: open() with external paths   <- text inside a docstring/prompt

# 2. Round-trip via utilities.file_io
python check45.py
```

Output:

```
OK default round-trip preserves non-ASCII via \uXXXX escapes
OK ensure_ascii=False writes raw UTF-8 bytes; read_json still round-trips
OK open_utf8 round-trips curly quote on this Windows shell
```

Outcome:

- [x] No bare `open()` in non-test code (only legitimate `"rb"` and a docstring fragment) ✅
- [x] `read_json` / `write_json` round-trip a dict with `{"name": "héllo 日本語", "list": [..., "’"]}` byte-identically ✅
- [x] `write_json(..., ensure_ascii=False)` writes raw UTF-8 bytes (verified `e6 97 a5 e6 9c ac e8 aa 9e` for the JP run); `read_json` still parses it back to the same dict ✅
- [x] `open_utf8` round-trips a U+2019 curly quote on this Windows shell (default cp1252 console) ✅
- [ ] Did not run a full `openant scan` against a non-ASCII codebase end-to-end — would have required an LLM call. The above confirms the helper works on Windows; the migration's call-site coverage is enforced by the new regression test.

ar7casper (commented):
One thing I'd like your call on before merging: `utilities/file_io.py` defines `open_utf8` /
   `read_json` / `write_json` / `run_utf8`, but the migration inlines `encoding="utf-8"` at
  every call site rather than routing through these helpers. So as it stands, the helper
  module is only imported by its own test file — production code never touches it.

  Three ways to go:

  1. **Adopt the helpers in this PR** — single chokepoint to change later (e.g. if we ever
  want to switch to `errors="strict"` or add logging). Bigger churn though, and it's literally
   one keyword argument we'd be hiding.
  2. **Keep the helpers as the recommended pattern for new code** — current state. Fine, but
  they'll quietly rot unless we tell contributors to reach for them. Worth a one-liner in
  `libs/openant-core/CLAUDE.md` if you go this way.
  3. **Drop the helpers** and rely on inlined `encoding="utf-8"` + the regression scanners as
  the contract. Less surface area, one less thing to import.

joshbouncesecurity (commented):
Went with option 1. All production code (and the test pipelines) now routes through the helpers — `read_json`, `write_json`, `open_utf8`, and `run_utf8` — rather than inlining `encoding="utf-8"` at every call site. The four regression scanner tests in `test_file_io.py` enforce this going forward and will catch any backslide on bare `open()`, `Path.read_text/write_text`, or `subprocess.run(text=True)` without the helpers.

The only remaining `encoding="utf-8"` references in production code are calls that also pass `errors="ignore"` or `errors="replace"` to read arbitrary source files — `open_utf8` passes `**kwargs` through to `open()`, so those use `open_utf8(path, errors="replace")` etc. rather than `read_text`.

joshbouncesecurity:
@ar7casper I think this is ok now

PR #53, opened 2026-05-05
feat: add --staged diff scope for pre-commit hook scanning (#18)

### LINKED ISSUE #18: [feat] diff based analysis
This project gave me a great idea:

* https://securitydepth.hashnode.dev/diffguard-llm-powered-security-review-for-staged-git-diffs#bugs-found-along-the-way
* https://github.com/nick79/diffguard

Instead of taking units from the whole codebase and looking at reachability, can you just take units that are modified by a diff and then scan those to perform an incremental scan?

ChrisJr404:
## What

Adds a `--staged` flag to `openant scan` and `openant diff` that runs the
incremental pipeline against the staged index (`git diff --cached`) instead
of a committed ref. This is the workflow the [DiffGuard][1] project from #18
is built around: scan what is about to be committed, before the commit lands.

[1]: https://securitydepth.hashnode.dev/diffguard-llm-powered-security-review-for-staged-git-diffs

The existing `--diff-base` / `--pr` / `--incremental` modes already cover
post-commit and PR-time scanning. `--staged` fills the pre-commit gap, which
makes the diff-mode pipeline usable from a `pre-commit` hook or a quick local
"scan what I'm about to commit" run without first committing.

## Usage

```bash
# Pre-commit hook style: scan only staged changes against HEAD.
git add path/to/file.py
openant scan --staged --skip-dynamic-test

# Same thing via the shorter diff verb.
openant diff --staged
```

In a `.git/hooks/pre-commit`:

```sh
#!/bin/sh
exec openant diff --staged --skip-dynamic-test --output .openant/staged
```

## How

- New `BuildStagedManifest`, `StagedChangedFiles`, `StagedHunksForFile` in
  `internal/git/diff.go`. Same `Manifest` shape as the committed-diff path,
  with `base_ref="STAGED"`, `base_sha=HEAD`, and a zero placeholder for
  `head_sha` (the index has no SHA).
- The Python core reads these fields as opaque strings, so reports surface
  `base=STAGED` cleanly without any change in `libs/openant-core`.
- `diffOpts` grows a `staged bool`; `validate()` makes `--staged` mutually
  exclusive with `--diff-base` / `--pr`.
- `modeOpts` / `modeDecision` thread the staged choice through `selectMode`,
  so `scan`, `diff`, and the project `meta.json` running state agree on the
  scan kind.
- README gets a short "Incremental and diff-based scans" section listing the
  four scopes (`--diff-base`, `--pr`, `--staged`, `--incremental`).

## Notes

Untracked files are not in the index, so they do not appear in the staged
manifest. Pre-commit hooks typically `git add` first; not second-guessing
that contract here.

`git diff --cached --unified=0` is the same hunk format the existing
committed-diff path consumes, so the rest of the pipeline (parser tagging,
`diff_filter`, reporter) needs no changes.

## Tests

10 new tests, all passing:

- `internal/git`: `TestStagedChangedFilesAndHunks` (stages two files plus a
  worktree-only edit, asserts `--cached` does not bleed),
  `TestBuildStagedManifest`, `TestBuildStagedManifestChangedFilesScopeOmitsHunks`,
  `TestBuildStagedManifestEmptyIndex`, `TestBuildStagedManifestRejectsBadScope`.
- `cmd`: `TestValidateModeFlags` gains five rows for the new conflict combos
  (`staged + diffBase`, `staged + pr`, `staged + incremental`, `full + staged`,
  `staged alone`); `TestSelectModeStagedSetsStagedFlag`.
- `cmd/diff_shared_test.go`: `TestDiffOptsValidateStaged` (7 sub-cases) and
  `TestDiffOptsIsSetStaged`.

`go vet ./...` clean. `go test ./...` green on all packages.

Closes #18.

### DIFF
--- a/apps/openant-cli/cmd/diff_shared.go
+++ b/apps/openant-cli/cmd/diff_shared.go
@@ -9,22 +9,33 @@ import (
 	"github.com/knostic/open-ant-cli/internal/output"
 )
 
-// diffOpts collects the three diff-mode flags that scan/parse/diff all share.
+// diffOpts collects the diff-mode flags that scan/parse/diff all share.
 type diffOpts struct {
 	base   string
 	pr     int
+	staged bool
 	scope  string
 }
 
 // isSet reports whether any diff flag was provided.
 func (o diffOpts) isSet() bool {
-	return o.base != "" || o.pr > 0
+	return o.base != "" || o.pr > 0 || o.staged
 }
 
 // validate enforces flag rules common to all entry points.
 func (o diffOpts) validate() error {
-	if o.base != "" && o.pr > 0 {
-		return fmt.Errorf("--diff-base and --pr are mutually exclusive")
+	set := 0
+	if o.base != "" {
+		set++
+	}
+	if o.pr > 0 {
+		set++
+	}
+	if o.staged {
+		set++
+	}
+	if set > 1 {
+		return fmt.Errorf("--diff-base, --pr, and --staged are mutually exclusive")
 	}
 	if o.isSet() {
 		if o.scope == "" {
@@ -60,21 +71,30 @@ func prepareDiffManifest(repoPath, outputDir string, opts diffOpts) (string, err
 		return "", fmt.Errorf("create output dir %s: %w", outputDir, err)
 	}
 
-	baseRef := opts.base
-	if opts.pr > 0 {
-		fetched, err := git.FetchPR(repoPath, opts.pr, nil)
+	var m *git.Manifest
+	if opts.staged {
+		built, err := git.BuildStagedManifest(repoPath, opts.scope)
 		if err != nil {
-			return "", err
+			return "", fmt.Errorf("build staged diff manifest: %w", err)
 		}
-		baseRef = fetched
-		if !quiet {
-			fmt.Fprintf(os.Stderr, "PR #%d: base=%s (fetched and checked out pr-head)\n", opts.pr, baseRef)
+		m = built
+	} else {
+		baseRef := opts.base
+		if opts.pr > 0 {
+			fetched, err := git.FetchPR(repoPath, opts.pr, nil)
+			if err != nil {
+				return "", err
+			}
+			baseRef = fetched
+			if !quiet {
+				fmt.Fprintf(os.Stderr, "PR #%d: base=%s (fetched and checked out pr-head)\n", opts.pr, baseRef)
+			}
 		}
-	}
-
-	m, err := git.BuildManifest(repoPath, baseRef, opts.scope, opts.pr)
-	if err != nil {
-		return "", fmt.Errorf("build diff manifest: %w", err)
+		built, err := git.BuildManifest(repoPath, baseRef, opts.scope, opts.pr)
+		if err != nil {
+			return "", fmt.Errorf("build diff manifest: %w", err)
+		}
+		m = built
 	}
 
 	manifestPath := filepath.Join(outputDir, "diff_manifest.json")
@@ -83,8 +103,13 @@ func prepareDiffManifest(repoPath, outputDir string, opts diffOpts) (string, err
 	}
 
 	if !quiet {
-		output.PrintKeyValue("Diff base", fmt.Sprintf("%s (%s)", m.BaseRef, shortSHA(m.BaseSHA)))
-		output.PrintKeyValue("Diff head", shortSHA(m.HeadSHA))
+		if m.BaseRef == git.StagedRef {
+			output.PrintKeyValue("Diff base", fmt.Sprintf("HEAD (%s)", shortSHA(m.BaseSHA)))
+			output.PrintKeyValue("Diff head", "index (staged)")
+		} else {
+			output.PrintKeyValue("Diff base", fmt.Sprintf("%s (%s)", m.BaseRef, shortSHA(m.BaseSHA)))
+			output.PrintKeyValue("Diff head", shortSHA(m.HeadSHA))
+		}
 		output.PrintKeyValue("Diff scope", m.Scope)
 		output.PrintKeyValue("Changed files", fmt.Sprintf("%d", len(m.ChangedFiles)))
 	}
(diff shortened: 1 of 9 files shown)


apps/openant-cli/cmd/diff_shared.go
```diff
@@ -83,8 +103,13 @@ func prepareDiffManifest(repoPath, outputDir string, opts diffOpts) (string, err
 	}
 
 	if !quiet {
-		output.PrintKeyValue("Diff base", fmt.Sprintf("%s (%s)", m.BaseRef, shortSHA(m.BaseSHA)))
-		output.PrintKeyValue("Diff head", shortSHA(m.HeadSHA))
+		if m.BaseRef == git.StagedRef {
```
ar7casper, line 106: **🟡 Medium** — UX · Python report consumers don't mirror this STAGED special-case

**Issue:** Nice that the CLI output renders `Diff head: index (staged)` for staged manifests. But the same `head_sha = 0000…0000` placeholder flows into `libs/openant-core/generate_report.py:109-121` (`_build_header_subtitle`) and `libs/openant-core/export_csv.py:54-65` (`_format_diff_banner`), both of which call `_short_sha(head_sha)` → `00000000`. End-users will see `Incremental {repo} · 5e1d2a4..00000000 · …` in the HTML report and CSV banner.

**Suggestion:** Mirror this special-case on the Python side — a one-line guard `if base_ref == "STAGED": head = "index"` in both `_build_header_subtitle` and `_format_diff_banner` keeps the report symmetric with the CLI output.

apps/openant-cli/cmd/scan.go
```diff
@@ -132,6 +134,7 @@ func runScan(cmd *cobra.Command, args []string) {
 	if decision.Kind == config.ScanKindDiff {
 		manifestOpts.base = decision.Base
 		manifestOpts.scope = decision.Scope
+		manifestOpts.staged = decision.Staged
```
ar7casper, line 137: **🟡 Medium** — Architecture · `Staged` is consumed here but never persisted to `ScanMeta`

**Issue:** `decision.Staged` flows into `manifestOpts.staged` for *this* run, but the meta written a few lines down at `scan.go:300-311` only persists `Kind=diff, Base="", Scope=...` — the staged marker is lost. Two consequences:

1. **Crashed-resume silently degrades to full.** If a staged scan writes `meta.status=running` and then crashes, a subsequent `openant scan` (no flags) hits the resume path at `scan.go:272-276`, which returns `modeDecision{Kind: "diff", Base: "", Scope: ...}` (`Staged` defaults to `false`). In `runScan`, `prepareDiffManifest` then sees an unset `diffOpts`, `isSet()` returns false, and the diff manifest is silently skipped — falling back to a full scan without telling the user.
2. **Audit trail is wrong.** Terminal meta after a successful staged scan is indistinguishable from a malformed `--diff-base` run with empty base.

**Suggestion:** Either (a) add a `Staged bool` field (with a `staged` json tag, omitempty) to `ScanMeta` and propagate through both the writer and the resume path, or (b) skip persisting meta for staged runs entirely (since staged is meant to be ephemeral). (b) is simpler and matches the "pre-commit ephemeral" framing in the description.

gadievron:
Thanks @ChrisJr404 — `--staged` for pre-commit scanning is a genuinely useful scope, and the diff-plumbing (`StagedChangedFiles`/`StagedHunksForFile`/`BuildStagedManifest`) is clean and well-tested. Two things to fold in before merge:

**1. Rebase (currently CONFLICTING vs master `e1f0401`) — exactly 2 conflicts, both mechanical:**
- `README.md` (~L195): keep-both — master added a section at the same anchor where this PR inserts the `--staged` docs.
- `apps/openant-cli/cmd/scan.go` (the flag `var` block): take master's block (it grew `scanLLMReachability`, `scanLLMReachabilityMaxCodeBytes`, `scanLibraryMode`) and add your `scanStaged bool`. **One decision for you:** your block declares `scanModel string`, but master has since renamed that flag to `scanLLMConfig` — so drop `scanModel` in favor of master's `scanLLMConfig` (unless `--model` was meant as a genuinely separate flag, in which case say so). All 7 other files auto-merge cleanly.

**2. Correctness — `--name-only` without `-z` silently drops non-ASCII/space/newline paths (a false negative, which matters most in a SAST tool).** `StagedChangedFiles` (and `StagedHunksForFile`) split `git diff --cached --name-only` on newlines, but git C-quotes such paths by default. Demonstrated on a real repo:

```
# current (no -z):
  path='"caf\303\251.py"'   openable=NO  -> file SKIPPED from the scan (FN)
  path='normal.py'          openable=yes
# with -z (NUL-separated, no quotepath):
  path='café.py'            openable=yes
  path='normal.py'          openable=yes
```

Fix: add `-z` to the git args and split on `\0` (`bytes.Split(out, []byte{0})`) instead of `bufio.Scanner`. e.g.:
```go
cmd := exec.Command("git", "-C", repoPath, "diff", "--cached", "--name-only", "--find-renames", "-z")
// ...
for _, b := range bytes.Split(out, []byte{0}) {
    if f := string(b); f != "" { files = append(files, f) }
}
```

**Note (not your bug):** the pre-existing committed-diff `ChangedFiles` has the *same* latent issue (also no `-z`) — worth fixing the shared pattern in the same PR, or filing a follow-up.

After the rebase + `-z`, this triggers the current Go+Python CI (which hasn't run since your branch predates the current `test.yaml`). Nice feature — full credit is yours; happy to help if any of the above is unclear.

gadievron:
Re-reviewed against current `master` — this is now correct and I have nothing to add. Thanks for applying the `-z` fix, @ChrisJr404: both `ChangedFiles` and `StagedChangedFiles` use `git diff … --name-only -z … ` with NUL-splitting, which closes the silent false-negative where a non-ASCII staged filename was `core.quotepath`-escaped and dropped from the scan. Verified end-to-end: a staged `café.py` shows as `"caf\303\251.py"` under the default output but as the exact `café.py` under `-z` (what this PR uses), so it's correctly included. Builds clean, and GitHub now reports it MERGEABLE (was CONFLICTING earlier). LGTM from my side — ready for maintainer review + merge. No replacement PR needed; this is your work and it's right.

PR #57, opened 2026-05-10
fix: correct venv Python path detection on Windows

joshbouncesecurity:
The managed venv path was hardcoded to Unix-style bin/python, which fails on Windows where venvs use Scripts/python.exe. Now detects the correct path based on runtime.GOOS.

Also adds comprehensive tests for venvPython() to catch platform-specific path issues.


joshbouncesecurity:
I tested this branch doing init and parse starting with an empty venv, I don't have spare Claude capacity to do a full end to end test right now

PR #58, opened 2026-05-12
feat(Release): 2026-05-12

### LINKED ISSUE #6: JS/TS parser fails on fresh install: missing npm dependencies

### LINKED ISSUE #21: [bug] JavaScript parser misses Express.js anonymous route handler callbacks

ar7casper:
## Summary

Release bundle for 2026-05-12. Combines seven independently-reviewed PRs into one RC for cleaner master history. Scope is parser depth (TypeScript DI, Express anonymous handlers), dependency UX (auto-reinstall on `pyproject.toml` change, JS parser bootstrap), CLI consistency (`parse` default), and a new opt-in LLM reachability stage.

### Included PRs

- **#35 — `fix: default parse --level to reachable to match scan and Python CLI`**
  Brings the Go CLI's `parse` command into alignment with `scan` and the Python CLI, both of which already defaulted to `reachable`. The documentation has always said the default is `reachable` — PR 35 makes the code match the documented contract.

- **#36 — `feat: auto-detect dependency changes and reinstall openant`**
  SHA-256 hash of `pyproject.toml` stored at `~/.openant/venv/.deps-hash`. Every CLI invocation compares the stored hash and re-runs `pip install -e <core>` when they differ. Catches stale-venv after `git pull`.

- **#37 — `fix: lazy-install JS parser npm deps on first use`**
  `openant parse` on a JS/TS repo no longer fails with `Cannot find module 'ts-morph'`. Auto-runs `npm install` once on first JS parse with `node_modules/.package-lock.json` as the completion sentinel. Cross-platform file lock prevents concurrent install corruption. Closes #6.

- **#39 — `feat: DI-aware call resolution with nominal type matching for TypeScript/NestJS`**
  Resolves `this.userService.findById()` style calls in NestJS / Angular codebases. Covers constructor injection, field-decorator injection (`@Inject` / `@InjectRepository` / etc.), and Angular's functional `inject()` API. Resolution priority: exact type → nominal (`implements`/`extends`) → unambiguous prefix. All steps return `null` on ambiguity. Class-level metadata file-qualified by `relativePath:className` for multi-module monorepos.

- **#40 — `feat: auto-detect language in init`**
  `openant init` now works on non-git directories (`commit_sha = "nogit"` placeholder). Shared `config/languages.json` consumed by both Go CLI and Python parser adapter (eliminates Go↔Python extension-list drift). Language auto-detection exposed as opt-in via `-l auto` (experimental dominance heuristic; see #61 for the validation work needed before it becomes the default).

- **#49 — `fix: extract Express.js anonymous route handler callbacks`**
  `router.post('/x', auth, async (req, res) => {...})` style handlers are now extracted as units. Synth units carry `route_handler` (last callback) or `route_middleware` (earlier callbacks); both registered in `ENTRY_POINT_TYPES` so the reachability filter doesn't drop them. Receiver filter prevents false positives on cache/query-builder `.get/.post(...)` calls. Named middleware identifiers become call-graph edges so `authenticateToken` shows up as an upstream dependency. Closes #21.

- **#50 — `feat: LLM review stage for enhanced reachability detection`**
  New opt-in `--llm-reachability` pass that uses Opus to surface entry points the structural analysis misses (framework handlers, plugin/CLI registrations, message queues, external input sites). Promote-only semantics — never demotes structurally-detected units. Bundle includes `call_graph.json` being written by all 5 previously-missing parsers (C/Ruby/PHP/JS/Go) so the post-LLM re-filter works across languages.

### Changelog

`CHANGELOG.md` has a top entry for `[2026-05-12] — Parser depth, dependency UX, and LLM reachability (opt-in)` covering the user-facing impact of each PR. _(Not yet committed — landing in a separate commit before merge.)_

## Test plan

- [x] Each constituent PR's test suite passes individually (verified at merge time).
- [x] Combined CI green on Ubuntu / macOS / Windows.
- [ ] Manual: end-to-end scan across each supported language to confirm DI resolution + Express handler extraction + LLM reachability flow.

Closes #6
Closes #21

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -52,6 +52,12 @@ jobs:
           cache: "npm"
           cache-dependency-path: libs/openant-core/parsers/javascript/package-lock.json
 
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: libs/openant-core/parsers/go/go_parser/go.mod
+          cache-dependency-path: libs/openant-core/parsers/go/go_parser/go.mod
+
       - name: Install Python dependencies
         working-directory: libs/openant-core
         run: pip install -r requirements.txt && pip install ".[dev]"
@@ -67,6 +73,16 @@ jobs:
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         working-directory: libs/openant-core/parsers/javascript
         run: npm ci
+
+      - name: Build go_parser binary (Linux/macOS)
+        if: runner.os != 'Windows'
+        working-directory: libs/openant-core/parsers/go/go_parser
+        run: go build -o go_parser .
+
+      - name: Build go_parser binary (Windows)
+        if: runner.os == 'Windows'
+        working-directory: libs/openant-core/parsers/go/go_parser
+        run: go build -o go_parser.exe .
 
       - name: Run Python and parser tests
         working-directory: libs/openant-core
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,6 @@ __pycache__/
 node_modules/
 apps/openant-cli/bin/
 libs/openant-core/parsers/go/go_parser/go_parser
+libs/openant-core/parsers/javascript/.openant-npm-install.lock
 _docs/
+
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ apps/openant-cli/bin/
 libs/openant-core/parsers/go/go_parser/go_parser
 libs/openant-core/parsers/javascript/.openant-npm-install.lock
 _docs/
-
+docs/
+.worktrees/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,101 @@
 
 All notable changes to OpenAnt are documented in this file.
 
+## [2026-05-12] — Parser depth, dependency UX, and LLM reachability (opt-in)
+
+### Fixed
+
+- **`openant parse` now defaults `--level` to `reachable`.** The Go CLI's
+  `parse` command previously defaulted to `--level all`, contradicting
+  `scan` and the Python CLI which both default to `reachable`. The
+  documentation has always said the default is `reachable`. Anyone running
+  `openant parse <repo>` without `-l` now gets the same dataset as
+  `openant scan <repo> --steps parse` — the documented behavior. Set
+  `--level all` explicitly to restore the previous output. (#35)
+
+- **JS parser dependencies are now auto-installed on first use.**
+  `openant parse` on a JavaScript/TypeScript repository previously failed
+  out of the box with `Cannot find module 'ts-morph'` because nothing in
+  the install flow ran `npm install` for `parsers/javascript/`. The Python
+  parser adapter now runs `npm install` once on first JS parse using
+  `node_modules/.package-lock.json` as the completion sentinel (catches
+  Ctrl+C-interrupted installs). Python/Go-only users still never need
+  `npm`. Includes a cross-platform file lock to prevent concurrent install
+  corruption. Closes #6. (#37)
+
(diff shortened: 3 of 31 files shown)


ar7casper:
closes #6

shahar-davidson (approved):

PR #60, opened 2026-05-12
feat: DI-aware forward tracing and get_static_dependencies tool for agentic enhancer

joshbouncesecurity:
## Context

Split from #39 per reviewer request — the agentic enhancer changes have a different risk profile from the parser changes and warrant a separate eval pass before merging.

## Summary

- Adds `get_static_dependencies` tool to the agentic enhancer's tool set, so the exploration agent can retrieve statically-resolved callee and caller IDs and examine them with `read_function`
- Adds `resolve_dependencies()` to `RepositoryIndex` to map raw function IDs / qualified names back to function metadata
- Adds `set_unit_context()` to `ToolExecutor` so per-unit static dep data is available when the tool is called
- Updates the enhancer prompt to instruct the agent to trace forward into called functions (service/repository methods) and check for delegated authorization or validation before classifying a finding as EXPLOITABLE

## Why independent of #39

The parser DI changes (#39) improve call graph edges regardless of how the LLM consumes them. This PR changes what the LLM does with those edges (and with the existing call graph for all languages). The classification impact — particularly the new "delegated security controls" branch — needs a before/after eval on a known dataset before hitting production.

## Test plan

- [ ] `test_enhancer_tools.py`: `resolve_dependencies`, `get_static_dependencies` via `ToolExecutor`, context reset between units
- [ ] Before/after eval comparing verdict distributions on a representative scan corpus (EXPLOITABLE / VULNERABLE_INTERNAL / NEUTRAL / SECURITY_CONTROL rates)

### DIFF
--- a/libs/openant-core/tests/test_enhancer_tools.py
+++ b/libs/openant-core/tests/test_enhancer_tools.py
@@ -1,0 +1,127 @@
+"""Tests for the agentic enhancer tools, specifically the get_static_dependencies tool."""
+import pytest
+
+from utilities.agentic_enhancer.repository_index import RepositoryIndex
+from utilities.agentic_enhancer.tools import ToolExecutor
+
+
+def _make_index(functions: dict) -> RepositoryIndex:
+    """Create a RepositoryIndex from a minimal functions dict."""
+    return RepositoryIndex({"functions": functions})
+
+
+SAMPLE_FUNCTIONS = {
+    "src/user.controller.ts:UserController.getUser": {
+        "name": "UserController.getUser",
+        "code": "async getUser(id) { return this.userService.findById(id); }",
+        "className": "UserController",
+        "unitType": "class_method",
+        "startLine": 10,
+        "endLine": 12,
+    },
+    "src/user.service.ts:UserService.findById": {
+        "name": "UserService.findById",
+        "code": "async findById(id) { return this.repo.findOne(id); }",
+        "className": "UserService",
+        "unitType": "class_method",
+        "startLine": 5,
+        "endLine": 7,
+    },
+    "src/auth.guard.ts:AuthGuard.canActivate": {
+        "name": "AuthGuard.canActivate",
+        "code": "canActivate(context) { return this.validate(context); }",
+        "className": "AuthGuard",
+        "unitType": "class_method",
+        "startLine": 3,
+        "endLine": 5,
+    },
+}
+
+
+class TestResolveDependencies:
+    """Test RepositoryIndex.resolve_dependencies."""
+
+    def test_resolves_by_function_id(self):
+        index = _make_index(SAMPLE_FUNCTIONS)
+        result = index.resolve_dependencies([
+            "src/user.service.ts:UserService.findById"
+        ])
+        assert len(result) == 1
+        assert result[0]["id"] == "src/user.service.ts:UserService.findById"
+        assert result[0]["className"] == "UserService"
+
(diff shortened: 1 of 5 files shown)


gadievron:
Thanks @joshbouncesecurity — DI-aware forward tracing + a `get_static_dependencies` tool is a genuinely valuable capability, and it's cleanly scoped to `utilities/agentic_enhancer/*` (no collision with anything merged; `get_static_dependencies`/`RepositoryIndex.resolve_dependencies` are absent from master, so it's not superseded). A few things before it's ready:

**1. It's still a draft** — mark ready-for-review when you'd like it looked at (both reviewers initially missed the draft flag).

**2. Encoding regression (deterministic).** In `repository_index.py` the diff replaces:
```python
-from utilities.file_io import read_json
-    analyzer_output = read_json(analyzer_output_path)
+    with open(analyzer_output_path, 'r') as f:
+        analyzer_output = json.load(f)
```
`read_json` is `open_utf8(path, "r")` — it **forces UTF-8**. A bare `open(path, 'r')` uses the platform default encoding, so on Windows (cp1252) a UTF-8 `analyzer_output.json` containing any non-ASCII byte fails or mis-decodes. Please revert to `read_json(analyzer_output_path)` (restore the import). (We just hit the analogous Windows-encoding class on another PR's CI — it only fails on the windows-latest runner.)

**3. FN-safety of the verdict nudge (the important design point).** The forward-tracing enhancer can downgrade `EXPLOITABLE → SECURITY_CONTROL` when it sees a guard on the path. In a reachability-SAST tool a downgrade is the dangerous direction — a mis-identified/no-op `authorize()`-style callee would suppress a real finding (false negative). Consider a **dominance/soundness guard**: only downgrade when the control provably dominates every path to the sink (and ideally leave a downgrade-to-review rather than downgrade-to-safe). Same spirit for the `search_by_name(exact=…)`/bare-name fallback — a bare-name match across modules can bind the wrong symbol.

**4. CI** is red/expired on all three OSes on the current head — a rebase + push will re-run it.

Really nice direction — full credit is yours. Happy to help with the dominance-guard shape if useful.

gadievron:
Thanks @joshbouncesecurity — I've opened #259, a fork-minimal replacement that carries your `get_static_dependencies` tool + DI-aware forward tracing **verbatim** (your original commit preserved, authorship intact) rebased onto current `master`. It drops two things that trade an FP for an FN (the cardinal sin for a reachability tool): (1) the classification prompt let a *reachable* dangerous sink be downgraded to `SECURITY_CONTROL`/lower when a callee "has controls" — but a callee control doesn't prove the tainted path is guarded, so a real EXPLOITABLE could be hidden; #259 keeps the forward-tracing as context with an over-seed guard (lower severity only if a callee *provably* guards the tainted path). (2) the `read_json`→bare `open()` swap dropped UTF-8 forcing (UnicodeDecodeError under a non-UTF-8 locale) — restored. Full credit to you for the capability. Recommend landing #259 and closing this in its favor.

gadievron:
Closing as superseded — your DI-aware forward tracing + `get_static_dependencies` tool shipped in #259 (now merged to `master`), with you preserved as co-author on the commit. #259 kept your capability and dropped two things that traded a false-positive for a false-negative in a reachability tool (the classification prompt downgrading a reachable sink to SECURITY_CONTROL, and a read_json UTF-8 regression). Thanks for the contribution, @joshbouncesecurity — it's live on master.

PR #72, opened 2026-05-28
fix(parsers/python): repair 6 call-graph edge-fidelity bugs in the Python CallGraphBuilder

gadievron:
Six independent defects in libs/openant-core/parsers/python/call_graph_builder.py, each
pinned by a RED->GREEN regression test. All fixes are Python-AST-native (ast.Call,
ast.Assign, ast.Name, the extractor's classes[...]['bases'] model); no code or tokens
copied from the sibling go/php/ruby/zig builders, which are untouched.

- Package __init__.py re-export resolution: _resolve_via_package_init follows names
  re-exported through a package __init__.py to their origin module (was: no __init__.py
  probing -> dropped edge).
- Local-variable type dispatch: _collect_local_types + _resolve_class_method dispatch
  `v = ClassName(); v.method()` to the constructed type's method (was: routed to
  _resolve_module_call, which knows only import/same-file class names -> dropped edge).
- super() resolution: _resolve_super_call walks the caller class's bases (cross-file) to
  resolve super().method() (was: stub returning None).
- Deterministic output: sort callee + reverse-caller lists on emit for reproducible output
  (was: set-iteration order, PYTHONHASHSEED-dependent).
- Higher-order callback args: _resolve_callback_args emits edges for function references
  passed as call arguments, e.g. sorted(xs, key=func) (was: only node.func inspected).

Deliberately not addressed here (documented scope boundaries, no code change):
- getattr(obj,'m')(args) is statically unresolvable string dispatch.
- A chained inner().method() outer method on a Call result is unresolvable; ast.walk
  already captures the resolvable inner call.

Tests: tests/test_python_call_graph_builder.py (unique importlib name; 5 tests covering
the 6 fixed bugs). RED 5 failed (pre-fix) -> GREEN 5 passed. Suite 181 passed / 63 skipped
/ 0 failed (base 176/63/0). Determinism verified across PYTHONHASHSEED 0/1/12345.

### DIFF
--- a/libs/openant-core/parsers/python/call_graph_builder.py
+++ b/libs/openant-core/parsers/python/call_graph_builder.py
@@ -183,17 +183,76 @@ class CallGraphBuilder:
             # Fall back to regex-based extraction
             return self._extract_calls_regex(code, caller_id)
 
+        # Local variable -> constructor-type map, so that `v = ClassName(); v.method()`
+        # dispatches to the bound type's method.
+        local_types = self._collect_local_types(tree)
+
         for node in ast.walk(tree):
             if isinstance(node, ast.Call):
-                resolved = self._resolve_call_node(node, caller_file, caller_class)
+                resolved = self._resolve_call_node(node, caller_file, caller_class, local_types)
                 if resolved:
                     calls.add(resolved)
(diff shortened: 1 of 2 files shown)


gadievron:

PR #73, opened 2026-05-28
fix(parsers): repair Ruby + PHP call-graph resolution

gadievron:
Resolution defects in the Ruby and PHP call-graph builders dropped real edges
and emitted false ones. Each parser is fixed parser-native with its own grammar;
no token was copied between them.

Ruby (parsers/ruby/call_graph_builder.py):
- Module-function resolution: add a methods_by_module index + thread caller
  module_name. Module self/sibling and Module.method calls now resolve; a bare
  call from outside any module no longer leaks an edge to a module function (the
  same-file and unique-name fallbacks now require both class_name and module_name
  unset).
- super: visit the bare `super` node and the `super(args)` call head; resolve to
  the same-named method on the superclass.
- Class.new: resolve `new` to the class initialize before the builtin filter that
  previously dropped it.
- require_relative anchoring: anchor require_relative to the caller dir and
  normalize ./ and ../; match require by anchored file name, replacing the
  unanchored substring that over-matched any path containing the import string.
- send/public_send/__send__: read the literal symbol argument and resolve the
  dispatched method.

PHP (parsers/php/call_graph_builder.py):
- Cross-namespace bare-call leak: thread the caller's namespace_name and scope the
  unique-name fallback to the caller's namespace so a bare call no longer leaks
  across namespaces. PHP-native analog of the Ruby module-leak fix (different
  grammar/fields; implemented independently).

Tests: tests/parsers/ruby/test_call_graph_builder.py (10),
tests/parsers/php/test_call_graph_builder.py (2). Both modules loaded under
unique importlib names (call_graph_builder.py / function_extractor.py are basenames
shared by every parser). 11 failed pre-fix (1 same-namespace guard green at base by
design) -> 12 passed after the fix; ruff clean.

### DIFF
--- a/libs/openant-core/parsers/php/call_graph_builder.py
+++ b/libs/openant-core/parsers/php/call_graph_builder.py
@@ -161,6 +161,7 @@ class CallGraphBuilder:
         caller_file = caller_id.split(':')[0]
         caller_func = self.functions.get(caller_id, {})
         caller_class = caller_func.get('class_name')
+        caller_namespace = caller_func.get('namespace_name')
 
         code_bytes = code.encode('utf-8', errors='replace')
         try:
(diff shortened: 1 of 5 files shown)


gadievron:
**Merge-order note** (not a defect in this PR — flagging for whoever lands the stack)

`parsers/php/call_graph_builder.py` here overlaps **#71** (same call-node dispatch in `_extract_calls_from_code`). Resolve as a **union**, not accept-ours/theirs:
- keep `'object_creation_expression'` in the handled node types (#71) — else `new X()` → `__construct` edges are dropped;
- keep passing `caller_namespace` into `_resolve_call_node`/`_resolve_function_call` (this PR).

gadievron:

PR #74, opened 2026-05-28
Fix JS DependencyResolver call-edge fidelity

gadievron:
dependency_resolver.js only. Four independent call-graph defects:

- Bare call `foo()` no longer binds to a class method `Class.foo`. _resolveCall
  now resolves only free functions (!className) at both the same-file and
  unique-simple-name sites; methods stay reachable via this./obj./Class. paths.
  Mirrors the Ruby/PHP siblings.
- `const x = new C(); x.m()` now resolves to C.m via a per-function
  _extractLocalTypes pre-pass + a local-var-type branch in _resolveMethodCall;
  built-in constructors (Map/Set/...) skipped.
- buildCallGraph drops self-edges (calledId === funcId) from the standalone-call
  regex matching a function's own name; matches the C sibling's `c != func_id`
  invariant.
- CLI JSON.parse is wrapped in try/catch emitting a readable diagnostic + exit 1
  instead of an uncaught V8 SyntaxError.

Deferred / not-a-defect (not fixed here):
- Import-based resolution needs typescript_analyzer.js to emit imports
  (cross-cutting); repoRoot is a dead field.
- eval/Function are in a call-graph noise filter, not a security sink;
  indirect_calls is a phantom field. Correct-by-design over-claim, no resolver
  defect.
- The silent-drop / dead this.imports path: no consumer for an unresolvedCalls
  channel; import disambiguation is cross-cutting.

The five funcId parse sites all use split(':')[0] (first colon), consistent with
the unit_generator.js id-parse fix; no resolver change was needed there.

Tests: tests/parsers/javascript/test_dependency_resolver_edges.py (Node harness on
the resolver + CLI subprocess). 7 failed pre-fix (3 control/guard tests green at
base) -> 10 passed after the fix. Full suite 227 passed / 22 skipped
(Go-binary, unrelated) / 0 failed.

### DIFF
--- a/libs/openant-core/parsers/javascript/dependency_resolver.js
+++ b/libs/openant-core/parsers/javascript/dependency_resolver.js
@@ -84,10 +84,15 @@ class DependencyResolver {
         }
       }
 
-      this.callGraph[funcId] = calls;
+      // Drop self-edges: the standalone-call regex matches a function's own
+      // name in its declaration/body, and recursion is not a dependency on a
+      // different unit. Mirrors the C sibling's `c != func_id` invariant.
+      const resolvedCalls = calls.filter((calledId) => calledId !== funcId);
+
+      this.callGraph[funcId] = resolvedCalls;
 
       // Build reverse graph
-      for (const calledId of calls) {
+      for (const calledId of resolvedCalls) {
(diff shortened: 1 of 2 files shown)


gadievron:

PR #75, opened 2026-05-28
fix(reachability): seed native/cross-language entry points, fix Zig reachability crash, stop silent zero-seed blackout

gadievron:
Six filed defects converge on the shared EntryPointDetector + reachability seeding
path; one is a Zig-local crash.

Central detector (utilities/agentic_enhancer/entry_point_detector.py):
- Add 'main' / 'http_handler' / 'middleware' to ENTRY_POINT_TYPES. The C/Go/Zig
  parsers emit these program/web entry unit_types, but the set only knew the
  Python/Express web vocabulary, so a compiled binary seeded zero entry points
  (total reachability blackout).
- Add _unit_type() dual-key read (snake 'unit_type' + camel 'unitType') for
  Check-1, Check-4, details and statistics. The per-parser reachable path
  normalizes under camelCase 'unitType' (parsers/{c,php,ruby}/test_pipeline.py:257)
  while the detector read snake-only, so Check-1/Check-4 were dead on that path
  even for valid entry types. (The earlier-filed file_path/filePath locus is
  phantom; the operative residual is exactly this dual-read gap.)

Zig classifier (parsers/zig/function_extractor.py):
- _classify_function: main -> 'main' (was generic 'function'), matching C/Go so a
  Zig binary's entry point is now seedable.

Zig reachability filter (parsers/zig/test_pipeline.py):
- Rewrite apply_reachability_filter to the real EntryPointDetector(functions,
  call_graph).detect_entry_points() / ReachabilityAnalyzer(functions,
  reverse_call_graph, entry_points).get_all_reachable() contract. The old code
  called a non-existent API; imports succeed (sys.path), so except ImportError
  never fired and the wrong-arity TypeError crashed every Zig parse at
  --processing-level reachable. Derived against Zig's own snake-case data shape
  (no token-copy of the C normalizer).

Empty-seed safety-net (core/parser_adapter.py):
- A zero entry-point seed previously emptied the dataset silently (100% reduction
  reported as SUCCESS) — the dominant failure for non-web library/stdlib targets.
  Degrade to pass-through (units preserved, filtering NOT applied) + record a loud
  warning in the filter metadata so the blackout can never be silent. The broader
  generic-library seeding heuristic is architectural and out of scope.

Tests (RED->GREEN): tests/test_entry_point_detector_native_seeds.py (6),
tests/parsers/zig/test_zig_main_classification.py (3),
tests/parsers/zig/test_zig_reachability_api.py (2, reproduced the exact
TypeError), tests/test_reachability_empty_seed.py (2). 12 failed pre-fix (1
guard green at base by design) -> 13 passed after the fix. Full suite 189 passed /
63 skipped / 0 failed. go test/vet/build clean in parsers/go/go_parser (types.go
unchanged, gofmt-clean). ruff clean.

### DIFF
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -262,6 +262,34 @@ def apply_reachability_filter(
     if extra_entry_points:
         entry_points = entry_points | extra_entry_points
 
+    units = dataset.get("units", [])
+    original_count = len(units)
+
+    # Empty-seed safety-net: a zero entry-point seed would prune EVERY unit
+    # (the BFS frontier starts empty), silently emptying the dataset and
+    # reporting a 100% reduction as success. That is the dominant failure mode
+    # for non-web library / stdlib targets, whose ordinary functions are not a
+    # seedable entry type. Rather than a silent total blackout, degrade to
+    # pass-through (keep all units, unfiltered) and record a loud warning so the
+    # degraded result is never silent. Higher-level callers may still seed
+    # ``extra_entry_points`` to get real filtering.
+    if not entry_points and original_count > 0:
+        warning = (
+            "No entry points detected — reachability cannot seed a frontier. "
+            "Returning all units unfiltered to avoid a silent blackout; "
+            f"'{processing_level}' filtering was NOT applied."
+        )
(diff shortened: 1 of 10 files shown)


gadievron:

PR #77, opened 2026-05-28
fix(parsers/c): exclude on path components in extract_all, not a substring of the abspath

gadievron:
c/function_extractor.py extract_all skipped files via `any(excl in str(file_path) for excl in
['.git','build','test','node_modules'])` -- an unanchored substring test against the absolute path. So
files whose path merely contained a token were wrongly skipped ('src/latest/main.c' and 'contest/sol.c'
contain 'test'), and an ancestor of repo_path containing a token poisoned the whole scan (a checkout
under '/home/tester/' excluded every file). Match on path COMPONENTS relative to repo_path instead,
using c's own token set.

Scope: c's member of the cross-parser extract_all substring family. The python/php/ruby extract_all
siblings carry DIFFERENT token sets (vendor*/tmp*/venv*) and are not widened here.

Tests: tests/test_c_extract_all_path_components.py (spies process_file: files with a token in a path
*segment* are processed; real test/ and .git dirs stay excluded). RED 1 failed (all excluded via
ancestor-poison) -> GREEN 1 passed; full suite 177 passed / 63 skipped. The test loads the C
function_extractor under a unique module name via importlib so it does not pollute
sys.modules['function_extractor'] for the sibling python parser tests.

### DIFF
--- a/libs/openant-core/parsers/c/function_extractor.py
+++ b/libs/openant-core/parsers/c/function_extractor.py
@@ -530,8 +530,11 @@ class FunctionExtractor:
             all_extensions = C_EXTENSIONS | CPP_EXTENSIONS
             for ext in all_extensions:
                 for file_path in self.repo_path.rglob(f'*{ext}'):
-                    path_str = str(file_path)
-                    if any(excl in path_str for excl in ['.git', 'build', 'test', 'node_modules']):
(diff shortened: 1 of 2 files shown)


gadievron:

PR #78, opened 2026-05-28
fix(parsers/go): correct call-graph receiver classification, package resolution, dispatch, os filtering

gadievron:
Four coupled defects in parsers/go/go_parser/callgraph.go, the Go call-graph resolver:

1. Receiver classification ignored the import table. isLikelyPackage classified a selector receiver `x`
   in `x.Sel()` as a package using a name-shape heuristic (lowercase first letter, len<=10) with no
   import context, so every short lowercase local (db, tx, ctx, w, r) was treated as a package and its
   method calls were misrouted to package resolution. Now a receiver is a package iff its name is one of
   the file's import aliases; the heuristic function is removed.

2. Package-call resolution matched the alias, not the package. resolvePackageCall resolved the import
   path (pkgPath := imports[pkgAlias]) but then matched candidates with strings.Contains(funcID,
   pkgAlias) -- the user-chosen alias. So `import u "example.com/user"` failed to resolve (the alias `u`
   is absent from the funcID), and any funcID merely containing the alias as a substring matched,
   emitting edges to unrelated packages. Now it matches the package directory (the last component of the
   resolved import path) against the directory the candidate's file lives in.

3. Dispatch misrouted a plain function's simple call. resolveCalls took the self/method branch when
   `call.Receiver == callerInfo.ClassName`, which is ""=="" for a plain function (ClassName=="") making
   a simple call (Receiver==""), so `greet()` from `main()` went to resolveMethodCall(name, "", file),
   found nothing, and the edge was lost. The branch is now guarded on callerInfo.ClassName != "", so
   non-method callers fall through to resolveSimpleCall.

4. The whole os package was skipped as a builtin. builtins included "os": true, so the call-extraction
   filter dropped os.StartProcess and other OS-level sinks before they could be seen. "os" is no longer
   blanket-skipped. (Such calls resolve to no in-repo target -- surfacing external sinks as graph nodes
   is a separate feature -- but they are no longer filtered out of the extracted call list.)

Scope: the Go AST import-resolution logic is unique to go_parser; the python/php/ruby/zig call-graph
builders are separate implementations and are not affected. extractor.go/types.go are unchanged.

Tests: parsers/go/go_parser/callgraph_test.go (new; the package had no tests). Four tests exercise the
stable extractCalls / resolvePackageCall / resolveCalls boundaries: local var not classified as a
package, package resolution by directory not alias, plain-func simple call not misrouted to method
resolution, and os sink not filtered as a builtin. RED 4 failed (pre-fix) -> GREEN 4 passed; go vet
clean; go test ./... ok.

### DIFF
--- a/libs/openant-core/parsers/go/go_parser/callgraph.go
+++ b/libs/openant-core/parsers/go/go_parser/callgraph.go
@@ -47,7 +47,8 @@ func NewCallGraphBuilder(repoPath string) *CallGraphBuilder {
 		"sort":    true,
 		"math":    true,
 		"io":      true,
-		"os":      true,
+		// "os" is intentionally NOT skipped: it carries security-relevant sinks (os.StartProcess,
+		// etc.) that downstream analysis must be able to see; blanket-skipping dropped them.
 		"path":    true,
 		"regexp":  true,
 		"json":    true,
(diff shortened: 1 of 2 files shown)


gadievron:

PR #79, opened 2026-05-28
fix(parsers): anchor is_test_file detection to path components + basename (zig/python/php/ruby)

gadievron:
Each parser's RepositoryScanner test-file detection used an unanchored
substring scan (`pattern in path_lower` / `pattern in part`) over its own
test-token set, so ordinary sources whose name merely contained a token
(latest_release, Contest.php, fastest.zig, contest/) were misclassified as
tests and silently pruned from scan_results['files'] when skip_tests=True.
The downstream extractor only consumes that file list, so functions in those
files vanished from the parsed dataset (silent recall loss).

Fix per parser, each keeping its OWN native token set (not copied across
parsers): match directory tokens against whole path components and filename
tokens against the basename via startswith/endswith.

- python  -- dirs {test, tests}; basename prefix test_; suffix _test.py; exact conftest.py
- php (PHPUnit) -- dirs {test, tests, spec}; prefix test_/phpunit; suffix _test.php; PSR
  PascalCase *Test.php matched on the original-case basename so FooTest.php is a test
  while Contest.php is not
- ruby (Minitest/RSpec) -- dirs {test, tests, spec}; prefix test_; suffix _test.rb/_spec.rb
- zig -- dirs {test, tests, spec, specs} exact; basename stem prefix test_/spec_; suffix
  _test.zig/_spec.zig

The zig scanner's `skip_tests` constructor default is also aligned to False (it
was the lone parser defaulting True, making the data loss active-by-default for
programmatic callers; the CLI already overrides via --skip-tests).

The C scanner's test detection and the JS scanner are out of scope here (the JS
scanner's isTestFile is already anchored). C function_extractor extract_all
exclusion is a different mechanism and is not touched.

Per-parser regression tests (importlib unique-name load, since repository_scanner.py
recurs across parser dirs) pin both directions: previously-misclassified non-test
names now return False, and genuine test files still return True. RED pre-fix:
python 3 / php 2 / ruby 2 / zig 6 failed. GREEN: 31 passed. Full suite 207 passed,
63 skipped, 0 failed. ruff clean.

### DIFF
--- a/libs/openant-core/parsers/php/repository_scanner.py
+++ b/libs/openant-core/parsers/php/repository_scanner.py
@@ -82,7 +82,17 @@ class RepositoryScanner:
 
         # Skip test files by default (can be overridden)
         self.skip_tests = options.get('skip_tests', False)
-        self.test_patterns = {'test_', '_test.php', 'Test.php', 'test/', 'tests/', 'spec/', 'phpunit'}
+        # Native PHP test conventions (PHPUnit). Directory names match whole
+        # path segments; filename rules are anchored to the basename so that
+        # ordinary sources like ``Contest.php``/``latest_helper.php`` are NOT
+        # misclassified as tests (an unanchored substring scan would).
+        self.test_dir_names = {'test', 'tests', 'spec'}
+        # Case-insensitive basename rules.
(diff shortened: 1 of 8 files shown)


gadievron:

PR #80, opened 2026-05-28
fix(cli): add --limit to the enhance command across all three layers

gadievron:
The `enhance` command lacked a `--limit` (max-units) capability while its siblings `analyze` and `scan`
have it, so a user could not cost-limit or test-run enhancement on N units -- enhance always processed the
full dataset. The gap spanned all three layers (a Go-only patch would be inert: the Python CLI rejected
--limit and the core could not consume it):

- Go (apps/openant-cli/cmd/enhance.go): add a --limit flag (default 0 = no limit) and forward it to the
  Python args when > 0, mirroring analyze.go.
- Python CLI (libs/openant-core/openant/cli.py): add --limit to the enhance subparser and pass
  limit=args.limit into enhance_dataset, mirroring analyze.
- Core (libs/openant-core/core/enhancer.py): add a `limit` parameter to enhance_dataset and slice
  dataset["units"] to units[:limit] before enhancement, mirroring analyzer.py (limit + units[:limit]).
  Slicing dataset["units"] (not only the local list) is what makes the limit reach both the agentic and
  single-shot enhancement paths.

Scope: analyze.go / analyze_p / analyzer.py already have --limit (the parity baseline, unchanged here);
verify/report/etc. do not enhance units, so they are out of scope.

Tests:
- libs/openant-core/tests/test_enhance_limit.py: with the LLM stubbed, enhance_dataset(limit=2) on a
  5-unit dataset makes the enhancer RECEIVE exactly 2 units (asserting on what the enhancer receives guards
  the load-bearing dataset["units"] slice -- verified by a mutant: dropping that line makes the enhancer
  see all 5 and the test fail); and `enhance ... --limit 3` driven through cli.main() forwards limit=3 to
  enhance_dataset (covers the enhance_p parser + cmd_enhance glue).
- apps/openant-cli/cmd/enhance_limit_test.go: enhanceCmd exposes a --limit flag (default 0).
RED (pre-fix, via stash): Python "unrecognized arguments: --limit" + Go "missing --limit flag" -> GREEN.
Full py suite 178 passed / 63 skipped; go test ./cmd/ ok; ruff + gofmt + go vet clean.

### DIFF
--- a/apps/openant-cli/cmd/enhance.go
+++ b/apps/openant-cli/cmd/enhance.go
@@ -32,6 +32,7 @@ var (
 	enhanceCheckpoint     string
 	enhanceWorkers        int
 	enhanceBackoff        int
+	enhanceLimit          int
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


gadievron:

PR #81, opened 2026-05-28
fix(analyze): add Go --exploitable-all parity; priority-sort --limit truncation

gadievron:
Two confirmed bugs on the analyze path; both fixes are additive and keep
existing callers' behavior unchanged when the new code path is not taken.

Go --exploitable-all missing: the Python backend (openant/cli.py analyze_p)
defines both --exploitable-all and --exploitable-only in a mutually-exclusive
group (-> exploitable_filter 'all'|'strict'|None), but the Go CLI
defined/forwarded only --exploitable-only, so
`openant analyze --exploitable-all` failed with `unknown flag`. The no-flag
DEFAULT is most inclusive (analyzer runs the filter under
`if exploitable_filter:`), so the omitted flag is a narrowing cost filter with
no recall loss. FIX: add --exploitable-all (BoolVar + forward) mirroring the
Python help; mark the two exploitable flags mutually exclusive. Extract the
inline argv build into a pure buildAnalyzePyArgs helper (mirrors
buildParsePyArgs) so the flag-forwarding contract is unit-testable without
spawning Python.

--limit drops high-value units: run_analysis truncated with a raw head-slice
`units = units[:limit]` over parser-sorted alphabetical-by-path units (Doc/
before Lib/), dropping high-value code with no relevance weighting. FIX:
extract _apply_limit(units, limit) that priority-sorts by enhancement
security_classification (exploitable > vulnerable_internal > other) before the
head-slice; stable within a tier, no-op when limit is unset, reads
classification mode-agnostically (agent_context or llm_context). The bias is
bounded to --limit runs; full runs pass no limit. The sibling head-slice at
experiment.py:526 is separate and not touched here.

Tests: cmd/analyze_flags_test.go (flag defined + forwarded via
buildAnalyzePyArgs) + tests/test_analyzer_limit_priority.py (6 tests). go test
./cmd/ ok, gofmt + vet clean; python regression suite passing; ruff clean.

### DIFF
--- a/apps/openant-cli/cmd/analyze.go
+++ b/apps/openant-cli/cmd/analyze.go
@@ -30,6 +30,7 @@ var (
 	analyzeAppContext     string
 	analyzeRepoPath       string
 	analyzeExploitOnly    bool
+	analyzeExploitAll     bool
 	analyzeLimit          int
 	analyzeModel          string
 	analyzeWorkers        int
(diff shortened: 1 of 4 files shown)


gadievron:

PR #82, opened 2026-05-28
fix(cli/config): enforce 0600 on the config file even when it pre-exists (CWE-732)

gadievron:
config.Save() wrote the config (which may hold an API key) with
os.WriteFile(path, data, 0600). os.WriteFile only applies the mode when it
CREATES the file; if config.json already existed with looser permissions
(e.g. 0644), the secret stayed world/group-readable after Save(). Enforce 0600
explicitly via os.Chmod after the write (CWE-732, insecure permissions for a
secret-bearing file).

Tests: internal/config/config_perms_test.go (a pre-existing 0644 secret config
becomes 0600 after Save). RED 1 failed (0644) -> GREEN; go test
./internal/config/ ok; gofmt + go vet clean.

## Coordination
Touches a file also modified by in-flight PR #69 (region-disjoint; textual merge only).

### DIFF
--- a/apps/openant-cli/internal/config/config.go
+++ b/apps/openant-cli/internal/config/config.go
@@ -107,6 +107,12 @@ func Save(cfg *Config) error {
 		return fmt.Errorf("failed to write config: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


gadievron:

PR #83, opened 2026-05-28
fix(export_csv): neutralize CSV/formula injection in exported cells (CWE-1236)

gadievron:
export_csv() wrote the verbatim scanned source (unit_code) and LLM-derived text
columns into the CSV with no neutralization. A scanned snippet beginning with =,
+, -, @ (or a leading tab/CR) is treated as a formula by Excel / Google Sheets
when the analyst opens the export, enabling formula execution / data
exfiltration on the analyst's machine (OWASP CSV Injection / CWE-1236).

Add a _csv_safe() helper that prefixes any cell whose first character is a
formula trigger with a single quote, applied to every cell before writing.

The payload originates in the scanned (untrusted) source and only fires in a
downstream spreadsheet app -- it is not reachable from OpenAnt's own callers --
so this is output hardening / defense-in-depth, not an exploitable-now
vulnerability. It is a real producer-side output-sanitization gap (CWE-1236),
distinct from the "OpenAnt processing hostile input" class.

Tests: tests/test_export_csv_formula_injection.py (2 cases: _csv_safe
neutralizes =/+/-/@/tab/CR and leaves safe cells unchanged; export_csv
neutralizes a formula unit_code end-to-end). RED 2 failed -> GREEN 2 passed
(py3.11).

### DIFF
--- a/libs/openant-core/export_csv.py
+++ b/libs/openant-core/export_csv.py
@@ -31,6 +31,20 @@ import os
 import sys
 from utilities.file_io import read_json
 
+# Characters that make a spreadsheet treat a cell as a formula on open (CSV / formula
+# injection, CWE-1236 / OWASP). Cells written from scanned source or LLM text must be
+# neutralized so opening the export in Excel / Google Sheets can't execute a payload.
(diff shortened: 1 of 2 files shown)


gadievron:

PR #86, opened 2026-05-28
fix(ruby-extractor): repair 7 Ruby FunctionExtractor extraction/classification bugs

gadievron:
The Ruby tree-walk dispatched on only method/singleton_method/class/module nodes
and tracked no method visibility, so several real Ruby constructs were silently
dropped or mislabeled, degrading the function inventory and the call graph /
entry-point analysis built on it.

Fixes:
- Nested (`module Outer; module Inner`) and compact (`module Outer::Inner`) module
  names were flattened/dropped — now read via the `name` field (handles constant +
  scope_resolution) and concatenated as `Outer::Inner` instead of replaced, so
  qualified calls resolve.
- `define_method(:sym)` do..end / { } class-body metaprogramming now emits method
  units (parses as a `call` node).
- `alias_method :a, :b` (symbol + string forms) now extracted.
- `alias new old` keyword form (`alias` node) now extracted.
- Method visibility (private/protected/public) now threaded through the traversal
  and stored on each unit; the arg-form `private :sym` is consumed without leaking
  into the bare-marker toggle.
- Non-public controller methods (params helpers, before_action targets) classified
  before the controller branch, so they are no longer over-flagged as route_handler
  / entry points; public actions unchanged.
- Top-level Sinatra route DSL (`get '/path' do..end`) now emits route_handler units.

Method-body calls are never traversed (no false positives); non-matching
top-level/class-body calls fall through to a normal child descent.

Scope / compatibility: adds a `visibility` field to emitted units (additive).
`unit_type` for controller non-public methods narrows from `route_handler` to
`private_method`/`protected_method` (correctness fix; removes the entry-point
over-claim). `_classify_function` and `_process_method_node` gained an optional
`visibility` parameter (default `public`); no public API signature break. The
arg-form `private :sym` targeted privatization is out of scope and left as a
documented no-op toggle.

Tests: tests/parsers/ruby/test_function_extractor.py. The test imports the Ruby
extractor package-qualified (from parsers.ruby.function_extractor import
FunctionExtractor) so it does not shadow the python/php modules of the same basename
in the shared pytest session. 9 failed pre-fix -> 9 passed after the fix (8 bug
tests + 1 arg-form visibility-leak regression guard). Full suite: 185 passed, 63
skipped, 0 failed. ruff clean.

### DIFF
--- a/libs/openant-core/parsers/ruby/function_extractor.py
+++ b/libs/openant-core/parsers/ruby/function_extractor.py
@@ -127,8 +127,15 @@ class FunctionExtractor:
 
     def _classify_function(self, func_name: str, class_name: Optional[str],
                            module_name: Optional[str], is_singleton: bool,
-                           file_path: str) -> str:
-        """Classify a function by its type/purpose."""
+                           file_path: str, visibility: str = 'public') -> str:
+        """Classify a function by its type/purpose.
+
+        ``visibility`` ('public' | 'private' | 'protected') is the declared
+        Ruby method visibility threaded from the class body. Only PUBLIC
+        controller methods are real route handlers; private/protected methods
+        in a controller are filter callbacks / params helpers and must not be
+        seeded as entry points.
+        """
         path_lower = file_path.lower()
 
         if func_name == 'initialize':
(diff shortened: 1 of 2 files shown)


gadievron:

PR #87, opened 2026-05-28
fix(parsers/zig): align function_extractor with the real tree-sitter-zig grammar + lazy import

gadievron:
Pinned against the actually installed grammar (tree-sitter-zig 1.1.2) by
AST-probing it directly; several AST node-type names had drifted. Three defects
in libs/openant-core/parsers/zig/function_extractor.py:

1. Struct extraction was dead and fields could be emitted as functions. Struct
   handling gated on node types the 1.1.2 grammar never emits (VarDecl /
   container_decl / ContainerDecl); the grammar emits variable_declaration and
   struct_declaration, so total_classes was always 0 and the literal
   container_field->method path was unreachable. Corrected the node-type names;
   methods now flow through the recursive walk which only treats
   function_declaration as a unit, so container_field (struct FIELDS) are never
   emitted. The dedicated _extract_struct_methods helper (home of the original
   container_field bug) is removed as superseded.

2. A struct method was emitted twice (qualified Type.method by the eager scan,
   then bare by an unconditional recursion that never threaded current_struct).
   Now thread the struct name into the recursion so a method produces the same
   func_id and the dict de-duplicates instead of producing a bare twin. Also fix
   a name-capture bug in _extract_function (it grabbed the last identifier,
   recording `fn init(...) Point {` under the return type `Point`); it now
   captures the first identifier and stops. Regression test pins that the de-dup
   does not drop nested struct / method / @import.

3. tree_sitter_zig was imported at module top level and called at class-body
   time, raising ImportError in any clean install lacking the package, which was
   declared in neither pyproject.toml nor requirements.txt. Load the grammar
   lazily on first FunctionExtractor construction with a clear error message, and
   declare tree-sitter-zig>=1.1.2 in pyproject.toml + requirements.txt.
   (call_graph_builder.py carries the same top-level import but is out of this
   change's scope; the dependency declarations cover it too.)

New tests/parsers/zig/test_zig_extractor_followup.py (loaded via importlib unique
name because function_extractor.py recurs across parser packages). RED at base
6 failed / 1 passed; GREEN 7 passed. Full suite 183 passed, 63 skipped, 0 failed.

### DIFF
--- a/libs/openant-core/parsers/zig/function_extractor.py
+++ b/libs/openant-core/parsers/zig/function_extractor.py
@@ -10,19 +10,37 @@ from typing import Dict, Any, Optional, List
 
 from utilities.file_io import write_json
 
-import tree_sitter_zig as ts_zig
 from tree_sitter import Language, Parser, Node
+
+
+def _load_zig_language() -> Language:
+    """Load the Zig tree-sitter grammar lazily.
+
+    The grammar package (``tree-sitter-zig``) is an optional runtime
+    dependency. Importing it at module top level made the entire Zig parser
+    unimportable in any environment where the package is absent (e.g. a clean
+    install that does not need Zig support). Resolving it here, on first use,
+    lets the module import unconditionally and surfaces a clear, actionable
+    error only when the Zig parser is actually exercised.
+    """
+    try:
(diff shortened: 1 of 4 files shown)


gadievron:

PR #89, opened 2026-05-28
fix(parsers/go): give CodeQL DB-create the same timeout budget as analyze

gadievron:
go/test_pipeline.py ran `codeql database create` with timeout=600s and `codeql database analyze` with
timeout=1800s. DB creation compiles the source -- the slower stage for compiled languages -- so the slower
stage had the smaller budget and times out first. On a create timeout run_codeql_analysis records the codeql
stage success=False and returns; run_full_pipeline prints "continuing with reachable units only" and proceeds,
and apply_codeql_filter (test_pipeline.py:1009, success-gated) is skipped -- so the written dataset is
reachable-only with CodeQL findings dropped. The run does report success=False / exit 1 (all_success ANDs the
failed stage, :1040-1045, :1199), so an exit-code CI gate still catches it; the harm is the silently-degraded
artifact for any consumer that reads the results rather than the exit code.

Extract both stage timeouts to named module constants and set the create budget equal to analyze (1800s), so
the create stage is no longer shortchanged.

Scope: the c/php/js/ruby test_pipeline.py orchestrators carry the same inverted timeouts and are separate
units -- not widened here.

Tests: tests/test_go_test_pipeline_codeql_timeout.py -- two checks that read the source rather than importing
the module (which does module-level sys.path manipulation and shares its name with sibling parser
test_pipeline.py): the create timeout >= analyze timeout, and that both call sites pass the named constants
(not inline literals, which could silently drift the budget back). RED -> GREEN; full suite 178 passed,
63 skipped, 0 failed.

### DIFF
--- a/libs/openant-core/parsers/go/test_pipeline.py
+++ b/libs/openant-core/parsers/go/test_pipeline.py
@@ -93,6 +93,13 @@ class ProcessingLevel(Enum):
     EXPLOITABLE = "exploitable"    # Level 4: Filter to reachable + CodeQL-flagged + exploitable
 
 
+# CodeQL stage subprocess timeouts (seconds). DB `create` compiles the source -- the slower stage for
+# compiled languages -- so it must not get a smaller budget than `analyze`: a too-short create times out,
+# the stage degrades to reachable-only, and CodeQL security findings are silently lost.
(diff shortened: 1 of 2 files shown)


gadievron:

PR #90, opened 2026-05-28
fix(parsers/python): segment-match path exclusion/classification + resolve relative-import anchors

gadievron:
Three independent defects in parsers/python/function_extractor.py:

1. extract_all(): the no-args scan excluded files with
   `any(excl in str(file_path) for excl in [...])` -- an unanchored substring test on the full path, so
   a file whose path merely contains a token ('myvenv/keep.py' contains 'venv') was silently dropped,
   and an ancestor directory containing a token could exclude the whole scan. Now matches whole path
   SEGMENTS: `{tokens} & set(file_path.relative_to(repo_path).parts)`. Python's own token set
   (__pycache__/.git/venv/.venv/node_modules) is preserved.

2. classify_function(): classification used `'<token>' in path_lower` substring tests, so
   'interviews/api.py' was classified 'view_function'. 'view_function' is in
   entry_point_detector.ENTRY_POINT_TYPES (:26-32), so that misclassification became a false entry-point
   seed that cascades into false reachability (consumed at entry_point_detector.py:177). The 'views'
   token now matches a whole path segment via a new _path_has_segment helper. The 'middleware' token is
   given the same segment fix because it shares the substring defect, but note 'middleware' (the python
   label) is NOT in ENTRY_POINT_TYPES -- so that half is classification accuracy, not a reachability
   change. The 'test' classifier is left as a substring on purpose (test-file conventions use 'tests/'
   and 'test_*'/'*_test' forms a segment match would miss; 'test' is not an entry-point type, so it
   seeds no false reachability).

3. extract_imports(): the ast.ImportFrom branch read node.module but never node.level, so relative
   imports lost their package anchor ('from . import X' stored bare 'X'; 'from ..pkg import Y' stored
   anchor-less 'pkg.Y'). call_graph_builder._resolve_import then rebuilt a wrong/no file path and the
   edges were dropped (verified: pre-fix the candidate resolves to None, post-fix it resolves to the real
   pkg/sub/helpers.py). Now reconstructs the absolute anchor from the importing file's package location
   (level=1 -> own package, level=2 -> parent, ...); over-deep levels degrade to no leading dot. Absolute
   imports (level=0) are unchanged.

Scope: the php/ruby function_extractor.py extract_all + classify siblings carry related defects and are
not widened here.

Tests: tests/test_python_function_extractor.py -- loads the module under a unique importlib name (the
bare 'function_extractor' name is shared by five other parsers, so a plain import would pollute
sys.modules for the rest of the suite). Three checks: segment-vs-substring exclusion, entry-point
classification by segment, and relative-import anchor reconstruction. RED 3 failed (pre-fix) -> GREEN 3
passed; full suite 179 passed / 63 skipped.

### DIFF
--- a/libs/openant-core/parsers/python/function_extractor.py
+++ b/libs/openant-core/parsers/python/function_extractor.py
@@ -194,6 +194,19 @@ class FunctionExtractor:
         """Extract docstring from a function or class."""
         return ast.get_docstring(node)
 
+    def _path_has_segment(self, file_path: str, token: str) -> bool:
+        """True if `token` equals a whole path segment (a directory name or the filename stem),
+        case-insensitively -- used instead of a bare ``token in path`` substring test so that, for
+        example, 'views' classifies ``app/views.py`` and ``app/views/x.py`` but NOT ``interviews/a.py``
+        or ``app/previews/b.py``."""
+        p = Path(file_path)
+        try:
+            p = p.relative_to(self.repo_path)
+        except ValueError:
+            pass
+        segments = {s.lower() for s in p.with_suffix('').parts}
+        return token.lower() in segments
+
     def classify_function(self, func_name: str, decorators: List[str],
                           class_name: Optional[str], file_path: str) -> str:
         """Classify a function by its type/purpose."""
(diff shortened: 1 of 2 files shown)


gadievron:

PR #91, opened 2026-05-28
fix(parsers/c): codeql build-mode-none, optional-stage success, exclude orchestrators from pytest

gadievron:
Three defects surfaced via parsers/c/test_pipeline.py (the C pipeline orchestrator):

1. CodeQL build mode: `codeql database create` for cpp omitted --build-mode=none, so CodeQL defaulted
   to autobuild and silently degraded (dropping findings) on the no-build / extracted-source repos this
   pipeline runs on. Pass --build-mode=none.

2. Optional-stage success: overall success ANDed over ALL recorded stages, but the optional stages
   (CodeQL analysis/filter, reachability filter, context enhancer, exploitable filter) write
   success=False when they fail or are skipped -- so an optional-stage failure forced a spurious
   pipeline failure (exit 1). Introduce OPTIONAL_STAGES + a _compute_success() that requires only the
   non-optional stages. (Cross-parser family: the same all_success conjunction exists in the
   go/php/ruby/javascript pipelines -- separate units, not widened here.)

3. pytest collection collision: the six parsers/<lang>/test_pipeline.py files are CLI pipeline
   ORCHESTRATORS, not pytest tests, but share a basename, so `pytest parsers/` fails with
   import-file-mismatch. __init__.py does not fix it (their bare local imports -- `from
   repository_scanner import ...` -- make them un-importable as package modules; adding __init__.py
   merely moves the collision). Add a root conftest.py that excludes them from collection
   (collect_ignore_glob), which is correct since they are not tests. The canonical suite is unaffected
   (pytest.ini already scopes testpaths = tests).

Scope: c-specific for #1 (go's codeql create may need its own --build-mode -- separate unit) and #2 (the
all_success family is fixed per-parser); #3 is repo-wide (the conftest excludes all six orchestrators).
function_extractor / other parsers unchanged.

Tests: tests/test_c_pipeline.py -- (1) source-read that the C codeql create cmd passes --build-mode=none;
(2) behavioral _compute_success() ignores optional-stage failures but fails on a required-stage failure
(loaded via a CONTAINED importlib import that pops the polluting sibling-parser modules in a finally,
since c/test_pipeline.py does bare local imports); (3) the root conftest excludes the orchestrators.
RED 3 failed (pre-fix) -> GREEN; ruff clean; full suite 179 passed, 63 skipped, 0 failed; `pytest parsers/
--co` now collects cleanly (was import-file-mismatch).

### DIFF
--- a/libs/openant-core/conftest.py
+++ b/libs/openant-core/conftest.py
@@ -1,0 +1,4 @@
+# The parsers/<lang>/test_pipeline.py files are CLI pipeline orchestrators (run as scripts), not
+# pytest tests. Their shared basename + script-only bare imports make pytest mis-collection fail
+# with import-file-mismatch under `pytest parsers/`. Exclude them from collection.
+collect_ignore_glob = ["parsers/*/test_pipeline.py"]
--- a/libs/openant-core/parsers/c/test_pipeline.py
+++ b/libs/openant-core/parsers/c/test_pipeline.py
@@ -68,6 +68,15 @@ class ProcessingLevel(Enum):
     EXPLOITABLE = "exploitable"
 
 
+# Pipeline stages that are OPTIONAL: they may fail or be skipped (e.g. CodeQL not installed, no entry
+# points) without the run being a failure. They record success=False on failure, so they must be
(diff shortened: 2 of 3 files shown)


gadievron:

PR #92, opened 2026-05-28
fix(parsers/js): discover and resolve .mjs/.cjs modules in context_assembler

gadievron:
context_assembler.js omitted .mjs (ESM) and .cjs (CommonJS) from both its source-discovery glob
(findSourceFiles) and its import-resolution extension lists. So .mjs/.cjs files were never loaded into
the TypeScript program (getSourceFile -> undefined) and imports targeting them were silently unresolved --
dropped call-graph edges. The sibling repository_scanner.js already includes .mjs/.cjs in its
sourceExtensions; this brings context_assembler to the same canonical set (.js/.ts/.jsx/.tsx/.mjs/.cjs).

Three sites updated:
- findSourceFiles discovery glob: add mjs|cjs.
- import-resolution extensions: add .mjs/.cjs + /index.mjs / /index.cjs.
- resolveModuleFunctionDefinition extensions (the sibling site -- it omitted even .jsx/.tsx): bring to
  the canonical set.

Tests: tests/test_context_assembler_mjs.py (ContextAssembler.findSourceFiles discovers .mjs/.cjs; skips
portably where the JS parser's node deps aren't installed). RED 1 failed (['main.js']) -> GREEN 1 passed;
JS parser suite 10 passed; full venv suite 218 passed, 22 skipped, 0 failed (with node deps installed,
which unlocks the otherwise-skipped JS tests). node --check clean.

### DIFF
--- a/libs/openant-core/parsers/javascript/context_assembler.js
+++ b/libs/openant-core/parsers/javascript/context_assembler.js
@@ -84,7 +84,7 @@ class ContextAssembler {
 
             if (entry.isDirectory()) {
                 this.findSourceFiles(fullPath, files);
-            } else if (/\.(js|ts|jsx|tsx)$/.test(entry.name) && !entry.name.endsWith('.d.ts')) {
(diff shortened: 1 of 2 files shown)


gadievron:

PR #93, opened 2026-05-28
fix(c-parser): anchor is_test_file patterns to path segments / filename boundaries

gadievron:
RepositoryScanner.is_test_file matched test_patterns as bare substrings
(`pattern in path_lower`), so ordinary source files whose name/path merely
CONTAINED a pattern were wrongly skipped as tests when skip_tests=True:
  - latest_value.c  (contains 'test_')  -> skipped
  - contest/foo.c   (contains 'test/')  -> skipped

Match directory patterns (test/, tests/, fuzz/) against whole path *segments*,
and filename patterns against the basename: '_test' as a stem suffix and
'test_' as a prefix. The _test suffix is matched on the stem (before the
extension) so _test.cc / _test.cxx — which the old substring check also caught
— stay detected, bounded by is_source_file.

Adds tests/parsers/c/test_repository_scanner_is_test_file.py (RED->GREEN; both
true-positive and false-positive directions, incl. the _test.cc regression case).
GREEN: 2 passed (venv py3.11); ruff clean.

The same substring bug exists in the python/php/ruby repository_scanner.py —
handled in their own PRs (systemic family).

### DIFF
--- a/libs/openant-core/parsers/c/repository_scanner.py
+++ b/libs/openant-core/parsers/c/repository_scanner.py
@@ -102,11 +102,33 @@ class RepositoryScanner:
         return ext in self.source_extensions
 
     def is_test_file(self, relative_path: str) -> bool:
-        """Check if a file is a test file."""
+        """Check if a file is a test file.
+
+        Patterns are matched against path *segments* and filename boundaries,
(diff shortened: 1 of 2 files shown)


gadievron:

PR #95, opened 2026-05-28
fix(enhance/analyzer): retry 529-overloaded, single-shot checkpoint, exploitable filter + bounded agentic retry

gadievron:
Four confirmed bugs across the enhance and analyzer paths; all fixes keep
existing callers' behavior (new params optional, default to prior behavior).

529-overloaded never retried: is_retryable_error()'s string branch omitted
"529"/"overloaded", and the analyzer detection path feeds str(e) (not a dict),
so Anthropic-overloaded errors were silently never auto-retried. Add
"529"/"overloaded" to the string-branch tuple, mirroring the dict branch's
status_code >= 500.

exploitable filter dropped single-shot units: the analyzer exploitable filter
read only agent_context.security_classification, silently dropping every
single-shot unit (which writes llm_context). Add
_unit_security_classification() consumer fallback (agent_context ->
llm_context) plus a loud WARNING guard when a filter is set but no unit carries
any classification.

single-shot had no checkpoint/resume: single-shot enhance_dataset had no
checkpoint/resume. Add an optional checkpoint_path (per-unit save + resume +
KeyboardInterrupt handling); generalize the shared checkpoint helpers with an
optional context_key (default agent_context, so agentic is unchanged);
auto-derive + thread the checkpoint path for both modes in core/enhancer.py.

agentic post-loop retry was single-pass: the agentic post-loop transient retry
was a single pass. Wrap it in a bounded multi-round loop (MAX_RETRY_ROUNDS=3)
that stops early when no retryable units remain or a round recovers nothing.

Tests: tests/test_enhance_resilience.py (9 tests). ruff + py_compile clean.

## Coordination
Touches a file also modified by in-flight PR #51 (region-disjoint; textual merge only).

### DIFF
--- a/libs/openant-core/core/analyzer.py
+++ b/libs/openant-core/core/analyzer.py
@@ -47,6 +47,24 @@ except ImportError:
     load_context = None
 
 
+def _unit_security_classification(unit):
+    """Return a unit's security_classification, regardless of enhance mode.
+
+    Agentic enhance writes ``unit['agent_context']['security_classification']``;
+    single-shot enhance writes ``unit['llm_context']`` and historically had no
+    classification at all. Reading only ``agent_context`` silently dropped every
+    single-shot unit from ``--exploitable-only/all``. Prefer
+    agent_context, fall back to llm_context, else None.
+    """
+    agent_ctx = unit.get("agent_context")
+    if isinstance(agent_ctx, dict) and agent_ctx.get("security_classification") is not None:
+        return agent_ctx.get("security_classification")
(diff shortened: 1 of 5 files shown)


gadievron:

`core/analyzer.py` here overlaps **#81**. The two helpers feed different call sites, so keep **both**:
- `_apply_limit`/`_limit_classification` (#81) — priority-sort for `--limit`;
- `_unit_security_classification` (this PR) — the `--exploitable` filter.

A plain accept-ours/theirs silently drops one feature.

gadievron:

PR #96, opened 2026-05-28
fix(file_io): write_json atomically (temp + os.replace) to survive interrupted writes

gadievron:
write_json opened the target in "w" (truncating it to 0 bytes) and then streamed json.dump.
A crash mid-serialization (SIGKILL / OOM / power loss) left the target partial or empty and
destroyed the prior good copy, so the next read_json raised JSONDecodeError -- e.g. a
corrupted checkpoint.json that breaks resume.

Serialize to a temp file in the same directory, flush + fsync, then os.replace onto the
target (atomic rename). An interrupted write leaves only the temp file (unlinked on a caught
error); the existing target is never truncated. All write_json call sites inherit the fix.
The temp uses a `.tmp` suffix (not `.json`) so a hard-crash leftover can't be miscounted by
directory scanners that do os.listdir + endswith(".json") (e.g. core/checkpoint.py).

Tests: tests/test_file_io_atomic_write.py (2 cases: interrupted write preserves the prior
file + no temp leak; normal round-trip incl. non-ASCII). RED 1 failed (JSONDecodeError) ->
GREEN 2 passed (venv py3.11). ruff clean.

### DIFF
--- a/libs/openant-core/tests/test_file_io_atomic_write.py
+++ b/libs/openant-core/tests/test_file_io_atomic_write.py
@@ -1,0 +1,42 @@
+"""Regression test — write_json is non-atomic (truncate-then-write).
+
+write_json opens the target in "w" (truncating it to 0 bytes) and then streams json.dump. If the process is
+killed mid-serialization (SIGKILL / OOM / power loss), the target is left partial/empty and the previous good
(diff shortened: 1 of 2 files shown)


gadievron:

PR #99, opened 2026-05-28
fix(parsers/python): surface top-level callGraph/reverseCallGraph in analyzer_output

gadievron:
generate_analyzer_output() emitted only {"functions": ...}, dropping the top-level callGraph /
reverseCallGraph that the analyzer_output schema carries. They were already computed in
call_graph_result (CallGraphBuilder.export() returns call_graph / reverse_call_graph) but were
never passed in, so the python parser's analyzer_output.json was schema-inconsistent with the
documented format.

The analyzer_output schema uses camelCase top-level keys: PARSER_UPGRADE_PLAN documents
"callGraph", and the JS pipeline emits/asserts camelCase callGraph / reverseCallGraph
(funcId -> [funcIds]). Pass call_graph_result into generate_analyzer_output (optional arg,
back-compatible) and emit those two keys from its snake_case source keys.

Scope: filePath is intentionally NOT a per-function field -- consumers (RepositoryIndex,
dependency_resolver) derive it from the func_id key via funcId.split(':')[0]. The original
filing's indirect_calls field is phantom (grep finds no such data produced anywhere) and is
excluded. The classes/imports top-level keys are a related but separately-scoped schema item.

Tests: tests/test_parse_repository_analyzer_schema.py (2 cases). RED 1 failed -> GREEN 2 passed
(venv py3.11). ruff clean.

### DIFF
--- a/libs/openant-core/parsers/python/parse_repository.py
+++ b/libs/openant-core/parsers/python/parse_repository.py
@@ -55,7 +55,7 @@ from unit_generator import UnitGenerator
 from utilities.file_io import read_json, write_json, open_utf8
 
 
-def generate_analyzer_output(extractor_result: dict) -> dict:
+def generate_analyzer_output(extractor_result: dict, call_graph_result: dict | None = None) -> dict:
     """
     Generate analyzer_output.json format for Stage 2 verification.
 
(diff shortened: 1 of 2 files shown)


gadievron:

PR #100, opened 2026-05-28
fix(reachability): make get_all_reachable() overwrite the cache so a stale bounded False is corrected

gadievron:
ReachabilityAnalyzer has two cache-population paths that disagree:
- is_reachable_from_entry_point() is a depth-BOUNDED reverse BFS; for a function farther than max_depth
  hops from an entry point it caches False -- a false negative;
- get_all_reachable() is an UNBOUNDED forward BFS (the complete reachable set).

get_all_reachable() updated the shared _reachability_cache only for keys NOT already present, so a stale
False seeded by an earlier bounded per-query call was frozen and never corrected -- the complete pass
could not fix it (order-dependent wrong result). Overwrite the cache unconditionally: get_all_reachable()
is authoritative. Because its forward graph is built from the same reverse edges the bounded BFS walks,
this can only flip a stale False->True, never introduce a false positive.

Tests: tests/test_reachability_cache.py -- a function reachable only beyond max_depth is cached False
by the reverse BFS, then get_all_reachable() corrects it to True (and a disconnected node stays
unreachable, a within-depth node stays reachable). RED 1 failed (stale False frozen) -> GREEN 2 passed
(venv py3.11); ruff clean.

### DIFF
--- a/libs/openant-core/tests/test_reachability_cache.py
+++ b/libs/openant-core/tests/test_reachability_cache.py
@@ -1,0 +1,36 @@
+"""Regression test -- ReachabilityAnalyzer stale-cache.
+
+is_reachable_from_entry_point() is a depth-BOUNDED reverse BFS: for a function farther than max_depth
+hops from an entry point it caches False (a false negative). get_all_reachable() is an UNBOUNDED forward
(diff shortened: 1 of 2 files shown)


gadievron:

PR #103, opened 2026-05-28
fix(js-parser): export FILE_BOUNDARY at module level for cross-parser parity

gadievron:
The JavaScript unit generator declared FILE_BOUNDARY as a function-local
`const` inside _assembleEnhancedCode and exported only `{ UnitGenerator }`, so
`require(...).FILE_BOUNDARY` was undefined. Every sibling parser (python:60,
php/c/ruby:35) declares this marker at module level and makes it importable;
the JS parser was the outlier, trapping the canonical boundary string in a
method body where no external consumer could reach it.

Move the declaration to module level next to the requires, reference it
unchanged inside _assembleEnhancedCode (resolves via lexical scope, output
byte-identical), and add it to module.exports. No runtime/behavior change.

Tests: new tests/parsers/javascript/test_unit_generator_file_boundary.py
(node-subprocess seam) asserts the export exists and equals the canonical
marker string. RED 2 failed pre-fix; GREEN 2 passed. Full suite 219 passed,
22 skipped, 0 failed; ruff + node --check clean.

### DIFF
--- a/libs/openant-core/parsers/javascript/unit_generator.js
+++ b/libs/openant-core/parsers/javascript/unit_generator.js
@@ -49,6 +49,11 @@ const fs = require('fs');
 const path = require('path');
 const { DependencyResolver } = require('./dependency_resolver');
 
+// File boundary marker for enhanced code (module-level for parity with the
+// python/php/c/ruby parsers, so external consumers can import the canonical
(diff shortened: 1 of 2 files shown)


gadievron:

PR #104, opened 2026-05-28
fix(cli/runtime): key dependency-staleness on corePath, not only the pyproject hash

gadievron:
The managed Python venv is a single global path (~/.openant/venv) shared by every worktree. The
staleness check (depsStalenessAt) keyed only on pyproject.toml content, so two worktrees with
identical pyproject.toml were considered "up to date" even though the venv's editable install
(`pip install -e <corePath>`) pointed at whichever worktree installed last -- a binary built in
worktree B would silently import worktree A's Python source.

Introduce depsHash(corePath) = sha256(corePath + "\0" + pyproject.toml contents) and use it for both
the install baseline stamp and the staleness comparison. A change of corePath (switching worktrees)
now reads as stale and triggers a reinstall that re-points the editable install at the active source.
(hashFile remains a tested file-hash utility, no longer on the staleness path.)

Tests: internal/python/runtime_corepath_test.go (two worktrees with byte-identical pyproject but
different corePath -> the second is detected as stale; same corePath -> not stale). RED 1 failed ->
GREEN; go test ./internal/python/ ok (existing staleness + hashFile tests unaffected); gofmt + vet clean.

### DIFF
--- a/apps/openant-cli/internal/python/runtime.go
+++ b/apps/openant-cli/internal/python/runtime.go
@@ -201,8 +201,7 @@ func CheckOpenantInstalled(pythonPath string) error {
 	}
 
 	// Save dependency hash so CheckDepsStale knows this is the baseline.
-	pyprojectPath := filepath.Join(corePath, "pyproject.toml")
-	if h, err := hashFile(pyprojectPath); err == nil {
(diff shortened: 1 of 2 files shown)


gadievron:

PR #105, opened 2026-05-28
fix(agentic-enhancer): cap conversation input to avoid context overflow

gadievron:
The agent built the LLM conversation with no input budget: get_user_prompt
inlined primary_code verbatim (prompts.py) and the loop appended raw,
untruncated tool results every iteration via json.dumps(result) (agent.py
append sites). On a large unit the constructed input grew without bound across
iterations until it overflowed the model context (400), losing the
enhancement. MAX_ITERATIONS=20 bounds depth but not per-item size.

Operative guard: an input budget at the consumption points.
- prompts.py: cap inlined primary_code (MAX_PRIMARY_CODE_CHARS).
- agent.py: add cap_tool_result_content() and use it at both tool-result
  append sites (MAX_TOOL_RESULT_CHARS).
- agent.py: graceful anthropic.BadRequestError (400) fallback returns a neutral
  incomplete result instead of crashing the enhancement.

A deferred sibling is out of scope here: tools.py _read_function returns the
whole function body uncapped; the consumption-point cap above contains it.

Tests: tests/test_agent.py (5 tests, RED before fix). All passing.

## Coordination
Touches a file also modified by in-flight PR #51/#60 (region-disjoint; textual merge only).

### DIFF
--- a/libs/openant-core/tests/test_agent.py
+++ b/libs/openant-core/tests/test_agent.py
@@ -1,0 +1,93 @@
+"""Regression tests for the agentic enhancer input budget.
+
+The agentic context enhancer built the LLM conversation with no input
+token/char budget: ``primary_code`` was inlined verbatim into the initial
+user prompt, and raw, untruncated tool results were appended to ``messages``
+every iteration. On a large unit the constructed input grew unbounded across
+iterations until it overflowed the model context (400 error), losing the
+enhancement.
+
+Operative guard: an INPUT BUDGET applied at the consumption points.
+  - ``get_user_prompt`` caps the inlined ``primary_code`` (prompts.py).
+  - ``cap_tool_result_content`` caps each serialized tool result before it is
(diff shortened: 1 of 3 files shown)


gadievron:
Superseded by #133 (2026-06-24), which already landed the input-cap fix on master: `MAX_PROMPT_CHARS`/`MAX_TOOL_RESULT_CHARS` + `cap_tool_result_content` in agent.py, `MAX_PRIMARY_CODE_CHARS`/`_cap_primary_code` in prompts.py, and tests/test_agent.py. This branch is anchored to the pre-#133 dict-shaped tool_results and now conflicts. Closing.

Note for maintainers: the only novel bit here not on master is the `except anthropic.BadRequestError` graceful-400 fallback — defense-in-depth behind caps that already contain the failure mode, and written against the obsolete AgentResult/dict shape. If wanted, it can be re-filed fresh against the current ToolResultBlock structure.

PR #106, opened 2026-05-28
fix(reporter): correct build_pipeline_output return annotation + docstring to tuple[str, int]

gadievron:
build_pipeline_output(...) was declared ``-> str`` with a docstring claiming it
returns the single output_path, but its sole return is
``return output_path, len(findings_data)`` — a tuple[str, int]. The annotation
and docstring went stale when the function gained a second return value; the
binding caller (cli.py cmd_report) was migrated to unpack the tuple but the
signature + docstring were not.

Fix is annotation/docstring-only (runtime return unchanged, since the caller
depends on the tuple):
  - ``-> str`` -> ``-> tuple[str, int]``
  - Returns: docstring now describes the (output_path, findings_count) tuple.

No cli.py edit: the sole binding caller already unpacks the tuple correctly and
now type-aligns with the corrected annotation.

New regression test pins the real contract (return is a 2-tuple) and asserts
the annotation + docstring agree with it. RED -> GREEN: 2 failed -> 3 passed.
ruff + ast clean.

### DIFF
--- a/libs/openant-core/core/reporter.py
+++ b/libs/openant-core/core/reporter.py
@@ -187,7 +187,7 @@ def build_pipeline_output(
     application_type: str = "web_app",
     processing_level: str | None = None,
     step_reports: list[dict] | None = None,
-) -> str:
+) -> tuple[str, int]:
     """Build ``pipeline_output.json`` from analysis results.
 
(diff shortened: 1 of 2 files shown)


gadievron:

PR #107, opened 2026-05-28
fix(js-parser): treat a zero-JS-file repo as an empty result, not a crash

gadievron:
A directory with zero JS-family source files made the JavaScript parser
abort the entire scan instead of returning a clean empty result.

repository_scanner.js succeeds and emits an empty file list (exit 0), but
run_typescript_analyzer hit `if not files: return False`, so
run_full_pipeline early-returned without results['success'], main() did
sys.exit(1), and _parse_javascript raised RuntimeError on the non-zero
exit (uncaught at the scan layer => whole scan aborts). The Python, Ruby
and Zig parsers all handle an empty repo gracefully; the JS guard was the
lone outlier promoting valid empty input to a fatal exit.

Mirror the zig graceful-empty pattern entirely within
parsers/javascript/test_pipeline.py: on zero files, write valid empty
analyzer_output.json / dataset.json / call_graph.json, record synthetic
successful stages, set self.empty_repo, and return True; run_full_pipeline
then skips the unit generator and reports success. _parse_javascript reads
units_count=0 and returns a valid empty ParseResult. parser_adapter.py and
zig/test_pipeline.py are correct as-is and left untouched.

Adds regression test tests/parsers/javascript/test_empty_repo.py
(RED before, GREEN after). Full openant-core suite: 219 passed, 22 skipped.

### DIFF
--- a/libs/openant-core/parsers/javascript/test_pipeline.py
+++ b/libs/openant-core/parsers/javascript/test_pipeline.py
@@ -126,6 +126,11 @@ class PipelineTest:
         self.analyzer_output_file = None
         self.dataset_file = None
 
+        # Set when the repository contains zero JS-family source files. An empty
+        # repo is a valid 0-unit result (mirroring the Python/Ruby/Zig parsers),
+        # not a failure, so the downstream unit-generator stage is skipped.
+        self.empty_repo = False
+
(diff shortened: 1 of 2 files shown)


gadievron:

PR #112, opened 2026-06-10
Python parser: fix 10 finder bugs

gadievron:
## Python parser: fix 10 finder bugs

Local-only branch `pr/python-parser` off master `368b559`. One of a coordinated 7-PR set fixing parser/reachability bugs found by the OpenAnt finder. **File-disjoint from the other 6 PRs** (no merge collision; any order).

Addresses 11 finder bug-ids (10 distinct fixes; 1 ride a same-PR canonical).

### Bugs fixed
- `[9]` py-nested-extraction (python/extraction_gap) — genuinely-new (no prior logged entry)
- `[10]` py-builtin-time (python/masking_floor) — genuinely-new (no prior logged entry)
- `[12]` py-dataflow-alias (python/dataflow_loss) — genuinely-new (no prior logged entry)
- `[21]` python-extraction_gap-498 (python/extraction_gap) — genuinely-new (no prior logged entry)
- `[23]` python-metadata-293-startline (python/metadata_correctness) — genuinely-new (no prior logged entry)
- `[27]` python-wrong_edge-333 (python/wrong_edge) — genuinely-new (no prior logged entry)
- `[34]` python-metadata-225-property (python/metadata_correctness) — genuinely-new (no prior logged entry)
- `[45]` python-extraction_gap-325-nestedclass (python/extraction_gap) — genuinely-new (no prior logged entry)
- `[46]` python-namespace_dispatch-264 (python/namespace_dispatch) — genuinely-new (no prior logged entry)
- `[47]` python-masking_floor-201 (python/masking_floor) — duplicate of [10] in this PR; closed by the same fix (no separate change)
- `[48]` python-metadata-232-testsubstr (python/metadata_correctness) — **Fixes BUG-059** (previously logged; was fixed only on an unmerged branch, still live on master 368b559)

### Dedup status (independent + judge, from raw)
9 genuinely-new · 1 duplicate-of-curated (cross-ref above) · 0 covered-by-curated · 1 intra-PR-duplicate. All re-confirmed STILL PRESENT on pristine master `368b559` (`git show 368b559:`).

### Tests
Verified GREEN in isolation off pristine `368b559`: **25 passed (tests/parsers/python/)**. New per-bug regression tests. (No `test_<lang>_schema_completeness.py` / `test_callgraph_symmetry.py` parser-infra in this PR — a follow-up; master has none either.)

### Notes
- TDD per bug (RED→GREEN); independent + judge verified each fix from raw against master.
- No behavior change outside the listed parser(s).

[...]

PR #115, opened 2026-06-11
fix(parsers/python): resolve self/cls-aliased method calls in the call graph

gadievron:
## Root cause
`CallGraphBuilder._resolve_call_node` (`libs/openant-core/parsers/python/call_graph_builder.py`) matched a method-call receiver only when it was the literal name `self` or `cls`. A receiver that is a *local alias* of self/cls — `obj = self; obj.method()`, or `alias = cls; alias.method()` inside a classmethod — fell through to `_resolve_module_call`, which finds no import/class for a plain local variable and returns `None`. The self/class-method call produced **no call-graph edge**, so the callee can look unreachable (a false negative in the reachable set).

## Reproduction
```python
class C:
    def target(self):
        return 1
    def caller(self):
        obj = self
        return obj.target()
```
Before: `call_graph["m.py:C.caller"]` is `[]` (the `C.caller -> C.target` edge is missing).

## How
- New `_collect_self_aliases(tree)`: scans the function body for local names **single-bound** to `self`/`cls` (a name assigned more than once, or ever rebound to anything else, is not treated as an alias — so no spurious edge).
- `_extract_calls_from_code` builds `self_aliases = {'self','cls'} | _collect_self_aliases(tree)` and passes it to `_resolve_call_node`.
- `_resolve_call_node` takes `self_aliases` (defaulting to `{'self','cls'}` so existing callers are unchanged) and resolves any `<alias>.method()` through the existing `_resolve_self_call`. The two prior `self`/`cls` branches collapse into one `obj.id in self_aliases` check.

This only **adds** edges that were previously dropped; it never removes one — so it raises reachability, never lowers it. The change is confined to the receiver-resolution branch (no new subsystem / data model).

## Regression test
`libs/openant-core/tests/parsers/python/test_call_graph_self_calls.py` (+62 lines):
- `test_self_alias_method_call_edge` — `obj = self; obj.target()` edge present.
- `test_cls_alias_method_call_edge` — `alias = cls; alias.target()` edge present.
- `test_reassigned_alias_does_not_force_self_edge` — soundness guard: a name reassigned away from self is NOT aliased.

RED (before fix):
```
FAILED tests/parsers/python/test_call_graph_self_calls.py::test_self_alias_method_call_edge
FAILED tests/parsers/python/test_call_graph_self_calls.py::test_cls_alias_method_call_edge
2 failed, 5 passed in 157.48s
```
GREEN (after fix), full `tests/parsers/python/`:
```
7 passed in 50.40s
```

## Compatibility
None. `_resolve_call_node`'s new parameter is optional and defaults to the previous behavior. No public API change.

## Notes
The canonical per-parser `tests/parsers/python/test_callgraph_symmetry.py` / `test_python_schema_completeness.py` do not yet exist in this repo (a pre-existing gap across all parsers, not introduced here); this fix extends the existing `test_call_graph_self_calls.py`.

## Author notes
- **Implementing lines:** `_collect_self_aliases` (new), the `obj.id in self_aliases` receiver branch, and the `self_aliases` thread-through in `_extract_calls_from_code`.
- **Input fixed:** `obj = self; obj.target()` — previously no `C.caller -> C.target` edge; now resolved.
- **Likely pushback:** "wrong edge if the alias is reassigned?" — no; only single-unconditional `= self`/`= cls` bindings count, pinned by `test_reassigned_alias_does_not_force_self_edge`.

## Coordination with open PRs
**#112** adds a `_build_alias_map` to the same resolution path, but for **function-value** aliases (`fn = helper; fn()`) — its docstring notes class/method targets are out of scope. This PR handles the distinct **self/cls** alias case (`obj = self; obj.method()`), which #112 leaves unresolved. The two are complementary but touch the same code; they should be unified rather than stacked. Happy to rebase onto #112 and fold the self/cls aliasing into its alias map.

### DIFF
--- a/libs/openant-core/parsers/python/call_graph_builder.py
+++ b/libs/openant-core/parsers/python/call_graph_builder.py
@@ -186,10 +186,13 @@ class CallGraphBuilder:
         # Local variable -> constructor-type map, so that `v = ClassName(); v.method()`
         # dispatches to the bound type's method.
         local_types = self._collect_local_types(tree)
+        # Receiver names that refer to the current instance/class: self/cls plus any
+        # local name single-bound to them (obj = self; obj.method()).
+        self_aliases = {'self', 'cls'} | self._collect_self_aliases(tree)
 
         for node in ast.walk(tree):
             if isinstance(node, ast.Call):
-                resolved = self._resolve_call_node(node, caller_file, caller_class, local_types)
+                resolved = self._resolve_call_node(node, caller_file, caller_class, local_types, self_aliases)
                 if resolved:
                     calls.add(resolved)
                 # Higher-order-function callbacks: a function reference passed as an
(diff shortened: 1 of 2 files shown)


gadievron:

This is an explicit follow-up to **#112** (same `parsers/python/call_graph_builder.py`) — consider adding `Depends-on: #112`. **Union**: `_resolve_call_node` must thread all three — `local_types` + `aliases` (#112) **and** `self_aliases` (this PR). Whichever lands second will conflict; keep all three params.

gadievron:

Python parser tests green (14 + 7). Real union onto master (kept master local_types + this PR self/cls aliases).

gadievron:

PR #116, opened 2026-06-12
fix(parsers/python): match imports by module components in both directions

gadievron:
## Root cause
`_resolve_import` Strategy 2 (`parsers/python/call_graph_builder.py`) bound an imported function to a file whose dotted module path string-`endswith` the imported module (in either direction). That crosses component boundaries: `from utils import helper` matched `extra_utils.py:helper` because `'extra_utils'.endswith('utils')`, and `from auth import login` matched `extra_auth.py:login`. No Python module layout produces those — they are false call-graph edges to the wrong module, and they inflate the reachable set with units that are never really called.

## Fix
Match by dotted **components**, in **either direction**:
- the imported module is a component-suffix of the file path — a repo-root prefix on the file (`src/pkg/auth.py` for `from pkg.auth import ...`); or
- the file path is a component-suffix of the imported module — the repo-IS-the-package self-import (`auth.py` == `myapp/auth.py` recorded shallow via `relative_to(repo_path)`, for `from myapp.auth import ...`).

Both are real layouts and are kept. Only the substring-crossing matches are dropped. Relative/bare imports (`from . import x`, no module) keep the prior name-only behavior.

## Reachability / recall safety
An adversarial review caught that an earlier forward-only version dropped the reverse direction (repo-is-package self-imports) — a **recall regression**. Because the reachability filter keeps only reachable units, a dropped real edge silently removes attack surface. So an edge that *could* be real — `from pkg.auth import login` to a top-level `auth.py`, which is structurally identical to a self-import and indistinguishable from the data — is **kept**, not forbidden. Net: strictly fewer false edges than master, never fewer real ones.

## Regression test
`tests/parsers/python/test_call_graph_import_resolution.py` — RED→GREEN:
- 2 precision drops: substring-crossing `extra_utils`/`utils`, `extra_auth`/`auth`.
- 6 recall guards: exact package path, repo-root prefix, top-level, repo-is-package self-import, nested reverse-prefix, relative import.

Full Python suite: **12 passed**.

## Compatibility
None. Resolution only; the change is strictly more precise on false edges and recall-preserving on real ones.

## Coordination with open PRs
This edits the `_resolve_import` Strategy-2 block in `parsers/python/call_graph_builder.py`, which **#72** also reworks (it adds a `seen` param for `__init__.py` re-exports) and **#112** also touches. Neither fixes this bug — both still carry the bare `module_path.endswith(expected_module) or expected_module.endswith(module_path)`. Expect a merge conflict in that block; happy to rebase onto whichever of #72/#112 lands first.

### DIFF
--- a/libs/openant-core/parsers/python/call_graph_builder.py
+++ b/libs/openant-core/parsers/python/call_graph_builder.py
@@ -331,11 +331,27 @@ class CallGraphBuilder:
         for func_id in self.functions:
             func_data = self.functions[func_id]
             if func_data.get('name') == parts[-1]:
-                # Check if file path matches module path
+                # Check if file path matches module path, by dotted COMPONENTS in EITHER
+                # direction:
+                #   - imported module is a component-suffix of the file path — a repo-root
+                #     prefix on the file (e.g. `src/pkg/auth.py` for `from pkg.auth ...`); OR
+                #   - the file path is a component-suffix of the imported module — the
+                #     repo-IS-the-package self-import (e.g. `auth.py` == `myapp/auth.py`
+                #     recorded shallow via relative_to(repo_path), for `from myapp.auth ...`).
+                # Both are real layouts, so both are kept (recall). The OLD bare `endswith`
+                # ALSO matched ACROSS component boundaries — `'extra_utils'.endswith('utils')`
+                # bound `from utils import helper` to `extra_utils.py`. No module layout
+                # produces that, so those substring-crossing matches are the ONLY edges
+                # dropped. `not em` keeps the prior name-only behavior for relative/bare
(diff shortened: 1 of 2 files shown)


gadievron:

`parsers/python/call_graph_builder.py` here overlaps **#112** (same Strategy-2 import-resolution block). Resolve as a **union**, not accept-ours/theirs:
- keep **#112**'s name check `func_data.get('name') == func_name` (match the *called* name, not the module-path tail `parts[-1]`);
- keep **this PR**'s component-based module match in both directions (replacing the old cross-boundary `endswith`).

Whichever lands second will conflict on that one `if` line; keep both fixes.

gadievron:

PR #94, opened 2026-05-28
fix(context): anchor entry-point path exclusions on components, not substrings

gadievron:
detect_entry_points() skipped candidate files via `any(token in str(path))`
against the full absolute path at two sites (Python branch + JS/TS branch).
Short tokens ('test','tests','venv','dist','build') matched as substrings, so:
  - real entry points whose name merely contains a token were skipped
    (protest_api.py, latest/main.py, redistribute.js, buildings/);
  - worse, a token in any ANCESTOR of repo_path (a CI checkout under
    /build/ or /latest/) suppressed EVERY file -> detect_entry_points
    returned '' -> downstream ApplicationType misclassification.

Fix: match on relative-path COMPONENTS via a shared _path_excluded() helper
(set-intersection on `relative_to(repo_path).parts`) plus anchored test-file
detection (test_*.py / *_test.py). Genuine node_modules/venv/dist/build dirs
and real test files stay excluded.

Tests: tests/test_application_context_entry_points_anchored.py (4 cases:
substring-FP python, parent-path-token, substring-FP js, genuine-exclusions
preserved). RED 4 failed -> GREEN 4 passed.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #118, opened 2026-06-12
feat(context): gate local-only suppression on trust_boundaries, not just app type

gadievron:
### Problem
The context generator sets `requires_remote_trigger: False` for every `library`/`cli_tool`. The Stage 1 and Stage 2 prompts (`prompts/vulnerability_analysis.py`, `prompts/verification_prompts.py`) then emit *"This is a CLI tool/library… only flag vulnerabilities exploitable by a REMOTE attacker, not local users."* That framing is correct for a library whose inputs are all operator-controlled — but **wrong for a data-processing library** (parser, deserializer, codec) whose untrusted *input data* is the attack surface even with no network listener. It tells the analyzer to discard exactly that bug class.

Observed on tree-sitter (a C parsing library): every parser memory-safety finding was suppressed by this clause until `requires_remote_trigger` was hand-set.

### Fix
Add `ApplicationContext.suppress_local_only()` — keep the suppression **only when no trust boundary is `untrusted`**. This reuses the already-captured `trust_boundaries` (values: `trusted`/`semi_trusted`/`untrusted`); no new schema field. Gate all 7 emission sites (6 live in `prompts/`, 1 display in `context/`). Generator instruction reworded to set `requires_remote_trigger` from whether untrusted input is processed, not blanket-false per type (defense-in-depth; the gate holds even if the model keeps the old behavior).

- A library with all-`trusted` boundaries → **unchanged** (still suppressed; genuine no-attack-surface libraries are not affected).
- A library with any `untrusted` boundary (e.g. tree-sitter's `source_code_being_parsed: untrusted`) → now analyzed.

### Scope note
This fixes the `requires_remote_trigger` suppression mechanism only. A *separate* over-suppression — the generator sometimes emits over-broad `not_a_vulnerability` entries (e.g. "memory operations in C are fine") — is not addressed here and is worth a follow-up.

### Not a security fix
This is an analyzer-targeting/coverage change (it changes which findings the LLM is told to consider), not a code-vulnerability patch.

### Tests
9 in `tests/test_threat_model_untrusted_input_gate.py`: `suppress_local_only()` unit cases (incl. case-insensitive `untrusted` match, since the values are LLM-generated) + Stage 1 and Stage 2 formatter present/absent for all-trusted vs untrusted-input contexts. No existing test pinned the prior behavior (verified: `grep -rn suppress_local_only tests/` = 1 file).

### Depends-on
None (standalone; the suppression clause is pre-existing on master). Co-exists with #94's `application_context.py` changes (different region; no conflict).

### DIFF
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -145,6 +145,26 @@ class ApplicationContext:
         """Get detailed information about this application type."""
         return APPLICATION_TYPE_INFO.get(self.application_type, {})
 
+    def suppress_local_only(self) -> bool:
+        """Whether to tell the analyzer to flag only REMOTE-attacker vulnerabilities.
+
+        The "local users have access, only flag remote" framing is correct for a
+        CLI/library whose inputs are all operator-controlled. But a data-processing
+        library (parser, deserializer, codec) takes UNTRUSTED INPUT DATA — and that
+        data crossing into the code IS the attack surface, even with no network
+        listener. ``requires_remote_trigger`` alone (False for every library) would
+        suppress exactly those bugs. Gate on the already-captured ``trust_boundaries``:
+        if any input source is ``untrusted``, do NOT suppress, regardless of type.
+        """
+        if self.requires_remote_trigger:
+            return False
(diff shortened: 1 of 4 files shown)


gadievron:

PR #120, opened 2026-06-12
feat(reachability): library-mode for all parsers + silent-blackout warning

gadievron:
## feat(reachability): library-mode for all parsers + silent-blackout warning

Two coupled changes that make OpenAnt usable on data-processing libraries (parsers, deserializers, codecs), where the reachability filter otherwise prunes the entire public-API surface. Both touch the same files (`core/parser_adapter.py`, `utilities/agentic_enhancer/entry_point_detector.py`, the six subprocess pipelines), so they ship together.

### Base / dependency
This PR is based on `staging/parser-fix-stack` (the integrated #70–#117 parser-fix stack), not `master`. It extends #117's library-mode and complements #75's zero-seed net, and builds on the zig parser state from #75/#85/#87/#110 — so it only applies on top of that stack. **Depends-on: #75, #117** (and the parser-fix stack generally). Retarget the base to `master` once the stack lands.

---

### Part 1 — silent-blackout warning (advisory; never changes filtering)
**Problem:** the #75 zero-seed net only fires at *exactly* zero entry points. A library trips a handful of incidental seeds (code merely containing an input-reading pattern), so a 90%+ reduction looks like a successful filter while the public-API core was dropped. tree-sitter `lib/`: 661 → 24 units (all `wasm_store.c`); `parser.c`/`lexer.c`/`stack.c`/`subtree.c` removed, reported as 96% "reduction" with no warning.

**Fix:** `blackout_warning()` in `utilities/agentic_enhancer` (shared by all filter sites). Warns when 0 of N kept, OR ≥90% pruned with no *structural* entry point (every seed is an incidental `input_pattern` match, never `unit_type:`/`name:main`/`decorator:`). Suppressed under `library_mode`. Wired into `core/parser_adapter.py` + all six subprocess pipelines. **Advisory only — never changes which units are kept.** Calibration: silent on a real app (Arkime C: 63%, real `main`/`cli_handler` seeds), fires on a library.

### Part 2 — opt-in library-mode for all parsers + CLI
**Problem:** #117 added `--library-mode` (seed the exported public API) only on the Python path. The other six parsers run as subprocesses with their own reachability copy, so a C/JS/Go/Ruby/PHP/Zig library still collapsed.

**Fix:** centralize the seed logic into `library_seed_ids()` (handles both `is_exported` snake_case on disk and `isExported` camelCase from the pipelines' normalize). Thread an opt-in `library_mode` through all four surfaces of each subprocess parser (dispatch → subprocess cmd → `--library-mode` argparse → seed union before the BFS), plus `scan_repository` and the `openant parse`/`scan` CLI flags. Union-only — never drops a structural entry point, so app scans are unaffected. #117's Python path is behavior-identical.

**Verified end-to-end (parse, no API):**

| Project | lang | without `--library-mode` | with `--library-mode` |
|---|---|---|---|
| tree-sitter `lib/` | C | 24/661 (blackout) | 550/661 (parser core reachable) |
| google/uuid | Go | 0/78 | 77/78 |
| unicode-display_width | Ruby | 0/13 | 13/13 |
| slugify | JS | 0/8 | 7/8 |
| nikic/iter | PHP | 0/120 | 119/120 |
| ziglibs/known-folders | Zig | 9/21 | 21/21 |

Arkime (a real app) is unchanged either way. (Zig CLI selection needs the companion `fix(cli)` PR to be reachable via `openant parse --language zig`.)

### Not a security fix
Tooling/coverage changes (they change which code reaches the analyzer), not code-vulnerability patches.

### Tests
15 new: `tests/test_blackout_warning.py` (7 — both triggers + structural/library-mode suppression) and `tests/test_library_seed_ids.py` (8 — both casings + name heuristic). #117's `test_library_mode_reachability.py` unchanged (behavior-preserving). Full `libs/openant-core` suite: 624 passed, 22 skipped.

### DIFF
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -254,6 +254,7 @@ def apply_reachability_filter(
     _epd = _load_module("entry_point_detector", "entry_point_detector.py")
     _ra = _load_module("reachability_analyzer", "reachability_analyzer.py")
     EntryPointDetector = _epd.EntryPointDetector
+    blackout_warning = _epd.blackout_warning
     ReachabilityAnalyzer = _ra.ReachabilityAnalyzer
 
     call_graph_path = os.path.join(output_dir, "call_graph.json")
@@ -348,6 +349,12 @@ def apply_reachability_filter(
         f"({reduction_pct}% reduction)",
         file=sys.stderr,
     )
+
+    _blackout = blackout_warning(detector.entry_point_details, original_count,
+                                 len(filtered_units), library_mode=library_mode)
+    if _blackout:
+        dataset["metadata"]["reachability_filter"]["warning"] = _blackout
+        print(f"  [Warning] {_blackout}", file=sys.stderr)
 
     # Warn about unimplemented higher-level filters
     if processing_level == "codeql":
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -80,6 +80,7 @@ def parse_repository(
     name: str = None,
     diff_manifest: str | None = None,
     fresh: bool = False,
+    library_mode: bool = False,
 ) -> ParseResult:
     """Parse a repository into an OpenAnt dataset.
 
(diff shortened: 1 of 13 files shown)


gadievron:

Base retargeted staging→master. Kept BOTH master/#133 --fresh AND this PR's --library-mode (a blind take would have dropped --fresh); #75 zero-seed net retained. Reachability tests green (15 + 87).

gadievron:

PR #122, opened 2026-06-14
fix(parsers/php,ruby): keep both same-name defs from conditional/guarded branches

gadievron:
Companion to #121 (the C/C++ same-`(file,name)` collision fix). A blind independent re-derivation + an adversarial judge — both running the real `php`/`ruby`/`node` interpreters — established that of the non-C parsers, **only PHP and Ruby** have a *harmful* same-name collision; JS and Python do not (JS keeps the last-wins copy / skips conditional defs; Python already keeps both via a `#L<line>` suffix). This PR fixes PHP and Ruby; it deliberately does **not** touch JS/Python.

## The bug (PHP + Ruby)

A PHP `function`/Ruby `def` in a **mutually-exclusive conditional branch** is runtime-reachable depending on the environment, and the **earlier-in-source branch is often the live one**:

```php
if (!function_exists('h')) { function h(){ return real(); } }      // runs when h is undefined
if (!function_exists('h')) { function h(){ return fallback(); } }  // skipped
```
```ruby
if COND then def k; if_branch; end else def k; else_branch; end end
```

Both extractors keyed on `qualified_name` with a plain keep-last store, so the **later (often dead) branch silently overwrote the earlier (often live) one** — a false negative for a SAST tool. Confirmed: `php` prints `real()` / `ruby` runs `if_branch`, but the extractor kept the other branch.

## The fix — keep BOTH (not prefer-first/larger)

Which branch is live is environment-dependent, so neither is statically dead. The fix keeps **both** via a deterministic `#L<line>` suffix (earlier-in-source keeps the clean id) — the exact `_store_function` strategy the **Python** extractor already uses, which is precisely why Python was found immune. Ported into the PHP (`:558`, `:600`) and Ruby (`:549`, `:613`) store sites.

- **Collision-only:** a unique name keeps its byte-identical `path:name` id (pinned by a `test_unique_name_id_unchanged` per language).
- **PHP:** plain redefinition is a fatal error, so the only same-name duplicates are conditional/guarded — keep-both is purely beneficial there.
- **Ruby:** unconditional reopening is last-wins at runtime; keeping both there is the same benign tradeoff Python accepts (Stage 2 attacker-simulation filters dead overrides).

## Tests

New: `tests/parsers/php/test_php_conditional_collision.py`, `tests/parsers/ruby/test_ruby_conditional_collision.py`.

```
$ pytest tests/parsers/php/ tests/parsers/ruby/ -q
83 passed
```
Full repo suite: 606 passed, 22 skipped.

### DIFF
--- a/libs/openant-core/parsers/php/function_extractor.py
+++ b/libs/openant-core/parsers/php/function_extractor.py
@@ -482,7 +482,7 @@ class FunctionExtractor:
             'unit_type': unit_type,
         }
 
-        self.functions[func_id] = func_data
+        self._store_function(func_id, func_data)
         self.stats['total_functions'] += 1
 
         if class_name:
@@ -524,7 +524,7 @@ class FunctionExtractor:
 
         func_id = f"{relative_path}:{qualified_name}"
 
-        self.functions[func_id] = {
+        self._store_function(func_id, {
             'name': name,
             'qualified_name': qualified_name,
             'file_path': relative_path,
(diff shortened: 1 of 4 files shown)


gadievron:

Base retargeted staging→master. PHP+Ruby tests green (47).

gadievron:

PR #124, opened 2026-06-14
fix(parsers/js): extract module-level block-scoped functions

gadievron:
## What
Extract module-level block-scoped function declarations (inside `if`/`try`/`for`/`switch`/bare blocks), which the JS/TS analyzer silently dropped.

## Root cause
`extractFunctionsFromFile` and `buildCallGraphForFile` both enumerated functions via `sourceFile.getFunctions()` — which returns only top-level declarations. A function declared inside any block was therefore absent from BOTH the function inventory AND the call graph (`typescript_analyzer.js:231` / `:1253`). The resolver's standalone-function strategy had the same blind spot (`:1588`). For a SAST tool this is a recall gap: a function gated behind `if (process.env.X) {...}` or defined in a `catch` fallback is never a unit, never a reachability node, never an entry point — so it is never analyzed, and its call to (e.g.) `eval`/`child_process.exec` is invisible.

## Reproduction
```
$ printf 'if (c) { function vuln(){ return eval(x); } }\nfunction top(){}' > /tmp/r/a.js
$ node typescript_analyzer.js /tmp/r /tmp/r/a.js | jq '.functions | keys'
["a.js:top"]            # vuln dropped
```

## How
- New `_moduleLevelFunctionNodes` / `_moduleLevelFunctionEntries`: descend the AST for `FunctionDeclaration`s and keep those with NO function-like ancestor — surfacing module-level block-scoped functions while omitting functions nested inside another function/method (their text already rides inside the parent unit) and inside a class `static {}` block (block-scoped to the initializer, callable nowhere).
- The inventory builder, the call-graph builder, and the resolver all iterate this one list, so their keys match exactly — a block function gets its real outgoing edges (not a backstop `[]`) and its incoming calls resolve, so it is reachable.
- Sibling-block same-name functions (`if(c){function f(){A}}else{function f(){B}}`) are both runtime-reachable, so keep BOTH via a colon-free `#L<line>` suffix; collision-only, so unique names keep the byte-identical `path:name` id.

Scope: deliberately NOT touched — functions nested inside another function/method (text rides in the parent), `static {}`-block functions (uncallable), arrow/function-expression assignments inside blocks (separate variable-declaration extraction path; follow-up), classes-in-blocks (follow-up). The resolver's constructor-pattern and cross-file-import strategies stay top-level (a block-scoped function can't be exported/imported).

## Regression test
`tests/parsers/javascript/test_block_scoped_functions.py` — 8 block shapes surfaced with real edges; incoming-edge resolution; sibling-block keep-both; static-block and nested-in-function omitted; the `len(callGraph) == len(functions)` invariant; unique-name id unchanged.

RED (before fix):
```
9 failed  (the 8 block shapes + sibling keep-both)
```
GREEN (after fix):
```
$ pytest tests/parsers/javascript/test_block_scoped_functions.py -q
14 passed
$ pytest tests/parsers/javascript/ -q
66 passed
```
Full repo suite: 615 passed, 22 skipped. express smoke: 7 files, callGraph keys == functions keys, 0 inflation (no block collisions in idiomatic code).

## Compatibility
No id change for code without block-scoped functions: `getDescendantsOfKind` returns top-level declarations in the same document order as `getFunctions()`, so last-writer-wins resolves identically (verified — express output unchanged). New units are additive; the call-graph/inventory key invariant is preserved.

Note: `tests/parsers/javascript/test_callgraph_symmetry.py` does not yet exist at the canonical path (a pre-existing repo-wide gap); this PR's regression test asserts the functions↔callGraph key invariant inline as `test_invariant_callgraph_keys_equal_functions_keys`.

### DIFF
--- a/libs/openant-core/parsers/javascript/typescript_analyzer.js
+++ b/libs/openant-core/parsers/javascript/typescript_analyzer.js
@@ -219,6 +219,68 @@ class TypeScriptAnalyzer {
   /**
    * Extract all functions/methods from a source file
    */
+  // Module-level FunctionDeclarations: top-level OR nested only inside blocks /
+  // control-flow statements (if/else, try/catch, for/while, switch, bare {}) —
+  // NOT inside another function/method/accessor (their text already rides inside
+  // the parent unit) and NOT inside a class `static {}` block (block-scoped to
+  // the initializer, callable nowhere). `getFunctions()` returned only top-level
+  // declarations, so block-scoped functions were silently dropped from both the
+  // inventory and the call graph. Returns [{node, id}] with COLLISION-ONLY
+  // `#L<line>` disambiguation (sibling-block same-name functions are both
+  // runtime-reachable, so keep both; unique names keep the plain `path:name`
+  // id). Both the inventory and call-graph builders iterate this same list so
+  // their keys match exactly (the `len(callGraph) === len(functions)` lockstep).
+  // The named FunctionDeclaration nodes that are module-level (top-level OR only
+  // block-nested), excluding function-nested and static-block-scoped ones. Used
+  // by the inventory builder, the call-graph builder, AND the resolver — all
+  // three must see the same set so a block function is a unit, has its own
+  // edges, and is resolvable as a call target.
+  _moduleLevelFunctionNodes(sourceFile) {
+    const STOP = new Set([
+      ts.SyntaxKind.FunctionDeclaration,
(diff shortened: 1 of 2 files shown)


gadievron:

Base retargeted staging→master. Independent of Wave 1 (clean rebase). JS tests run in CI (no node_modules locally).

gadievron:

PR #111, opened 2026-06-10
PHP parser: fix 8 finder bugs

gadievron:
## PHP parser: fix 8 finder bugs

Local-only branch `pr/php-parser` off master `368b559`. One of a coordinated 7-PR set fixing parser/reachability bugs found by the OpenAnt finder. **File-disjoint from the other 6 PRs** (no merge collision; any order).

Addresses 9 finder bug-ids (8 distinct fixes; 1 ride a same-PR canonical).

### Bugs fixed
- `[7]` php-nested-extraction (php/extraction_gap) — genuinely-new (no prior logged entry)
- `[20]` php-dataflow_loss-199 (php/dataflow_loss) — **Fixes BUG-011** (previously logged; was fixed only on an unmerged branch, still live on master 368b559)
- `[25]` php-metadata-420-attrstartline (php/metadata_correctness) — genuinely-new (no prior logged entry)
- `[28]` php-schema_field_drift-239-isstatic (php/schema_field_drift) — genuinely-new (no prior logged entry)
- `[36]` php-extraction_gap-407-enum (php/extraction_gap) — genuinely-new (no prior logged entry)
- `[43]` php-unit_metadata-163-namespace (php/unit_metadata_correctness) — genuinely-new (no prior logged entry)
- `[52]` php-namespace_dispatch-253-parent (php/namespace_dispatch) — **Fixes BUG-022** (previously logged; was fixed only on an unmerged branch, still live on master 368b559)
- `[53]` php-extraction_gap-249-enum (php/extraction_gap) — duplicate of [36] in this PR; closed by the same fix (no separate change)
- `[54]` php-metadata-390-bracelessns (php/metadata_correctness) — genuinely-new (no prior logged entry)

### Dedup status (independent + judge, from raw)
6 genuinely-new · 2 duplicate-of-curated (cross-ref above) · 0 covered-by-curated · 1 intra-PR-duplicate. All re-confirmed STILL PRESENT on pristine master `368b559` (`git show 368b559:`).

### Tests
Verified GREEN in isolation off pristine `368b559`: **16 passed (tests/parsers/php/)**. New per-bug regression tests + a `test_<lang>_schema_completeness.py` field-contract guard.

### Notes
- TDD per bug (RED→GREEN); independent + judge verified each fix from raw against master.
- No behavior change outside the listed parser(s).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-06-10
Zig parser: fix 6 finder bugs

gadievron:
## Zig parser: fix 6 finder bugs

Local-only branch `pr/zig-parser` off master `368b559`. One of a coordinated 7-PR set fixing parser/reachability bugs found by the OpenAnt finder. **File-disjoint from the other 6 PRs** (no merge collision; any order).

Addresses 6 finder bug-ids (6 distinct fixes; 0 ride a same-PR canonical).

### Bugs fixed
- `[3]` zig-local-type-dispatch (zig/namespace_dispatch) — same defect as **BUG-044 (zig leg)** (characterized but unmerged; this lands it on master)
- `[16]` zig-extraction_gap-147 (zig/extraction_gap) — genuinely-new (no prior logged entry)
- `[17]` zig-builtin_filter_leak-234 (zig/builtin_filter_leak) — genuinely-new (no prior logged entry)
- `[26]` zig-metadata-265-test (zig/metadata_correctness) — genuinely-new (no prior logged entry)
- `[37]` zig-extraction_gap-106-structmethod (zig/extraction_gap) — same defect as **BUG-095/096** (characterized but unmerged; this lands it on master)
- `[41]` zig-dataflow_loss-298-alias (zig/dataflow_loss) — genuinely-new (no prior logged entry)

### Dedup status (independent + judge, from raw)
4 genuinely-new · 0 duplicate-of-curated (cross-ref above) · 2 covered-by-curated · 0 intra-PR-duplicate. All re-confirmed STILL PRESENT on pristine master `368b559` (`git show 368b559:`).

### Tests
Verified GREEN in isolation off pristine `368b559`: **12 passed (tests/parsers/zig/)**. New per-bug regression tests. (No `test_<lang>_schema_completeness.py` / `test_callgraph_symmetry.py` parser-infra in this PR — a follow-up; master has none either.)

### Notes
- TDD per bug (RED→GREEN); independent + judge verified each fix from raw against master.
- No behavior change outside the listed parser(s).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #126, opened 2026-06-17
fix(go-parser): key generic-type methods on base type, not "unknown"

gadievron:
## Summary
The Go parser keys methods on **generic types** (`func (s *Stack[T]) Push(...)`, `func (p *Pair[K,V]) Key(...)`) under a bogus class name **`unknown`** instead of the real type, because the receiver-type stringifier `typeToString` has no case for Go's generic-instantiation AST nodes (`*ast.IndexExpr` / `*ast.IndexListExpr`) and falls through to `default: return "unknown"`.

Impact (all confirmed on this branch's base):
- **Wrong unit id** — `m.go:unknown.Push` instead of `m.go:Stack.Push`.
- **Silent data loss** — distinct generic types' same-named methods (`Stack[T].Len`, `Queue[T].Len`) collapse to one id `m.go:unknown.Len`; the store (`output.Functions[id] = info`) overwrites, so one real unit is dropped.
- **Dropped call edges** — generic call sites `Gen[K,V]()` (`*ast.IndexListExpr`) and generic method calls `obj.M[T]()` were not recovered by `analyzeCallExpr`.

Generic types are idiomatic Go since 1.18, so this hits modern codebases broadly.

## Changes
- **`extractor.go` `typeToString`**: add `case *ast.IndexExpr` and `case *ast.IndexListExpr`, each recursing into `t.X` so the class key is the bare base type (`Stack`, `Pair`).
- **`callgraph.go` `analyzeCallExpr`**: unwrap a leading `IndexExpr`/`IndexListExpr` then reuse the existing `Ident`/`SelectorExpr` classification — fixes all four generic call shapes and recovers the receiver for generic method calls.
- **`extractor.go` `duplicateIDWarning`**: emit an `os.Stderr` warning when a `funcID` collides on store, so any future class-key collapse fails loudly instead of silently overwriting. JSON output shape unchanged.

## Tests
5 new Go tests: generic-receiver class keys (pointer/value/multi-param), the no-collision data-loss regression (`Stack.Len` + `Queue.Len` stay distinct), generic call recovery, and the dup-guard. `go test ./...` green; `gofmt`/`go vet` clean.

## Scope & related work (please read before requesting "why not Zig/PHP too?")
This PR is **Go-only by design and is self-contained** — it touches only `parsers/go/go_parser/` and has **no cross-parser dependency**. The Go parser uses the standard-library `go/ast`, whose node types are stable, so this lands cleanly on `master`.

The same class-key-collapse family also affects the **Zig and PHP** parsers. Those fixes are in a **separate PR** that depends on the parser-rewrite PRs **#110** (Zig) and **#111** (PHP) and so targets the `staging/parser-fix-stack` integration branch, not `master`. This Go PR is **independent** of all of that — `go/ast` node types are stable — so do not block it on them.

## Notes
- For the class key the `[T]` type arguments are intentionally dropped (we want `Stack`, not `Stack[T]`); the receiver string is a display field, not used for resolution.
- Known pre-existing, out-of-scope edges (not introduced here): a slice/map-index call `funcs[i]()` reports the container identifier as the call name; a parenthesized receiver `func ((T)) M()` still hits `default` (now observable via the dup-guard).


gadievron:

PR #127, opened 2026-06-17
fix(zig,php parsers): attribute generic-container & anonymous-class methods correctly

gadievron:
# Zig + PHP: attribute generic-container & anonymous-class methods correctly

Fixes the class-key collapse bug family in the **Zig** and **PHP** parsers (cross-parser siblings of the Go generic-receiver fix).

---

## ⚠️ MERGE TARGET — this PR targets `staging/parser-fix-stack`, NOT `master` (humans and agents)

This change is part of the parser-fix stack and **must merge into the integration branch `staging/parser-fix-stack`**, which already contains every prerequisite it needs. It reaches `master` later, together with the rest of the stack — not on its own.

**Do NOT retarget this at `master`.** `master` is missing all three prerequisites below — the two parser rewrites AND the grammar pin:

```
Depends-on: #110   (Zig parser: fix 6 finder bugs — realigns the parser to current tree-sitter-zig node names)
Depends-on: #111   (PHP parser: fix 8 finder bugs — reworks _extract_functions_from_tree)
Depends-on: the tree-sitter-zig>=1.1.2 grammar pin — carried by the stack; absent from master's requirements.txt
```

On `master` the fix is inert and its regression tests fail (Zig) / methods aren't extracted (PHP) — see "Why" below. Because the stack carries all three prerequisites, on `staging/parser-fix-stack` this PR is exactly 4 files and fully green (73 tests pass).

### Why the order matters (verify it yourself)
`master`'s Zig parser matches **stale tree-sitter node names** (`VarDecl`, `container_decl`, `FnProto`); the installed `tree-sitter-zig` (>=1.1.2) emits `variable_declaration` / `struct_declaration` / `function_declaration`. #110 realigns the parser to the new names. Until then, Zig struct/method extraction is dead, and this fix (written against the new names) cannot work:
```
git show origin/master:libs/openant-core/parsers/zig/function_extractor.py | grep -c 'struct_declaration'
# -> 0   (the node names this fix depends on don't exist on master; #110 adds them)
```
PHP similarly needs #111's reworked `_extract_functions_from_tree` traversal.

**Self-guard:** the Zig regression test `skip`s with this exact explanation when run on a base that lacks the prerequisite, so an agent/human running the suite on the wrong base sees *why* instead of a cryptic failure (it does **not** silently pass — it reports SKIPPED with the reason).

---

## Summary
- **Zig** (`parsers/zig/function_extractor.py`): methods inside the generic-container idiom `fn List(comptime T: type) type { return struct { fn push() ... }; }` were emitted as bare top-level functions (`class_name=None`) and distinct containers' same-named methods collided. Fix: `_returns_type()` + thread the container fn name as struct context → `List.push`.
- **PHP** (`parsers/php/function_extractor.py`): methods inside `new class {}` anonymous classes had `class_name=None` and collided across distinct anon classes (silent overwrite / data loss). Fix: an `anonymous_class` branch synthesizing a stable `class@anonymous:<line>:<col>` identity (line+col so multiple anon classes on one line stay distinct).

## Tests
5 new (qualified-name + collision regressions, incl. same-line anon classes). On the prerequisite base (`staging/parser-fix-stack`) the full zig+php suites are green — `73 passed`. Independent + judge reviewed.

## Related
Companion **Go** PR (`fix(go-parser): key generic-type methods on base type`) fixes the same bug family for the Go parser. It is **independent** — Go uses the stdlib `go/ast` (stable node types), so it lands on `master` on its own with no dependency in either direction.

### DIFF
--- a/libs/openant-core/parsers/php/function_extractor.py
+++ b/libs/openant-core/parsers/php/function_extractor.py
@@ -437,6 +437,52 @@ class FunctionExtractor:
                         stack.append((child, class_name, new_namespace_name))
                 continue  # Don't walk children again
 
+            elif node.type == 'anonymous_class':
+                # `new class { ... }` (PHP 7+) has no source name. Without a synthetic
+                # identity its methods fall through the catch-all else with the OUTER
+                # class_name (None at top level), so they're keyed as bare functions and
+                # two distinct anonymous classes that both define e.g. handle() collide on
+                # one id (the later silently overwrites the earlier). Synthesize a stable,
+                # location-based name so each anonymous class is distinct and its methods
+                # are qualified (class@anonymous:<line>:<col>.method). Line AND column are
+                # both needed: two `new class {}` on one physical line share a start line,
+                # so column is what keeps them distinct (else they'd still collide).
+                anon_name = (
+                    f"class@anonymous:{node.start_point[0] + 1}:{node.start_point[1]}"
+                )
+                body_node = None
+                for child in node.children:
+                    if child.type == 'declaration_list':
+                        body_node = child
+                        break
+
+                if body_node:
+                    methods = []
+                    for child in body_node.children:
+                        if child.type == 'method_declaration':
+                            mname = self._get_function_name(child, source)
+                            if mname:
+                                if self._is_static_method(child, source):
+                                    methods.append(f"static:{mname}")
+                                else:
+                                    methods.append(mname)
+
+                    self.classes[f"{relative_path}:{anon_name}"] = {
+                        'name': anon_name,
+                        'file_path': relative_path,
+                        'start_line': node.start_point[0] + 1,
+                        'end_line': node.end_point[0] + 1,
+                        'methods': methods,
+                        'superclass': None,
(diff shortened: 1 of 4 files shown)


gadievron:
Heads-up for reviewers/agents: two follow-up fix PRs are **stacked on this branch** (`pr/zig-php-generic-anon`) and should merge **after** #127 (or be rebased onto master once this lands):

- **#130** `fix(parsers/c): guard macro-alias resolution against cyclic #defines` — this branch's C-parser rework still carries the unguarded macro-alias recursion in `_resolve_call`; #130 adds the cycle guard on top of it.
- **#131** `fix(parsers/php): resolve $this->/self:: calls into used traits` — builds on the `<?php` tagless-parse and `relative_scope` fixes already in this branch, adding only the net-new class→trait composition index.

Both are draft until #127 merges.

gadievron:

Base retargeted staging→master. PHP+Zig tests green (45). Clean rebase — master already carries the #110/#111 prerequisites.

gadievron:

PR #130, opened 2026-06-18
fix(parsers/c): guard macro-alias resolution against cyclic #defines

gadievron:
# fix(parsers/c): guard macro-alias resolution against cyclic #defines

**Base:** `pr/zig-php-generic-anon` · **Depends-on:** #127 · **Type:** bug fix · **Finding:** F4 (HIGH)

> Stacked on #127 because that branch's C-parser rework already restructures
> `_resolve_call` (adds `receiver_type`/`is_member`, `_resolve_same_file`); the
> macro-alias recursion it carries is still unguarded. Targeting master instead
> would hard-conflict with #127 in this same function. This layers only the cycle
> guard on top of #127's signature.

## What
`parsers/c/call_graph_builder.py` — `_resolve_call` threads an `_alias_chain`
visited-set through its macro-alias recursion, so a cyclic alias resolves to
`None` instead of recursing forever.

## Why
`macro_aliases` is built from function-like `#define`s. A cyclic pair —
`#define A(x) B(x)` / `#define B(x) A(x)` → `{"A":"B","B":"A"}` — made
`_resolve_call("A")` recurse `A→B→A→…` with no guard → `RecursionError`.
`build_call_graph` iterates every function with no try/except, so one poisoned
define pair **aborts the entire repository's C call-graph build** (seen on vendored
LLVM in ziglang/zig). A self-alias was already safe via the `resolved_name !=
call_name` short-circuit; only 2+ node cycles triggered it.

## Tests
`tests/test_c_macro_alias_cycle.py` (new): 2-node cycle, 3-node cycle, self-alias.
- RED (on the #127 base): `RecursionError`.
- GREEN: 3 passed; C call-graph suite 15 passed / 2 skipped.

## Upstream coordination
The cyclic-macro recursion is live on master, on the C staging stack (#114/#121),
and on #127 — all unguarded. This guard is non-overlapping with those refactors;
sequence after #127 (or the C stack), re-anchoring onto the post-rework
`_resolve_call`.

## Author notes
- **Fix line:** the `if resolved_name not in _alias_chain:` guard around the
  recursive `_resolve_call(..., _alias_chain=_alias_chain)`.
- **Input it now handles:** cyclic `#define` macro pairs that crashed the whole-repo
  parse.
- **Likely pushback:** "set vs depth cap?" — a visited-set terminates on the actual
  cycle with no arbitrary limit, mirroring the file's other visited guards.

### DIFF
--- a/libs/openant-core/parsers/c/call_graph_builder.py
+++ b/libs/openant-core/parsers/c/call_graph_builder.py
@@ -110,6 +110,11 @@ class CallGraphBuilder:
         self.macros = extractor_output.get('macros', {})
         self.macro_aliases = extractor_output.get('macro_aliases', {})
         self.prototypes = extractor_output.get('prototypes', {})
+        # class_name -> [direct base-class name, ...] for the inheritance walk in
+        # member dispatch (bug [30]). Defaults to {} when the extractor output
+        # predates base-class extraction, so resolution degrades to the [51]
+        # same-type behavior rather than erroring.
+        self.class_bases: Dict[str, List[str]] = extractor_output.get('class_bases', {})
         self.repo_path = extractor_output.get('repository', '')
 
         self.max_depth = options.get('max_depth', 3)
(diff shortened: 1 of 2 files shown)


gadievron:
Finding **F4** (HIGH). ⚠️ **Stacked on #127** (base `pr/zig-php-generic-anon`). The cyclic-macro `RecursionError` it fixes is live on `master`, the C staging stack (#114/#121), AND #127's C rework — all carry the unguarded `_resolve_call` macro-alias recursion. This adds only the `_alias_chain` cycle guard. Merge after #127, or rebase onto master.

gadievron:

Rebased onto the newly-rebased #127 (kept stacked, base pr/zig-php-generic-anon). Merge AFTER #127. C tests green (16).

gadievron:

PR #131, opened 2026-06-18
fix(parsers/php): resolve $this->/self:: calls into used traits

gadievron:
# fix(parsers/php): resolve $this->/self:: calls into used traits

**Base:** `pr/zig-php-generic-anon` · **Depends-on:** #127 · **Type:** bug fix · **Finding:** F12 (HIGH)

> Stacked on our own in-flight PR #127 (`pr/zig-php-generic-anon`), which already
> lands the prerequisite PHP-parser fixes (untagged-fragment `<?php` prepend in
> `_extract_calls_from_code`; `relative_scope` handling for `self::`/`static::`/
> `parent::` in `_resolve_scoped_call`). This PR adds ONLY the net-new
> trait-composition delta — no overlap with #127's hunks.

## What
- `parsers/php/function_extractor.py` — capture in-class `use_declaration` trait
  names onto a new `traits` field on the class record (`_extract_trait_names`).
- `parsers/php/call_graph_builder.py` — build a `traits_by_class` index and make
  `_resolve_self_call` fall back to the used traits' methods when no own method
  matches.

## Why
`$this->method()` / `self::method()` calls into a method defined in a `use`d trait
were never resolved: `_resolve_self_call` only indexed methods physically in the
class body, and trait `use` was never recorded as a class→trait composition. Trait
methods live under the trait's own key, so the edges were dropped. Measured: of
trait methods, symfony 77% orphaned (425/553), laravel 68%, composer 71%, guzzle
86% — a major contributor to PHP's high isolated-function ratio.

## Tests
`tests/test_php_trait_self_call.py` (new): trait `T{ g() }`, class
`C{ use T; f(){ $this->g(); } h(){ self::g(); } }`; asserts edges C::f→T::g and
C::h→T::g (builder-dict + end-to-end extractor layers).
- RED (on the #127 base): edges missing (`'a.php:T.g' not in []`).
- GREEN: 2 passed; full PHP suite 46 passed (incl. `test_php_schema_completeness`
  — the new `traits` field doesn't break the schema).

## Scope / siblings
Ruby has the same shape (`include`/`extend`/`prepend` mixins never recorded as a
class→module composition; `_resolve_self_call` has no mixin fallback) — confirmed
sibling, tracked separately, not fixed here.

## Upstream coordination
No open PR adds trait-composition resolution. PR #122 (php/ruby) handles only
conditional-branch same-name collisions, not traits — non-overlapping. Must merge
after #127.

## Author notes
- **Fix lines:** `traits` capture in `function_extractor.py`; `traits_by_class` +
  `_resolve_self_call` trait fallback in `call_graph_builder.py`.
- **Input it now handles:** `$this->g()` / `self::g()` where `g` is defined in a
  `use`d trait.
- **Likely pushback:** "why stacked, not vs master?" — the two prerequisite fixes
  live in #127; rebasing onto master would duplicate/conflict with them.

### DIFF
--- a/libs/openant-core/parsers/php/call_graph_builder.py
+++ b/libs/openant-core/parsers/php/call_graph_builder.py
@@ -134,6 +134,8 @@ class CallGraphBuilder:
         self.functions_by_name: Dict[str, List[str]] = {}
         self.functions_by_file: Dict[str, List[str]] = {}
         self.methods_by_class: Dict[str, List[str]] = {}
+        # class_key -> list of trait names the class composes via in-class `use`.
+        self.traits_by_class: Dict[str, List[str]] = {}
 
         self._build_indexes()
 
@@ -161,6 +163,13 @@ class CallGraphBuilder:
                 if class_key not in self.methods_by_class:
                     self.methods_by_class[class_key] = []
                 self.methods_by_class[class_key].append(func_id)
+
+        # Index each class's composed traits (in-class `use TraitName;`) so a
+        # $this->/self:: call can fall back to a method pulled in from a trait.
+        for class_key, class_data in self.classes.items():
+            traits = class_data.get('traits')
+            if traits:
+                self.traits_by_class[class_key] = list(traits)
 
     def _is_builtin(self, name: str) -> bool:
         """Check if name is a PHP builtin or common function."""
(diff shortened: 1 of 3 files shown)


gadievron:
Finding **F12** (HIGH). ⚠️ **Stacked on #127** (base `pr/zig-php-generic-anon`). #127 already lands the prerequisite PHP fixes (`<?php` tagless-parse, `relative_scope`); this PR adds ONLY the net-new class→trait composition index + `_resolve_self_call` trait fallback (symfony had 77% of trait methods orphaned). No overlap with #127's hunks. Merge after #127.

gadievron:

Rebased onto the newly-rebased #127 (kept stacked, base pr/zig-php-generic-anon). Merge AFTER #127. PHP tests green (22); #127 anon-class handling intact.

gadievron:

PR #134, opened 2026-07-05
feature/release 2026 29 06

ar7casper:
- **fix(parsers/python): match imports by module components in both directions**
- **feat(context): gate local-only suppression on trust_boundaries, not just app type**
- **fix(cli): add zig to the --language choices for parse and scan**
- **fix(go-parser): key generic-type methods on base type, not "unknown"**
- **build(parsers/go): stop tracking the compiled go_parser binary**
- **fix(reachability): anchor user-input entry-point patterns with word boundary**
- **fix(parsers): CM-A is_test_file anchored to path components, not substring (BUG-NEW 2,22)**
- **fix(parsers/zig): u13 call_graph_builder — local-type/builtin/const-alias (BUG-NEW 3,17,41)**
- **fix(parsers/zig): u14 function_extractor — fn-name/test-classification/container-methods (BUG-NEW 16,26,37)**
- **fix(parsers/c): u1 call_graph_builder — builtin-leak same-file pre-check (BUG-NEW 13)**
- **fix(parsers/c): u2 function_extractor — ctor/dtor/struct/union/lambda/operator/template/ns (BUG-NEW 14,15,32,33,35,39,40)**
- **fix(parsers/c): u3 unit_generator — carry is_inline into the unit (BUG-NEW 29)**
- **fix(parsers/php): u5 call_graph_builder — variable-fn + parent::/self::/static:: + <?php-tag root cause (BUG-NEW 20,52)**
- **fix(parsers/c): land deferred [51] member-dispatch + [30] virtual/inherited dispatch**
- **fix(parsers/ruby): u10 call_graph_builder — parenless/crossfile-class/builtin/method-object/substring-import (BUG-NEW 1,8,11,31,49)**
- **fix(parsers/php): u6 function_extractor — nested-fn/attribute-start_line/enum/braceless-ns (BUG-NEW 7,25,36,54)**
- **fix(parsers/php): u7 unit_generator — namespace key-drift + is_static (BUG-NEW 43,28)**
- **fix(parsers/ruby): u11 function_extractor — nested/singleton/define_method/alias/private/nested-class (BUG-NEW 6,18,24,42,44,50)**
- **fix(parsers/ruby): guard alias resolution to single unconditional binding (BUG-NEW 31)**
- **fix(reachability): u12 entry_point_detector — seed silent binary main (BUG-NEW 4)**
- **fix(parsers/go): u4 BUG-5 — normalize call_graph.json func records to snake_case (BUG-NEW 5)**
- **fix(parsers/go): [5] re-verify — complete the snake-case normalization schema**
- **fix(parsers/go): u4 callgraph — func-value alias edge + top-level call routing (BUG-NEW 19)**
- **feat(reachability): opt-in library-mode seeds the public API surface**
- **fix(parsers/python): u9 function_extractor — 6 finder bugs (BUG-NEW 9,21,23,34,45,48)**
- **fix(parsers/python): u8 call_graph_builder — builtin-leak/alias/import-overresolve/inherited-self (BUG-NEW 10,12,27,46)**
- **fix(parsers/python): [34] re-verify — token-aware property classification (cached_property)**
- **fix(parsers/python): guard alias resolution to single unconditional binding (BUG-NEW 12)**
- **fix(parsers/python): resolve self/cls-aliased method calls in the call graph**
- **fix(parsers/go): classify entry points by signature, not return type or bare next(**
- **fix(parsers/js): extract module-level block-scoped functions**
- **fix(parsers/c): resolve static-inline functions in included headers**
- **fix(parsers/php): keep both same-name defs from conditional/guarded branches**
- **feat(reachability): warn on silent library blackout across all parsers**
- **fix(parsers/c): stop same-(file,name) functions collapsing to one func_id**
- **fix(parsers/python): extract block-scoped def/class units**
- **fix(parsers/ruby): keep both same-name defs from conditional branches**
- **feat(reachability): extend opt-in library-mode to all parsers + CLI**
- **fix(zig,php parsers): attribute generic-container & anonymous-class methods correctly**
- **docs(zig,php parsers): surface grammar-alignment prerequisite in regression tests**
- **fix(parsers/php): resolve $this->/self:: calls into used traits**
- **fix(parsers/c): guard macro-alias resolution against cyclic #defines**
- **fix(parser_adapter): restore --fresh dataset deletion silently dropped by the #117 merge**


gadievron:
### reachability: `name:main` can silently drop analyzed units for library-shaped targets (default mode)

The new `name:main` seed changes default reachability output vs `master` for one specific target shape: a library whose vulnerable surface is not reachable from a function/method named `main`, where that `main` has no user-input pattern. In that case PR-default drops those units from the analyzed set — and unlike `master`, prints no warning.

**Minimal repro** (`--level reachable`, no `--library-mode`):

```
# lib.py
def process(untrusted):
    return _sink(untrusted)
def _sink(data):
    return eval(data)          # reachable only via process()
# cli.py
def main():
    return _greet()
def _greet():
    return "hi"
```

| | units kept |
|---|---|
| `master` | `cli.py:main`, `cli.py:_greet`, `lib.py:process`, `lib.py:_sink` (4) |
| PR-134 default | `cli.py:main`, `cli.py:_greet` (2) — `lib.py:process`/`_sink` dropped |
| PR-134 `--library-mode` | all 4 |

On this input the structural detector returns **0** entry points on `master` but **1** on PR (`cli.py:main`, reason `name:main`). The analyze stage consumes only `dataset["units"]` (`core/analyzer.py:419`), so the dropped `eval` unit is never sent to the model.

**Mechanism** — the filter logic is identical in both trees; only the seed set differs:

- `utilities/agentic_enhancer/entry_point_detector.py:221-222` seeds any function whose `name == 'main'`, ungated by `library_mode`.
- `core/parser_adapter.py:309` keeps all units only when the seed set is empty (`if not entry_points and original_count > 0: return dataset`). `master` hits this net (0 seeds) and keeps everything; the `name:main` seed makes the set non-empty, so PR bypasses the net and BFS prunes everything not reachable from `main`.

**Why it's silent** — `blackout_warning` (`entry_point_detector.py:322`) fires only on total blackout (`reachable_count == 0`, but `main` survives) or partial blackout (`reduction >= 0.90 AND structural == 0`). `"name"` is in `_STRUCTURAL_REASON_CATEGORIES` (line 319), so a `name:main` seed makes `structural >= 1`, suppressing the partial-blackout branch at any reduction ratio. In a 23-unit variant (2 `cli.py` units + a 21-function library chain, one incidental `main`) PR prints `Units: 23 -> 2 (91.3% reduction)` with no warning key; `master` keeps 23/23 and prints its "No entry points detected … returning all units unfiltered" warning.

**Scope** — this does *not* affect conventional CLIs: a `main` containing `argparse`/`sys.argv`/`click`/`input(` is already seeded via `input_pattern` on `master` too, so both trees filter identically (verified). The divergence is limited to input-pattern-free `main`s — e.g. a class method `def main(self)`, a `main()` delegating to a runner/group object defined elsewhere, or a dual-use console-script package (importable API + `[project.scripts] main`). `name == 'main'` is unqualified, so it also matches methods at any nesting. `--library-mode` recovers the dropped units but is opt-in and — because the warning is suppressed — the user gets no signal to reach for it.

**Suggested fix** (any one):
- Drop `"name"` from `_STRUCTURAL_REASON_CATEGORIES` so the partial-blackout warning fires and points to `--library-mode`; or
- When the only seeds are `name:main`, fall back to keep-all; or
- Gate `name:main` behind `library_mode`.

Also, the comment at `entry_point_detector.py:216-220` calls over-approximating `main` "reachability-safe"; that holds when it *adds* a seed to an already-seeded set, but when `name:main` is the *sole* seed it converts keep-all into keep-main-subtree, which removes units.

Non-blocking given the narrow trigger and the `--library-mode` recovery, but worth a signpost or gate.

PR #162, opened 2026-07-16
Update README.md: fix path

wurstbrot:
fix path


gadievron:
Thanks — but master's README already has the correct path. The binary builds to apps/openant-cli/bin/openant (Makefile BUILD_DIR=bin runs inside apps/openant-cli/), so $(pwd)/apps/openant-cli/bin/openant from the repo root is right; $(pwd)/bin/openant points at a nonexistent top-level bin/. Closing as not-needed.

PR #203, opened 2026-08-01
feat: AWS Bedrock provider adapter

NahumKorda:
Adds `bedrock` as a first-class provider type — Claude on AWS Bedrock — per the roadmap's "More provider adapters" item; related to #9 and the pricing-accounting invariants of #65.

## Design

The registry constructs every adapter as `adapter_cls(api_key=..., base_url=...)`, and `ProviderConfig` deliberately has no other fields. Rather than widening that surface for one provider, `BedrockAdapter` keeps the two-kwarg constructor and delegates credentials and region entirely to the AWS SDK chain (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars, `~/.aws` profiles, `AWS_REGION`) via `anthropic.AnthropicBedrock` — so `openant` authenticates exactly like the `aws` CLI on the same machine, and the diff stays Python-only (no Go or config-schema changes). `api_key` is accepted-and-ignored with a one-time warning, since Bedrock has no API-key auth in the pinned SDK.

The wire format is Anthropic's, so the adapter reuses the Anthropic adapter's translation helpers; `anthropic.py`'s shared warnings take an `adapter=` name so reused code attributes messages correctly.

## Bedrock-specific behavior

- Model IDs are inference profiles (`us.anthropic.claude-sonnet-4-6`, `global.anthropic.claude-haiku-4-5-20251001-v1:0`, ...), passed through verbatim; IDs and pricing records in `config/models.json` were verified against a live Bedrock account via `list-inference-profiles` (us-east-1), not derived from naming conventions — the `-v1:0` suffix is not uniform across profiles.
- AccessDenied 403 → `LLMAuthError` with a "Model access" hint (the most common first-run failure: the model isn't enabled in the Bedrock console).
- 400 ValidationException "model identifier is invalid" → `LLMNotFoundError` (Bedrock does not 404 unknown models), so the registry's init-time `validate()` reports a typo'd profile as such.
- The SDK's bare `RuntimeError("could not resolve credentials...")` from the SigV4 signer → `LLMAuthError` with a pointer at the env vars; unrelated RuntimeErrors still propagate.
- Throttling 429s report to the process-global `RateLimiter`, same as the reference adapters.

## Testing

- Contract harness row + 15 Bedrock-specific tests, all SDK-boundary-mocked. Full core suite passes (2 pre-existing Zig parser failures unrelated to this change, present on clean master).
[...]

PR #204, opened 2026-08-01
feat: OpenRouter provider adapter

NahumKorda:
Adds `openrouter` as a first-class provider type, per the roadmap's "More provider adapters" item; related to #9 and the pricing-accounting invariants of #65. Complements #203 (Bedrock) — the two are independent; whichever merges second has a trivial conflict in `providers/__init__.py`, `model_registry.py`, `models.json`, and the README table.

## Why a dedicated adapter (vs. `type: "openai"` + `base_url`)

The README already suggests OpenRouter can be reached through the OpenAI adapter with a `base_url` override, and a live gap-analysis confirmed that mostly works — text rounds and full tool round-trips pass. A dedicated type closes what doesn't:

- **Unknown models were untyped.** OpenRouter reports an unknown model as 400 `"... is not a valid model ID"`, not 404 (live-verified), so the generic mapping buried it in `LLMResponseError`; it now maps to `LLMNotFoundError` and the registry's init-time `validate()` reports it as the typo it is.
- **Cost accounting read $0.** OpenRouter's `vendor/model` slugs (`anthropic/claude-sonnet-4.6`, ...) had no `config/models.json` records. This PR adds 15, covering the current Claude/GPT/Gemini models, with rates pulled from the live OpenRouter catalogue (`GET /api/v1/models`, 2026-08-01).
- **Key handling.** `api_key` falls back to `OPENROUTER_API_KEY` and fails loud when unresolvable — never falling through to the SDK's `OPENAI_API_KEY` default, which would quietly send an OpenAI key to a third party.
- **OpenRouter-only error semantics.** 402 credits-exhausted → `LLMAuthError` with a top-up hint; 403 is OpenRouter's *moderation* signal → `LLMRefusalError` (a moderated scan prompt must not read as an auth failure — or as a clean pass); a provider failing mid-generation comes back as HTTP 200 with `finish_reason == "error"` → raised as `LLMResponseError` instead of warn-and-normalise.
- **Attribution headers** (`HTTP-Referer` / `X-Title`), per OpenRouter's app-attribution docs.

The wire format is OpenAI's, so the adapter reuses the OpenAI adapter's translation helpers (`openai.py`'s shared warnings take an `adapter=` name so reused code attributes messages correctly). `_token_param` already strips vendor prefixes, so `openai/o3`-style slugs get `max_completion_tokens` for free.

## Testing

- Contract harness row (reusing the OpenAI scenario scripts — same SDK boundary) + 16 OpenRouter-specific tests, all mocked. Full core suite passes (2 pre-existing Zig parser failures unrelated to this change, present on clean master).
- Live-validated against a real OpenRouter account: `validate()`, a text round, a full `echo`-tool round-trip on `anthropic/claude-sonnet-4.6`, and the unknown-model → `LLMNotFoundError` mapping.

## Notes for reviewers

- Go wizard support (`openant setup llm`) is intentionally out of scope, matching #203; the README documents hand-authored config.
- Gemini 2.0/1.5 records were deliberately not mirrored — OpenRouter no longer lists them in its catalogue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,9 @@ Wizard defaults reflect the project's per-phase recommendations (stronger reason
 | `anthropic` | [console.anthropic.com](https://console.anthropic.com/settings/keys) | Reference adapter. NOT included in Claude Pro / Max subscriptions — separate billing. |
 | `openai` | [platform.openai.com](https://platform.openai.com/api-keys) | NOT included in ChatGPT / Codex subscriptions — separate billing. |
 | `google` | [aistudio.google.com](https://aistudio.google.com/apikey) | NOT included in Gemini Advanced — separate billing. |
+| `openrouter` | [openrouter.ai](https://openrouter.ai/settings/keys) | Gateway to many providers with one key and one prepaid balance (also reads `OPENROUTER_API_KEY`). Model IDs are `vendor/model` slugs (`anthropic/claude-sonnet-4.6`, `openai/gpt-4o-mini`, ...) — browse them at [openrouter.ai/models](https://openrouter.ai/models). Not offered by the setup wizard yet; configure by hand (see below). |
 
-All three support tool calling, so any of them can drive the `enhance` and `verify` phases that use the agentic tool-use loop.
+All four support tool calling, so any of them can drive the `enhance` and `verify` phases that use the agentic tool-use loop.
 
 #### Quick path for Anthropic-only setups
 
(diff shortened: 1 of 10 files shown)


NahumKorda:
Added a detailed usage guide in 33cc127: [`utilities/llm/providers/OPENROUTER.md`](https://github.com/NahumKorda/OpenAnt/blob/feat/openrouter-adapter/libs/openant-core/utilities/llm/providers/OPENROUTER.md) — prerequisites, key resolution, a full config.json example, model-slug conventions, cost accounting, and a troubleshooting table. Placed next to the adapter code rather than under `docs/`, since `docs/` is gitignored.

gadievron:
Thanks so much for this, @NahumKorda — the OpenRouter adapter has **landed on master via #226**. Same story: master is squash-only, so this needed a rebase onto current master (the #206/#208 empty-content + Responses-path hardening had landed) plus follow-ups (setup-wizard wiring with an OpenRouter-specific probe endpoint, a sharper 403 moderation-vs-auth heuristic). Your commits were carried with **authorship preserved**, and the squash-merge carries a `Co-authored-by: NahumKorda` trailer so it's credited on your graph. Closing in favor of #226 — the credit and the code are yours. Thanks for the contribution.

PR #346, opened 2026-08-27
feat: first-class Ollama provider adapter — fully-local SAST

Roughn3ck:
## What

Adds a first-class **Ollama provider adapter** so OpenAnt's LLM-assisted scans run entirely against local models — no cloud, no API keys, no source code leaving the machine.

## Why

SAST is a natural LLM workload, but shipping source code to a third-party API is a non-starter for air-gapped environments, regulated data, and client-confidentiality constraints. Ollama is the de-facto local runtime, and it exposes an OpenAI-compatible endpoint — so this slots into the existing provider architecture cleanly.

## Changes

- **Provider adapter** (`libs/openant-core/utilities/llm/providers/ollama.py`): thin delegation over the existing OpenAI-compatible helpers. API key optional — stock local installs probe with a placeholder automatically; non-blank keys are forwarded verbatim so key-checking remote gateways still work.
- **Registry** (`providers/__init__.py`): `get_adapter_class` + supported-list wiring.
- **Setup wizard** (`apps/openant-cli`): Ollama offered as a first-class provider with connectivity probing, hint text, and tier-model defaults (`qwen3.8:27b` — current gen, 18GB/256K ctx). The Go probe rewrites the generic 404 into a fixable `ollama pull` hint, and blank-key phases are no longer skipped for Ollama (it needs no key, so its probe always runs).
- **Model registry** (`config/models.json`): `qwen3.8:27b`, `qwen3.5:9b` (modest-hardware alternative), `mistral-small3.2:latest` — all $0 local entries with provenance notes.
- **Tests**: contract-test factory + dedicated adapter suite, consistent with how every other provider is tested. **79 passed, 6 skipped** (adapter + contract suites). `go build`/`vet`/`test` clean.
- **Docs**: `OLLAMA.md` — install, model selection, warm-up behavior, and the base-vs-chat model gotcha.

## Verified end-to-end

Scanned a repo with planted vulnerabilities using `mistral-small3.2:latest` (15GB, local Ollama on a developer workstation):
- **Both planted SQL injection findings (CWE-89) detected correctly**
- Zero network egress beyond the machine

## Practical notes from real usage

- **Use chat/instruct models** — base completion models reject `/chat/completions`.
- **Cold loads**: 15GB+ models take minutes on first inference; the wizard probe can time out on a cold start (documented in OLLAMA.md — possible follow-up: bump probe timeout for local providers).
- **Size to hardware**: qwen3.8 ships 27b only on Ollama; modest-hardware users can pick a smaller model per-phase.

Happy to split or restructure anything for review — the adapter follows the existing provider recipe (`HOW_TO_ADD_AN_ADAPTER.md`).

*Also live as a community fork with docs: https://github.com/Roughn3ck/OpenAnt — happy to walk through the end-to-end results.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,9 @@ Wizard defaults reflect the project's per-phase recommendations (stronger reason
 | `google` | [aistudio.google.com](https://aistudio.google.com/apikey) | NOT included in Gemini Advanced — separate billing. |
 | `bedrock` | — (AWS credential chain) | Claude on AWS Bedrock. No `api_key`: credentials come from `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars or a `~/.aws` profile, region from `AWS_REGION`. Model IDs are inference profiles (`us.anthropic.claude-sonnet-4-6`, `global.anthropic.claude-haiku-4-5-20251001-v1:0`, ...) — enable them under "Model access" in the Bedrock console and list them with `aws bedrock list-inference-profiles`. Offered by `openant setup llm` (leave the API key blank — AWS credential chain, probe skipped) — full guide: [`utilities/llm/providers/BEDROCK.md`](libs/openant-core/utilities/llm/providers/BEDROCK.md). |
 | `openrouter` | [openrouter.ai](https://openrouter.ai/settings/keys) | Gateway to many providers with one key and one prepaid balance (also reads `OPENROUTER_API_KEY`). Model IDs are `vendor/model` slugs (`anthropic/claude-sonnet-4.6`, `openai/gpt-4o-mini`, ...) — browse them at [openrouter.ai/models](https://openrouter.ai/models). Offered by `openant setup llm` (leave the base URL blank for the OpenRouter default) — full guide: [`utilities/llm/providers/OPENROUTER.md`](libs/openant-core/utilities/llm/providers/OPENROUTER.md). |
+| `ollama` | — (local server) | Local models via [Ollama](https://ollama.com). No `api_key`: leave it blank (a placeholder is sent automatically); base URL defaults to `http://localhost:11434/v1`. Models must be pulled first (`ollama pull <model>`); model IDs are exactly what `ollama list` shows. Local inference is free — $0 cost reporting. Offered by `openant setup llm` — full guide: [`utilities/llm/providers/OLLAMA.md`](libs/openant-core/utilities/llm/providers/OLLAMA.md). |
 
-All four support tool calling, so any of them can drive the `enhance` and `verify` phases that use the agentic tool-use loop.
+All five support tool calling, so any of them can drive the `enhance` and `verify` phases that use the agentic tool-use loop. For Ollama, pick a tools-capable model for those phases — very small local models may not handle tool calls reliably.
 
 #### Quick path for Anthropic-only setups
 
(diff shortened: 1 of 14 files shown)


gadievron (changes requested):
Deep maintainer review (two independent adversarial passes, then a re-run adjudicating the design fork — all findings re-derived from source against both the branch tree and current master).

**Overall: this is a strong, recipe-conforming adapter — the delegation design, the no-env-redential rule (airtight; unlike the OpenAI adapter it never lets the SDK read another provider's env var), the redaction paths, and the error taxonomy are all correct and well-documented. Requesting changes for one blocker cluster and a few should-fixes.**

## Blockers (the PR is internally split)

1. **The pricing work is dead code.** `OllamaAdapter` never sets `pricing = LazyProviderPricing("ollama")` — every sibling adapter does (anthropic.py:128, openai.py:320, google.py:179, bedrock.py:124, openrouter.py per recipe). Without it, `lookup_pricing` returns None and every Ollama run emits the unknown-model warning and reports `cost_incomplete: true` — the exact outcome your own new invariant text says is wrong for a known, deliberately-free provider. Your commit b41d0f1 (models.json explicit-$0 entries + the `_LOCAL_PROVIDERS` exemption) describes design (a) but the runtime takes commit a42e264's design ("no pricing table"). Wire the pricing attribute and commit 2 goes live: $0 with `cost_incomplete=false`, no false warning.
2. **Docs must match the design you pick** (with (a): OLLAMA.md's "no pricing table shipped" section and the ollama.py docstring lines 21-25 need updating to "explicit $0 entries shipped"; note non-registry local models still take the warn path).
3. **The missing test that let the split ship**: no test covers the cost-report path for Ollama — assert `OllamaAdapter().pricing` resolves the $0 entries and a completion yields `cost_incomplete=False`. Also extend `testProviderTypes` (Go models_test.go) with "ollama", and pin the placeholder-key test to `== "ollama"` (it currently asserts truthiness only).

## Should-fix

4. **The unconditional 404 pull-hint**: a base_url misconfiguration (missing `/v1`) 404s and gets told to `ollama pull` — misleading for a config typo. Gate on the not-pulled markers (Python: the `_is_not_pulled` check is currently dead — the catch-all `NotFoundError` branch after it fires anyway; Go: gate the rewrite on body markers) and give a marker-less 404 a base_url-checking hint.
5. **Cold-load misclassification**: Python `validate()` maps `APITimeoutError` (a subclass of APIConnectionError) to the "start it with `ollama serve`" hint — the server IS running, the model is loading. Re-map the timeout ahead of the connection branch ("model may be loading — a cold 15GB load takes minutes"). Go: the wizard probe's hard 15s timeout fails a cold load as "could not reach" — classify or parameterize it (your OLLAMA.md already documents the behavior).
6. Minor: qualify OLLAMA.md's "never leaves the machine" as default-local-only (base_url is overridable; remote gateways are a supported config), and reconcile the model-name spellings across docs/registry/tests.

## Named follow-up (not this PR): local-model verdict quality

The deeper risk of fully-local SAST is epistemic, not transport: a weak model's verdicts flow through the same report surfaces as frontier-model verdicts with no disclosure — a false-clean report from an underpowered model. This is a model-capability problem the tool already has today via weak cloud models, so it is not a blocker for the adapter — but a doc paragraph on model selection expectations + a tracking issue for a local-model disclosure on the report would close the honesty gap.

Happy to re-review as soon as the blockers land — the architecture is welcome.

gadievron:
Merged via #446 (the completion branch carrying your two commits verbatim — authorship preserved; Co-authored-by on the PR). Thank you for the adapter — the architecture was strong and the review's blockers were all completion-grade (the pricing wiring, the probe classification), not design flaws. The review thread carries the full history.

PR #368, opened 2026-08-27
fix(context): oversized docs are truncated, not skipped; the degradation is visible in the artifact

### LINKED ISSUE #217: 10 KB file-read guard silently degrades application context on real repositories
## Summary

The context phase refuses to read documentation files above 10 KB, then proceeds to generate an application context — and a confidence score — from whatever remains, without reflecting that primary evidence was withheld.

## Evidence

The context phase emits a warning per skipped file and continues:

```
[2/7] Generating application context...
Warning: Could not read <doc>.md: <doc>.md is too large (N bytes > 10000); refusing to read
Warning: Could not read <doc>.md: <doc>.md is too large (N bytes > 10000); refusing to read
```

On our target this fired for both top-level documentation files — one roughly 3x the limit, the other about 1.6x. The scan then continued and emitted `application_context.json` with `"confidence": 0.85`, derived from dependency manifests and framework detection alone.

## Why it matters

The two files that describe the application most directly were both excluded, on a target whose main README is around 3x the limit — not an unusual size for one. The application context feeds the Stage 1 prompt and the Stage 2 attacker personas, so degrading it degrades every downstream stage. The emitted `confidence` gives no hint that inputs were skipped, so a consumer cannot tell a well-grounded context from a partly-guessed one.

The guard in `utilities/file_io.py` is correct to exist — repository content is attacker-controlled — but a 10 KB ceiling is low for documentation specifically.

## Suggested fix

Raise the ceiling for the context stage, or read a bounded prefix (first N KB) instead of refusing outright. Either way, reduce the reported `confidence` when a source was skipped, so the degradation is visible in the artifact rather than only in stderr.

---

*Observed on one full-pipeline run at production scale: a private TypeScript monorepo (Angular front end, NestJS back end), 1,351 analysis units from 866 source files, run with `--verify` and no `--limit`, ~6.5 hours wall clock. The engine was invoked directly (`python -m openant scan …`, Python 3.13); the Go CLI was not built, so none of this involves the Go↔Python envelope. Stage 1 detection and Stage 2 verification ran on **Claude Opus 5**; application context, enhancement and reporting on **Claude Sonnet 5** — all seven phases routed through **OpenRouter** (`anthropic/claude-opus-5`, `anthropic/claude-sonnet-5`) via a locally-added `openrouter` provider adapter, on a build based on upstream `2ed78f6`. Cost figures are expressed as proportions of the run total; absolute amounts are omitted deliberately, as is the identity of the scanned repository.*

gadievron:
## Summary

The context phase called `read_repo_file(max_bytes=10_000)` **without** `oversize="truncate"` — the default `"raise"` made every documentation file above 10 KB raise, get caught, and be **skipped**, while the truncation handler three lines down (append `[... truncated ...]`) was dead code the author clearly intended. On #217's reporter target **both** top-level docs (README ~3× the limit, CLAUDE.md ~3.1×) were dropped and the context was generated from dependency manifests alone, with a high confidence score and no artifact-level trace of the withheld primary evidence.

- Pass `oversize="truncate"` (the codebase's own documented semantics for exploration reads). The guard itself is unchanged — repo content is attacker-controlled; no new configuration; the 10,000 ceiling stays hard-coded.
- Unreadable docs are listed in a `[skipped_sources]` entry inside `sources`; truncated docs in `[truncated_sources]` — **string-valued** (the sources dict is `dict[str, str]`; a list-valued entry crashed both prompt builders the moment the bookkeeping first fired — a 4-seat wave catch, fixed pre-ship with a prompt-builder integration test). The generator prices the gap into its confidence via its existing instruction; no post-hoc arithmetic on the model's number.
- Truncation is detected on the **char basis** (read one past the cap: `len > 10_000` is the only precise signal) — a byte-based check false-positives on multi-byte files whose whole char count fits; a bare `>=` misfires at the exact boundary. Both edges test-locked (exact-fit, multi-byte, one-over).

## Test plan

8 hermetic tests: truncate-not-skip (prefix pinned to exactly 10,000 chars + marker); truncated/skipped listed as strings; small docs unaffected; exact-10,000 not marked; multi-byte file under the char cap complete; one-char-over truncated; the prompt-builder integration test driving the real fence loop over both new entries.

<details>

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue217_doc_oversize.py -q` | 8 passed (offline) |
| hermeticity oracle | `env -i HOME=/tmp/fakehome-noconfig python -m pytest tests/test_issue217_doc_oversize.py` | 8 passed |
| mutation smoke | drop `oversize="truncate"` | 2 tests fail (restored → green) |
| full suite | `pytest tests/ -q` | 3003 passed, 2 failed (pre-existing zig local-env, stash-replay verified; CI green with them), 32 skipped |

Adversarial loop: 4-seat wave → list-valued-entry crash BLOCKER (4-seat convergence) + boundary off-by-one, fixed; confirm round → byte-vs-char false-positive edge + weak length assertion, fixed — truncation now on the precise char basis.

</details>

Fixes #217

### DIFF
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -303,6 +303,8 @@ def gather_context_sources(repo_path: Path) -> dict[str, str]:
         Dictionary mapping filename to content.
     """
     sources = {}
+    skipped: list[str] = []
+    truncated: list[str] = []
 
     # Read priority files
     for filename in CONTEXT_FILES:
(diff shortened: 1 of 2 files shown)


gadievron:
Diagnosis of the windows-latest failure (from E2E validation of this batch): `test_multibyte_file_under_char_cap_not_marked_truncated` writes its fixture with `(tmp_path / "README.md").write_text("日" * 5_000)` — **no `encoding=` argument**. On windows-latest the locale default is cp1252, which cannot encode U+65E5, so the fixture write itself raises UnicodeEncodeError before the assertion is reached (ubuntu/macos default to UTF-8, hence green there). Platform-agnostic fix: add `encoding="utf-8"` to this and the sibling fixtures' write_text calls — the product code under test needs no change. Secondary (product-side, non-blocking, maintainer's call): `read_repo_file(max_bytes=N)` on a text handle effectively bounds **characters**, not bytes — worst case ~4× max_bytes resident (a 10K-char cap admits up to ~40KB for 4-byte chars). The 10 KB semantic is fine; the parameter name/comment overstate the byte guarantee — consider `max_chars` or documenting the 4× ceiling.

Issue #296, opened 2026-08-20
`_has_conclusive_exploit_path` tests a free-text explanation string, so an incomplete verification suppresses a correction — its mirror guard must NOT be tightened (PR #195/#243)


gadievron:
## Summary

`_has_conclusive_exploit_path` and `_has_conclusive_exploitable_path` decide whether a verification's
exploit-path analysis may be trusted. Both test `verification["explanation"] == "Max iterations
reached"` — one of five distinct markers the verifier writes for an incomplete verification — and
then fall back to checking that `exploit_path` is a dict. A verification that is marked
`incomplete: True` but carries a populated `exploit_path` and a model-authored explanation passes
the string test and is treated as conclusive.

The structured signal these guards want (`verification["incomplete"]`) is already present on the same
object.

> **Corrected 2026-08-21 — scope and direction.** As first filed this issue treated the two helpers
> as symmetric and asked for the same change to both. **They are not symmetric**, and the fix below
> has been rewritten. `_has_conclusive_exploit_path` (caller `:887`) returning True **suppresses a
> correction** — that is the real defect. `_has_conclusive_exploitable_path` (caller `:904`)
> returning True **blocks a downgrade** into `DISCLOSURE_DROPPED`; it is a security protection added
> by PR #195 and widened by PR #243, and making it stricter would re-open the silent
> `vulnerable → safe/inconclusive/rejected` false negative those PRs closed. **Only the first helper
> should change.**

## Evidence

`libs/openant-core/utilities/finding_verifier.py:947-959` (excerpt — 4 interior lines elided, 4 of them comments) (HEAD `b501962`):

```python
        verification = result.get("verification", {})

        # If max iterations was reached, the analysis is not conclusive
        if verification.get("explanation") == "Max iterations reached":
            return False

        exploit_path = verification.get("exploit_path")
        if not isinstance(exploit_path, dict):
            return False
```

`:980-985` mirrors it. The string is exact-matched, and it is one of several the file writes.

### An incomplete verification can carry a populated `exploit_path`

`finding_verifier.py:1092-1125` (excerpt — 18 interior lines elided, 14 of them comments) (excerpted — the `VerificationResult(...)` construction, with intermediate lines elided) (`_parse_finish_result`) sets `incomplete` and `exploit_path`
independently, and returns both on the same object:

```python
        agree_missing = "agree" not in finish_result
        ...
        incomplete = agree_missing
        ...
[...]

PR #390, opened 2026-08-28
fix(verifier): name the conclusive-guard marker; pin the protective asymmetry (#296 lower-risk items)

gadievron:
## Summary

Both conclusive-exploit-path guards exact-match **one free-text string** ("Max iterations reached" — one of five distinct incomplete markers the verifier writes) while the structured signal (`verification["incomplete"]`) sits unused on the same object. The issue's corrected analysis: the two helpers guard in **opposite directions** — `_has_conclusive_exploit_path` returning True **suppresses a correction** (the defect), while `_has_conclusive_exploitable_path` returning True **blocks a downgrade** into `DISCLOSURE_DROPPED` (the PR #195/#243 protection) — and any tightening of either re-opens the silent `vulnerable → safe/inconclusive/rejected` false negative.

The direction-aware behavior change is raised in the issue **as an open question, not a proposal** — and the traced consequence of the obvious fix (test `incomplete` in the first helper) is that an incomplete+broken-path unit stops being skipped, nothing blocks the update, and a downgrade to `safe` removes the record from disclosure. This PR therefore ships **only the explicitly-safe lower-risk items**:

- **`INCOMPLETE_VERIFICATION_MARKER`** names the matched marker as a constant with the match set **unchanged**; both guards carry the do-not-widen warning with the traced consequence (the issue's item 1, phase 1);
- **the asymmetry is pinned** (item 3): an incomplete verification carrying an **intact** exploit path still blocks the `DISCLOSURE_DROPPED` downgrade — at the predicate level AND at the `_check_consistency` call site on the PR #243 fixture shape — so any future signal-aware rework cannot silently lose the protection;
- **the open defect is characterized** (incomplete + broken path reads conclusive today and suppresses corrections) with the direction-aware constraint in the test docstring — changing it forces a conscious decision, not a drift;
- **item 4 answered**: there is no Stage-1 mirror guard to keep in sync — `utilities/stage1_consistency.py` (the PR #245 path) carries no exploit-path/conclusive guard at all; the direction-aware fix, if adopted, should consider both consistency paths together.

Behavior is preserved exactly: the pre-existing guard tests (`test_verifier_consistency_no_exploitable_downgrade`, `test_exploit_path_nondict`, `test_verifier_consistency_disclosure_dropped_downgrade`) pass unchanged.

**Refs #296** — deliberately not "Fixes": the behavioral question (the direction-aware guard change) remains open as the issue itself frames it; this PR is the safety net that makes answering it safe.

## Test plan

5 new hermetic tests: the constant is exactly today's match set (widening = the traced FN direction, requires the direction-aware fix first); the marker still disqualifies both guards; the asymmetry — incomplete + intact path blocks the downgrade (predicate + e2e call site on the #243 fixture shape); the open defect characterized with its constraint.

<details>

| check | command | result |
|---|---|---|
| new + existing guard tests | `pytest tests/test_issue296_conclusive_guards.py tests/test_verifier_consistency_no_exploitable_downgrade.py tests/test_exploit_path_nondict.py tests/test_verifier_consistency_disclosure_dropped_downgrade.py -q` | 17 passed (pre-existing tests unchanged) |
| mutation smoke | guard change stashed → new file | errors on import (the constant IS the change); restored → 5 passed |
| hermeticity oracle | `env -i HOME=/tmp/fakehome-noconfig python -m pytest <file>` | 5 passed |
| full suite | `pytest tests/ -q` | 3167 passed, 2 failed (pre-existing zig local-env, stash-verified), 32 skipped |

</details>

Refs #296

### DIFF
--- a/libs/openant-core/parsers/c/PARSER_PIPELINE.md
+++ b/libs/openant-core/parsers/c/PARSER_PIPELINE.md
@@ -117,7 +117,16 @@ Generates `dataset.json` and `analyzer_output.json` with identical schema to Pyt
 
 ### Limitations
 
-- Cannot resolve function pointer dispatch (LLM compensates)
+- Function-pointer dispatch through STATIC containers resolves as of #298 (a
+  braced initialiser of function references — ops struct designated
+  initialisers, function-pointer arrays, struct-array command tables — plus a
+  subscript or member call through the container name records edges to every
+  function the initialisers reference; over-seed is the safe direction).
+  REMAINING unmitigated on a default scan: dispatch through pointers assigned
+  at RUNTIME (indirect calls via variables holding function addresses with no
+  static initialiser table), which the opt-in `--llm-reachability` stage
+  (off by default; reads the first 1,500 bytes per unit) only partially
+  compensates for.
 - Cannot resolve complex macro expansions (tree-sitter sees the macro call, not the expansion)
 - Does not track struct field access patterns
 - C++ template instantiation is not tracked
--- a/libs/openant-core/parsers/c/call_graph_builder.py
+++ b/libs/openant-core/parsers/c/call_graph_builder.py
@@ -349,7 +349,18 @@ class CallGraphBuilder:
                 source = f.read()
             parser = self._get_parser_for_file(file_path)
             tree = parser.parse(source)
-            result = self._collect_container_map(tree.root_node, source, file_path)
+            # TOP-LEVEL ONLY (panel finding): _collect_container_map walks the
+            # whole tree, so a container nested inside an unrelated function's
+            # body would leak file-wide (an opaque same-named parameter in
+            # another function would dispatch to THIS function's targets).
+            # Walk the root's children without descending into function_definition.
+            top_level = [c for c in tree.root_node.children
+                         if c.type != 'function_definition']
+            class _ShallowRoot:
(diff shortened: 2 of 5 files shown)


gadievron:
Panel review (sonnet + fable, executed repros) — holding with findings before merge. C-parser half: (1) _file_containers walks the whole file tree, not just top-level — function-local containers leak file-wide (same-named local/param in another function can pick up another function's table targets = fabrication; or erase a genuine file-scope table = FN). One-line fix: skip function_definition bodies in the file-map walk. (2) _dispatch_base_identifier accepts a bare-identifier root, contradicting its docstring — a direct call through a name shadowing a file-scope container fabricates edges to all targets. Fix: require subscript/field root. (3) _initializer_targets resolves any identifier (incl. data + unary operands) via the repo-unique fallback — cross-file data/function name collision fabricates (same residual class as pre-existing _extract_callback_args; docstring overclaims 'non-function identifiers do not resolve'). (4) _file_containers catches only OSError — widen like the sibling path. The verifier half (#296) verified SAFE. Differential on tmux: +134/-0 (deterministic), genuine dispatch-table recoveries — with the caveat that a fraction may be fabrication-class per (1)/(3). #391 is byte-identically contained in this PR — will close as included when this lands.

Issue #299, opened 2026-08-20
Container-literal dispatch loses call edges in every language parser (Go, JS/TS, PHP, Ruby, Rust, Swift, Zig — after Python #295 and C #298)


gadievron:
## Summary

A function referenced only as a value inside a container literal — a map/dict/hash/object literal, an
array/slice/list — receives no call-graph edge in **any** of the nine parsers, and a call made through
a subscript (`table[key]()`) produces no out-edge from the calling function. The dispatcher is kept by
the reachability filter and its targets are pruned.

Python is filed as #295 and C as #298. This issue covers the remaining seven. Each parser is an
independent implementation — there is no shared base class and no cross-imports — so this is seven
separate fixes, not one.

## Evidence

Every language was measured with an executed fixture containing a **direct-call control** in the same
file. The control passed in every case, so the empty results are edge-recording failures rather than
harness artifacts. Two independent agents reproduced the table below separately.

| language | dispatcher out-edges | container-referenced targets | control | edge-construction site |
|---|---|---|---|---|
| Go | 0 | no callers | passed | `parsers/go/go_parser/callgraph.go` (`IndexExpr` handling) |
| JS/TS | 0 | no callers | passed | `parsers/javascript/typescript_analyzer.js` (`ElementAccessExpression` callee) |
| PHP | 0 | no callers | passed | `parsers/php/call_graph_builder.py` (`_resolve_function_call`, no subscript case) |
| Ruby | 0 | no callers | passed | `parsers/ruby/call_graph_builder.py` (`.call` resolves only via a local-var binding map) |
| Rust | 0 | no callers | passed | `parsers/rust/call_graph_builder.py` (`_describe_callee`, no `index_expression` case) |
| Swift | 0 | no callers | passed | `parsers/swift/call_graph_builder.py:677` |
| Zig | 0 | no callers | passed | `parsers/zig/call_graph_builder.py` (callee-type check omits array access) |

Representative fixture shape, adapted per language:

```go
var handlers = map[string]func(){"a": handlerA}
var tbl      = []func(){handlerB}
func dispatch(k string, i int) { handlers[k](); tbl[i]() }
func control()                 { directTarget() }   // control: edge present
```

### Language-specific detail worth noting

**Go — highest severity of this set.** `callgraph.go` is pure `go/ast` with no `go/types` and no
`x/tools/go/callgraph`, so there is no type-level resolution to fall back on. The `IndexExpr` unwrap
[...]

Issue #295, opened 2026-08-20
Python call graph records no edge for a function referenced as a dict value or list element, so dispatch-table targets are pruned with their dispatcher kept


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

The Python call-graph builder creates an edge when a function is *called by name*. A function
referenced as a value in a dict literal or an element of a list — the standard dispatch-table
pattern — produces no edge at all, even when the dispatcher plainly invokes it. The dispatcher is
kept as reachable and its targets are pruned, so live primary-flow code is excluded from analysis
while the entry point that reaches it is retained.

A call made through a subscript (`HANDLERS[mode]()`) also produces no outgoing edge from the calling
function, so the dispatcher's own out-degree is understated as well.

## Evidence

Minimal fixture, three reference shapes:

```python
# app.py
def handler_a(): return 1
def handler_b(): return 2
def handler_c(): return 3

HANDLERS = {'a': handler_a}          # dict literal value
LIST_REF = [handler_b]               # list element
BUILT    = dict(c=handler_c)         # dict() call kwarg

def main():
    mode = "a"
    HANDLERS[mode]()
    LIST_REF[0]()
    BUILT["c"]()
```

Run through the real builder at HEAD `b501962`:

```python
from parsers.python.function_extractor import FunctionExtractor
from parsers.python.call_graph_builder import CallGraphBuilder

ex  = FunctionExtractor("/tmp/dispatch")
out = ex.extract_all()
cg  = CallGraphBuilder(out); cg.build_call_graph()
print("edges out of main:", cg.call_graph["app.py:main"])
for h in ("handler_a", "handler_b", "handler_c"):
    k = next(k for k in cg.call_graph if k.endswith(":" + h))
    print(h, "callers:", cg.reverse_call_graph.get(k, []))
```

```
edges out of main: []
handler_a callers: []
handler_b callers: []
handler_c callers: ['app.py:__module__']
```

Three observations:

1. **`main` has zero outgoing edges** despite containing three call expressions. The callee of
[...]

PR #391, opened 2026-08-28
fix(c-parser): function-pointer dispatch through static containers records call-graph edges

### LINKED ISSUE #298: C call graph records no edge for function-pointer dispatch, and the documented compensation is opt-in and largely unread
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

`parsers/c/call_graph_builder.py` records a call edge when the callee resolves to an identifier. The
three canonical C indirect-dispatch idioms — an ops struct with designated initialisers, an array of
function pointers, and a name/handler command table — produce **no edge at all**, so the dispatching
function has zero out-edges and every target is orphaned.

`parsers/c/PARSER_PIPELINE.md:120` acknowledges the gap and states that the LLM compensates. That
compensation is opt-in (off by default), its output field has no non-test reader, and the stage sees
only the first 1,500 bytes of each unit.

This is the same defect class as #295 (Python dict/list dispatch), but in C indirect dispatch through
function pointers is not a style choice — it is the language's polymorphism mechanism.

## Evidence

Fixture (`src/d.c`), containing a direct-call **control** so a null result is meaningful:

```c
static void h_a(void) {}
static void h_b(void) {}
static void my_read(int x) {}
static void direct_target(void) {}

struct ops { void (*read)(int); };
static struct ops o = { .read = my_read };
static void (*tbl[])(void) = { h_a, h_b };

void use_ops(int x)      { o.read(x); }
void use_tbl(int i)      { tbl[i](); }
void direct_caller(void) { direct_target(); }
```

Run through the real pipeline at HEAD `b501962`:

```python
from parsers.c.function_extractor import FunctionExtractor
from parsers.c.call_graph_builder import CallGraphBuilder
ex = FunctionExtractor("/tmp/cfix"); out = ex.extract_all()
cg = CallGraphBuilder(out); cg.build_call_graph()
```

```
extracted units: ['direct_caller','direct_target','h_a','h_b','my_read','use_ops','use_tbl']

out-edges:
  direct_caller  -> ['src/d.c:direct_target']      <-- CONTROL PASSES
  use_ops        -> []
  use_tbl        -> []

callers:
  my_read        callers=[]
  h_a            callers=[]
  h_b            callers=[]
  direct_target  callers=['src/d.c:direct_caller'] <-- CONTROL PASSES
```

All seven functions were extracted and the control produced an edge in both directions, so the empty
results are edge-recording failures rather than a harness artifact.

A struct-array command table dispatched by `strcmp` loop (`cmds[i].fn()`) behaves the same way.

### Root cause

`_extract_call_name` (`parsers/c/call_graph_builder.py:336-368`) resolves `identifier`,
`field_expression` (by field name only), `qualified_identifier` and `template_function`. A callee
that is a `subscript_expression` (`tbl[i]`, `cmds[i].fn`) falls through to the
`text.isidentifier()` catch-all at `:368` and returns `None`, so no name ever reaches `_resolve_call`.
The `field_expression` branch at `:343-349` declines to bare-name-resolve, and the receiver-typed
member path never fires for a struct or array *literal* receiver because only local-variable
`type_identifier` declarations are tracked.

### One idiom that does work

A function name passed as an argument — `register_handler(my_callback)` — **does** produce a correct
edge, via `_extract_callback_args` (`:240-259`). That helper is scoped to a bare identifier or
`&identifier` appearing as an argument to a call expression; it does not fire for a struct-literal
field or an array-literal element. So the machinery for "a function referenced without being called"
already exists in this file and covers one of the four cases.

### The documented compensation

`parsers/c/PARSER_PIPELINE.md:120`:

```
- Cannot resolve function pointer dispatch (LLM compensates)
```

Checking that compensation:

- **It is opt-in and off by default.** `core/scanner.py:136` — `llm_reachability: bool = False`;
  `openant/cli.py:1443` registers `--llm-reachability` as a flag. A default scan does not run it.
- **Its output is almost entirely unread.** `grep -rn 'llm_reachability_signals' --include='*.py' .`
  outside tests returns only `core/llm_reachability.py` itself (`:32`, `:431`, `:459` — a docstring,
  a comment, and the `setdefault` that writes it). The only effect on the graph is the
  `is_entry_point` promotion at `:464-470`, and only for signals of kind `entry_point` at high
  confidence.
- **It sees a truncated unit.** `DEFAULT_MAX_CODE_BYTES = 1500` (`core/llm_reachability.py:63`), so
  the stage reads the first 1,500 bytes of each unit — and a dispatch table is frequently declared
  well past that point.

`parsers/c/PARSER_UPGRADE_PLAN.md:32` lists "Function pointer tracking" under *Potential
Improvements*, i.e. not implemented, and `:39` names an OpenSSL "provider architecture with function
pointer dispatch tables" as a known pain point.

### Not asserted as intended

`tests/parsers/c/test_call_graph_builder_dispatch.py` is the only C test with "dispatch" in its name
and covers exclusively C++ virtual and inherited method dispatch (`w.compute()`, `b->compute()`). No
test in `tests/parsers/c/` pins the function-pointer behaviour as deliberate.

## Why it matters

In C, an ops struct or function-pointer table is how the language expresses polymorphism: kernel
`file_operations`, driver op tables, syscall tables, plugin vtables, protocol handler tables, state
machines. These are exactly the edges that carry attacker-reachable control flow into a handler.

Because the dispatcher is retained by the reachability filter and its targets have no callers, the
result is the shape described in #295: the entry point is present, so the module looks covered, while
the code it reaches is pruned.

## Suggested fix

1. Extend `_extract_call_name` to resolve a `subscript_expression` callee to the functions its
   container references, or record a reference edge from the enclosing function.
2. Reuse the existing `_extract_callback_args` idea for initialiser contexts: when a function name
   appears as a value in a struct designated initialiser or an array initialiser, record a reference
   edge from the enclosing scope. Over-seeding is the safe direction for a reachability filter.
3. Correct `PARSER_PIPELINE.md:120` — either implement the compensation it claims, or state that the
   gap is unmitigated on a default scan.
4. Add fixtures for the three idioms above (with a direct-call control) to `tests/parsers/c/`.

## What I am not claiming

- I am **not** claiming a measured impact on a real C repository. The run that prompted this work
  targeted Python; the fixture above is synthetic and minimal. The mechanism is demonstrated, its
  magnitude on real C code is not.
- I am **not** claiming the LLM reachability stage never helps — with `--llm-reachability` enabled
  and a unit whose relevant code falls inside the first 1,500 bytes, a high-confidence `entry_point`
  signal does promote the unit. I am claiming that is a narrow path, off by default, and not the
  general compensation the pipeline doc describes.
- I have not checked whether the C++ paths (`qualified_identifier`, `template_function`) have an
  analogous gap for member function pointers.

---

**Umbrella cross-reference (added 2026-08-22, back-link).** #299 is the multi-language umbrella for
this defect class and names this issue in its opening: *"Python is filed as #295 and C as #298. This
issue covers the remaining seven."* It also records the reason the three are tracked separately —
*"Each parser is an independent implementation — there is no shared base class and no cross-imports —
so this is seven separate fixes, not one."*

Recorded here because the reference was one-way: #299 cited this issue and this body cited nothing
back, so a reader arriving here had no route to the umbrella or to the sibling language issues.

gadievron:
## Summary

The three canonical C indirect-dispatch idioms — an ops struct with designated initialisers, an array of function pointers, and a struct-array command table — produced **no edge at all**: `_extract_call_name` resolved identifier/field/qualified/template callees, a `subscript_expression` callee fell through the `isidentifier()` catch-all to `None`, and the member path never fired for struct/array *literal* receivers. The dispatcher kept zero out-edges and every target was orphaned — #295's shape, in C, where function-pointer dispatch is the language's **polymorphism mechanism** (kernel `file_operations`, driver op tables, syscall tables, plugin vtables).

The machinery for "a function referenced without being called" already existed in this file (`_extract_callback_args` covers `register_handler(cb)`); initialisers and container dispatch are the missing members of that family.

Fix (issue suggestions 1+2+3; 4 = the tests):
- **initializer_list reference edges**: a bare identifier (or `&id`) as an initialiser element, a designated-initialiser value, or a nested struct-array element resolves to an edge from the enclosing unit;
- **container dispatch**: a name initialised with a braced list of function references (file-scope, parsed from the file and cached, or function-local) plus a subscript call (`tbl[i]()`), member call over an identifier (`o.read(x)`), or member-over-subscript (`cmds[i].fn()`) records edges to **every** function the initialisers reference — over-seed, the safe direction; nothing invented (unknown bases record nothing, tested);
- single-declaration guard: a name declared twice is dropped rather than guessed;
- **`PARSER_PIPELINE.md`'s compensation line corrected** (suggestion 3): static containers now resolve; runtime-assigned pointer dispatch remains unmitigated on a default scan — the opt-in `--llm-reachability` stage (off by default, 1,500-byte unit window) only partially compensates.

**T3 real-corpus differential** (redis `src/t_hash.c` + `src/rdb.c` and linenoise — 264 units): base 445 edges → fix 447 edges, **LOST: 0** (monotone). The +2 gained are **real dispatch edges in redis code**: `hashTypeDbActiveExpire → hashTypeActiveExpire` and `hashTypeExpire → onFieldExpire` (the field-expiry dispatch).

Sibling: Python is #295 (PR #388); the other seven parsers are #299.

## Test plan

8 hermetic tests in the canonical parser-test location (`tests/parsers/c/`): the issue's fixture with a direct-call control (ops-struct dispatch, function-pointer table, no target orphaned) + the struct-array command table + a function-local table + unknown-subscript and non-function-initialiser negative controls.

<details>

| check | command | result |
|---|---|---|
| new tests | `pytest tests/parsers/c/test_issue298_fnptr_dispatch.py -q` | 8 passed (5 failed RED on pristine) |
| full C parser suite | `pytest tests/parsers/c/ -q` | 107 passed (symmetry included) |
| mutation smoke | builder stashed → new file | 5 failed, 3 passed (negative controls by design); restored → 8 passed |
| T3 real-corpus differential | redis t_hash.c/rdb.c + linenoise, base vs fix | 264 units; 445 → 447 edges; **LOST 0**; +2 real redis dispatch edges |
| full suite | `pytest tests/ -q` | 3170 passed, 2 failed (pre-existing zig local-env, stash-verified), 32 skipped |
| hermeticity oracle | `env -i HOME=/tmp/fakehome-noconfig python -m pytest <file>` | 8 passed |

</details>

Fixes #298

### DIFF
--- a/libs/openant-core/parsers/c/PARSER_PIPELINE.md
+++ b/libs/openant-core/parsers/c/PARSER_PIPELINE.md
@@ -117,7 +117,16 @@ Generates `dataset.json` and `analyzer_output.json` with identical schema to Pyt
 
 ### Limitations
 
-- Cannot resolve function pointer dispatch (LLM compensates)
+- Function-pointer dispatch through STATIC containers resolves as of #298 (a
+  braced initialiser of function references — ops struct designated
+  initialisers, function-pointer arrays, struct-array command tables — plus a
+  subscript or member call through the container name records edges to every
+  function the initialisers reference; over-seed is the safe direction).
+  REMAINING unmitigated on a default scan: dispatch through pointers assigned
+  at RUNTIME (indirect calls via variables holding function addresses with no
+  static initialiser table), which the opt-in `--llm-reachability` stage
+  (off by default; reads the first 1,500 bytes per unit) only partially
+  compensates for.
 - Cannot resolve complex macro expansions (tree-sitter sees the macro call, not the expansion)
 - Does not track struct field access patterns
 - C++ template instantiation is not tracked
--- a/libs/openant-core/parsers/c/call_graph_builder.py
+++ b/libs/openant-core/parsers/c/call_graph_builder.py
@@ -32,6 +32,7 @@ Output (JSON):
 """
 
 import json
+import os
 import re
 import sys
 from pathlib import Path
@@ -212,12 +213,31 @@ class CallGraphBuilder:
         # method on the receiver's known type.
         local_var_types = self._extract_local_var_types(tree.root_node, code_bytes)
 
+        # #298: function-pointer dispatch containers. A name initialised
+        # with a braced list of function references (an ops struct with
+        # designated initialisers, an array of function pointers, a
+        # struct-array command table) plus a subscript or member call
+        # through that name (tbl[i]() / o.read(x) / cmds[i].fn()) is C's
+        # polymorphism idiom; previously it produced NO edge, so the
+        # dispatcher kept zero out-edges and every target was orphaned.
+        # File-scope containers are visible to every function in the file
+        # (parsed from the file itself, cached); local containers override.
+        containers = dict(self._file_containers(file_path))
+        containers.update(self._collect_container_map(tree.root_node, code_bytes, caller_file))
+
         stack = [tree.root_node]
         while stack:
             node = stack.pop()
             if node.type == 'call_expression':
                 func_node = node.child_by_field_name('function')
                 if func_node:
+                    # #298: dispatch through a known container — the
+                    # over-seed direction (a superset of the true targets
+                    # is kept; nothing is invented: only names the
+                    # container's initialisers actually referenced).
+                    for target in self._container_dispatch_targets(
+                            func_node, code_bytes, containers):
+                        calls.add(target)
                     call_name, receiver = self._extract_call_name_and_receiver(
                         func_node, code_bytes
                     )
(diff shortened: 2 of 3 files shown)


gadievron:
This PR's C-parser content is byte-identically contained in #390 (verified: all three file sections match). Recommend closing as included-in-#390 once it lands; keeping both open would double-merge the same hunks.

gadievron:
Closing: the C-parser half of this PR is byte-identically contained in #390 (verified earlier this session), and #390 now carries its review amendments too. Keeping both would double-merge the same hunks.

PR #392, opened 2026-08-29
fix(go-parser): container dispatch edges, the IndexExpr false edge, and method values (#299, 1 of 7)

gadievron:
## Summary

Three defects in the Go call-graph builder — issue #299's Go member, its **highest-severity** row (1 of 7):

1. **FALSE EDGE** (the issue's worst finding): the generic-instantiation `IndexExpr` unwrap collapsed ANY `handlers[k]()` to the bare identifier `handlers`, so a same-named function received the edge while the real dispatch target got nothing (measured in the issue: a shadow function got the edge). Fix: a single-argument `IndexExpr` is unwrapped **only when the index is type-shaped** (uppercase ident, predeclared type name, selector, nested instantiation, or a type literal); a value index records nothing (tracked containers dispatch below; unknown bases abstain). Generic instantiation `fn[T]()` is preserved — proven by the unit fixture **and** the corpus: gin's 34 `getTyped[T]` call edges are identical base-vs-fix.
2. **MISSING EDGES**: a package-scope or local container of function references (`var handlers = map[string]func(){...}` / `var tbl = []func(){h}`) plus a subscript call through it recorded nothing, so dispatch-table targets were pruned with their dispatcher kept. Fix: file-scope containers are collected from a full parse of each file (cached); local containers from the body; a subscript call over a known container edges to **every** referenced function — over-seed, the safe direction; nothing invented (unknown bases abstain, tested).
3. **METHOD VALUES**: `f := v.Method; f()` lost the edge because the alias tracker rejected `SelectorExpr` RHS. Fix: method-value bindings are tracked and resolved against the receiver variable's locally-known type; abstain when the type is unknown.

**T3 differential** (real corpora: gorm + gin, full call graphs): **no regression attributable to this change** — gorm stable-edge delta 0/0; gin's only deltas are the **pre-existing** nondeterministic resolution of same-named functions in build-tag twin files (`validate`/`Default`: 27 edges flip base-vs-base with the **same binary** — demonstrated, not introduced here; noted as a latent issue). Container dispatch gained 0 on these corpora: gin's dispatch idiom is an attribute receiver (`c.handlers[i]`), out of the bare-identifier scope, unchanged.

**Tests** (per the issue's fixture requirements): the dispatcher's caller set **contains the dispatcher specifically**; the shadow negative — a function sharing the container's name gets **no** edge while the real targets do; generic-instantiation preservation; unknown-subscript abstention; method-value survival; direct-call controls throughout.

Siblings: Python #295 (PR #388) and C #298 (PR #391) are merged-adjacent; the remaining five parsers (Zig, Swift, PHP, Ruby, JS/TS, Rust) follow under this umbrella.

## Test plan

6 new Go tests in the builder's own test suite (real-file harness — file-scope containers only visible when the file is parsed) + the full existing Go suite + the end-to-end `test_call_graph_output.py` against the rebuilt binary.

<details>

| check | command | result |
|---|---|---|
| Go unit + suite | `go test ./...` (parsers/go/go_parser) | ok (6 new + existing) |
| mutation smoke | builder stashed → dispatch tests | FAIL to build (the machinery IS the change); restored → ok |
| corpus differential (gorm) | base vs fix, 3× stable-set | 873 edges; stable lost 0 / gained 0 |
| corpus differential (gin) | base vs fix + same-binary control | deltas = the pre-existing same-name flip class only (27 flips base-vs-base); getTyped[T] edges 34/34 preserved |
| end-to-end | `pytest tests/test_call_graph_output.py -q` | 17 passed (rebuilt binary) |
| full suite | `pytest tests/ -q` | 3187 passed, 2 failed (pre-existing zig local-env, stash-verified), 32 skipped |
| Go hygiene | `gofmt -l` + `go vet` | clean |

</details>

Refs #299 (1 of 7)

### DIFF
--- a/libs/openant-core/parsers/go/go_parser/callgraph.go
+++ b/libs/openant-core/parsers/go/go_parser/callgraph.go
@@ -6,6 +6,7 @@ import (
 	"go/token"
 	"path/filepath"
 	"strings"
+	"unicode"
 )
 
 // CallGraphBuilder builds call graphs from function information
@@ -20,6 +21,13 @@ type CallGraphBuilder struct {
 
 	// Import tracking per file
 	importsByFile map[string]map[string]string // file -> alias -> package_path
+
+	// #299: file -> container name -> function names referenced by the
+	// container's composite literal (var handlers = map[string]func(){...} /
+	// var tbl = []func(){...}). Dispatch through a known container
+	// (handlers[k]()) records edges to every referenced function —
+	// over-seed, the safe direction for reachability.
+	containersByFile map[string]map[string][]string
 
 	// Built-in functions to skip
 	builtins map[string]bool
@@ -62,13 +70,14 @@ func NewCallGraphBuilder(repoPath string) *CallGraphBuilder {
 	}
 
 	return &CallGraphBuilder{
-		repoPath:        repoPath,
-		fset:            token.NewFileSet(),
-		functionsByName: make(map[string][]string),
-		functionsByFile: make(map[string][]string),
-		methodsByType:   make(map[string][]string),
-		importsByFile:   make(map[string]map[string]string),
-		builtins:        builtins,
+		repoPath:         repoPath,
+		fset:             token.NewFileSet(),
+		functionsByName:  make(map[string][]string),
+		functionsByFile:  make(map[string][]string),
+		methodsByType:    make(map[string][]string),
+		importsByFile:    make(map[string]map[string]string),
+		containersByFile: make(map[string]map[string][]string),
+		builtins:         builtins,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review (sonnet + fable, compiled + executed repros) — holding with findings before merge. (1) HIGH: isTypeIndexExpr misclassifies exported-style VALUE keys (m[K]() where K is an exported const) as type-shaped — the exact false-edge class this PR fixes still reproduces for any container not built via composite literal (init()-populated registries, a common Go idiom). Compiled repro run against the real go_parser: a non-literal map whose name collides with an unrelated function still gets the fabricated edge via the uppercase-key unwrap + repo-unique fallback. (2) MEDIUM: compositeFuncTargets accepts any composite literal (map[string]int{} etc.) without a type-shape check — same insufficient-context class as the #388 lesson. (3) LOW: no nested-composite or method-value entries (parity gap vs the C parser's struct-array tables — plausibly deferred '1 of 7' scope). (4) LOW: TestContainerTargetsHaveDispatcherAsCaller re-asserts the forward graph — doesn't test its stated reverse-graph claim. (5) Fable also found an inverted-precedence defect between the container-dispatch path and analyzeCallExpr. Separately discovered (pre-existing, not this PR): the Go call-graph builder is NONDETERMINISTIC across runs — same input parsed twice yields different graphs (map iteration order feeding same-name resolution; reproduced with setProcGroupKill build-tag twins swapping arbitrarily). Recommend its own issue — it undermines future Go corpus differentials.

PR #393, opened 2026-08-29
fix(zig-parser): array/subscript dispatch records edges; file-scope containers visible (#299, 2 of 7)

gadievron:
## Summary

Zig already resolved the **struct-field** dispatch shape (`const h = T{ .cb = fn }; h.cb()` via `_collect_field_fn_bindings` — the closest in-repo template for this fix family). The **array** shape (`const tbl = [_]fn () void{ a, b }; tbl[i]()`) recorded nothing: the callee is an `index_expression`, a shape the callee-type check omits, and file-scope const declarations were invisible to the per-function alias index (the scope defect the umbrella documents). Issue #299's Zig member (2 of 7) — the narrowest gap of the nine.

Fix:
- `_collect_array_fn_bindings`: an array literal whose elements name **known** functions binds the bare name to every referenced function (the same set-semantics over-approximation as the plain-alias and field-init bindings; data arrays bind nothing — the `name_to_ids` gate);
- **subscript-call extraction**: an `index_expression` callee records its base name **only when that name is a known alias/container of the caller**, and the resolver expands it to the container's targets — an unknown base abstains (no bare-name false edges; the Go false-edge lesson applied proactively, negative-tested);
- `_collect_file_scope_aliases`: file-scope const declarations (the common placement for dispatch tables) are parsed once per file (cached) and merged under every function of that file; a function-local binding of the same name overrides (**BUG B9 per-function aliasing preserved** — the file pass walks top-level declarations only, and the existing u13 alias-semantics test guards it);
- the per-function walk and the file pass now share one `_collect_aliases_from_decl` (the binding shapes cannot drift apart).

**T3 differential** (real corpus: **zls** — the Zig language server; 43 files / 1052 units / 3786 edges): base vs fix, **LOST 0 / GAINED 0** — strict monotonicity (no regression; zls contains no bare-name array dispatch tables, so the mechanism is fixture-proven).

**Local-env note** (pre-characterization per the need-check): the 2 recorded "pre-existing zig local-env failures" are **not tree-sitter** — they are the zig subprocess (`test_pipeline.py`) failing to import `utilities` without an editable install; with `PYTHONPATH=<core>` the import resolves and they succeed — the full local suite reaches **3213/3213** under `PYTHONPATH` (CI, which installs the package, has always been clean). The standing record is corrected in this PR's evidence.

## Test plan

5 hermetic tests through the real pipeline (`parse_repository`): the file-scope array container dispatch (edges to every referenced function; the dispatcher specifically); a direct-call control; the struct-field template regression guard; a function-local container; the unknown-subscript abstention (no edge to a same-named function).

<details>

| check | command | result |
|---|---|---|
| zig parser suite | `PYTHONPATH=<core> pytest tests/parsers/zig/ -q` | 58 passed (5 new + existing incl. funcptr template + u13 alias semantics) |
| mutation smoke | builder stashed → new file | 2 failed (dispatch tests; negative controls pass by design); restored → 5 passed |
| T3 differential (zls) | base vs fix, full pipeline | 3786 edges / 3786 edges; **LOST 0**; GAINED 0 (strict monotone) |

</details>

Refs #299 (2 of 7)

### DIFF
--- a/libs/openant-core/parsers/zig/call_graph_builder.py
+++ b/libs/openant-core/parsers/zig/call_graph_builder.py
@@ -4,6 +4,7 @@ Stage 3: Call Graph Builder for Zig
 Builds bidirectional call graphs showing function dependencies.
 """
 
+import os
 import posixpath
 import re
 from collections import defaultdict
@@ -178,7 +179,13 @@ class CallGraphBuilder:
             # dispatches to A's foo, not to every same-named foo.
             var_types = self._collect_var_types(code)
 
-            calls = self._find_calls_in_code(code, file_path)
+            # #299 (2 of 7): the caller's alias NAMES gate subscript dispatch —
+            # `tbl[i]()` records the base name only when `tbl` is a known
+            # alias/container (so the resolver expands it to the referenced
+            # functions); an unknown base abstains rather than resolving the
+            # bare name (the Go false-edge lesson applied proactively).
+            alias_names = set(alias_to_target.get(func_id, {}).keys())
+            calls = self._find_calls_in_code(code, file_path, alias_names)
 
             for call_name in calls:
                 resolved_ids = self._resolve_call(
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review (fable + sonnet, executed) — HOLDING with findings. MED-HIGH: the per-function alias merge copies shared SET REFERENCES from the file map, then mutates them in place — a function-local shadowing container leaks into every function of the file (contradicts the B9 comment and the shadowing docstring). One-line fix: deep-copy on merge ({k: set(v)}). LOW: array bindings gated only by name-collision (no element type check); subscript-base takes first identifier child; except OSError only vs docstring claim.

PR #394, opened 2026-08-29
fix(swift-parser): container-literal dispatch records edges; file-scope lets visible (#299, 3 of 7)

gadievron:
## Summary

Swift dropped `tbl[i]()` entirely: the outer call's callee is an **inner** `call_expression` (the subscript `tbl[i]`), a shape `_extract_calls` handled only as `simple_identifier` or `navigation_expression`, and the inner node is excluded by the explicit `not _is_subscript` guard. Issue #299's Swift member (3 of 7), per the issue's **corrected** item 4 (the original narrowing is withdrawn — it fabricates edges):

- `_collect_aliases` accepts `array_literal` / `dictionary_literal` RHS: the variable binds to every **known** function among its elements (the `name_to_ids` gate — data containers bind nothing);
- `_build_alias_index` gains a **file-scope pass**: each file's top-level `let` bindings (plain aliases and container literals) are parsed once (cached) and merged under every function of that file; a function-local binding overrides (shadowing; top-level only, so per-function bindings cannot clobber each other);
- `_extract_calls` gains the subscript-callee case: `tbl[i]()` records its base name **only when it is a known alias/container of the caller** — tier-1 alias expansion then resolves it to the container's referenced functions. Unknown bases abstain (no bare-name false edges; the Go/Swift same-named-function negative the umbrella requires is tested);
- the `not _is_subscript` guard **stays** (it excludes genuine subscript-operator accesses from function-call resolution) and gains its rationale comment — the guard's docstring said what it detects, never what it prevents, which is how the withdrawn item 4 came to recommend removing it.

**T3 differential** (real corpora: **swift-argument-parser** 55 files + **Alamofire** 43 files; 2682 edges): base vs fix, **LOST 0 / GAINED 0** — strict monotonicity.

**Local-env note:** the swift-only test-directory batch shows a **pre-existing** ordering pollution (14 failed on pristine master in that batch; 0 failed in the full-suite ordering, which is what CI runs). Individually the affected tests pass. Not introduced here; noted as a follow-up.

## Test plan

5 hermetic tests through the real pipeline: array + dictionary + function-local container dispatch (the caller set contains the dispatcher specifically); the same-named-function negative (the umbrella's Go/Swift rule); a direct-call control; the unknown-subscript abstention.

<details>

| check | command | result |
|---|---|---|
| new tests | `PYTHONPATH=<core> pytest tests/parsers/swift/test_issue299_swift_container_dispatch.py -q` | 5 passed (2 failed RED on pristine) |
| mutation smoke | builder stashed → new file | 2 failed (dispatch tests; negatives by design); restored → 5 passed |
| T3 differential | swift-argument-parser + Alamofire, base vs fix | 2682 edges / 2682 edges; **LOST 0**; GAINED 0 (strict monotone) |

</details>

Refs #299 (3 of 7)

### DIFF
--- a/libs/openant-core/parsers/swift/call_graph_builder.py
+++ b/libs/openant-core/parsers/swift/call_graph_builder.py
@@ -22,6 +22,7 @@ reachability recall while avoiding gross same-name over-connection:
             -> caller -> referenced function (callback reachability)
 """
 
+import os
 from collections import defaultdict
 from typing import Any, Dict, List, Optional, Set, Tuple
 
@@ -200,7 +201,12 @@ class CallGraphBuilder:
             caller_class = func_info.get("class_name")
 
             var_types, local_names, var_qualified = self._collect_var_types(code, type_names)
-            call_sites, arg_refs = self._find_calls_in_code(code, file_path)
+            # #299 (3 of 7): the caller's alias NAMES gate subscript dispatch —
+            # `tbl[i]()` records the base only when `tbl` is a known
+            # alias/container (tier-1 expansion then resolves it); an unknown
+            # base abstains rather than resolving the bare name.
+            alias_names = set(alias_to_target.get(func_id, {}).keys())
+            call_sites, arg_refs = self._find_calls_in_code(code, file_path, alias_names)
 
             for site in call_sites:
                 sites_total += 1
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review — HOLDING. HIGH: the new subscript-callee branch ends in 'stack.extend(callee.children); continue', skipping _collect_arg_refs + the common child traversal for the OUTER call — any call/ref inside the dispatch call's arguments loses its edge (a dropped-edge regression, the series' only strictly-worse-than-base behavior; the continue fires even for UNKNOWN bases). Fix: fall through to the common tail. MED-HIGH: same shared-set mutation leak as zig. MED: test_local_and_dictionary_forms asserts handlers the array form alone already produces — the dictionary/local paths are unverifiable (possibly dead: the '!' postfix shape may never match callee.type == call_expression); needs isolated per-form fixtures.

PR #395, opened 2026-08-29
fix(php-parser): string-name dispatch containers resolve on subscript calls (#299, 4 of 7)

gadievron:
## Summary

PHP's dispatch-table idiom holds function names as **strings** — `$handlers = ['a' => 'handlerA', 'b' => 'handlerB']; $handlers[$k]();` — and `_resolve_function_call` had **no subscript case**: the callee is a `subscript_expression`, so nothing resolved and the dispatcher's targets were orphaned (pruned with their dispatcher kept — #295's shape). Issue #299's PHP member (4 of 7). The in-repo template is `_resolve_variable_function` (the single string-literal binding `$f = 'helper'; $f()` already resolves): **the container is the array form of that binding**.

Fix:
- **container collection**: `$var = array-literal-of-known-function-strings` (both the `'k' => 'fn'` value form and the bare list form) binds `$var` to every referenced function name — locally in the caller's body, plus a file-scope pass (top-level statements only, cached per file — the common placement, reached via `global $var;`), with local bindings overriding;
- **subscript dispatch**: a `function_call_expression` whose callee is a `subscript_expression` over a **known** container resolves to every function the container's strings reference (over-seed, the safe direction); unknown bases abstain — the base is a `$variable`, never a function name, so no false edge to a same-named function is possible (tested).

**T3 differential** (real corpora: **flysystem 3.x** 55 files + **psr/log** 8 files; 112 edges): base vs fix, **LOST 0 / GAINED 0** — strict monotonicity.

## Test plan

5 hermetic tests through the real pipeline: file-scope container dispatch (with the `global` import shape; the caller set contains the dispatcher specifically); a direct-call control; a function-local container; the bare list form; the data-array abstention negative.

<details>

| check | command | result |
|---|---|---|
| new tests | `PYTHONPATH=<core> pytest tests/parsers/php/test_issue299_php_container_dispatch.py -q` | 5 passed (3 failed RED on pristine) |
| PHP suite | `PYTHONPATH=<core> pytest tests/parsers/php/ -q` | 97 passed |
| mutation smoke | builder stashed → new file | 3 failed; restored → 5 passed |
| T3 differential | flysystem + psr/log, base vs fix | 112 edges / 112 edges; **LOST 0**; GAINED 0 (strict monotone) |

</details>

Refs #299 (4 of 7)

### DIFF
--- a/libs/openant-core/parsers/php/call_graph_builder.py
+++ b/libs/openant-core/parsers/php/call_graph_builder.py
@@ -776,8 +776,10 @@ class CallGraphBuilder:
         return None
 
     def _container_string_names(self, array_node, source: bytes) -> Set[str]:
-        """The string-literal elements of an array literal (both the
-        `'k' => 'fn'` value form and the bare `'fn'` list form)."""
+        """The string-literal elements of an array literal: the VALUE after
+        an `=>` arrow (`'k' => 'fn'` — the key is NEVER a target: a key like
+        'init'/'get' colliding with a real function would fabricate an
+        edge), or a bare list-form string element (`['fn', ...]`)."""
         names: Set[str] = set()
         for element in array_node.children:
             if element.type == 'string':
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review — HOLDING. HIGH: dispatch-table KEYS fabricate edges. In _container_string_names, for 'k' => 'fn' the KEY string is collected before the arrow is seen (taken starts False), so a key like 'init'/'get' colliding with a real function fabricates an edge — contradicting the 'value after the arrow' docstring; seen_arrow is assigned and never read (dead). Fix: collect only post-arrow values. MED: file-scope containers merged into every function with no 'global $var;' check (docstring claims the global requirement); a local non-array rebind doesn't suppress the file binding. LOW: rebind-drop pops on the 2nd binding but re-adds on the 3rd; dead scratch set; except OSError only.

PR #396, opened 2026-08-29
fix(ruby-parser): dispatch containers (method(:sym)/symbol tables) resolve on .call and send (#299, 5 of 7)

gadievron:
## Summary

Ruby's dispatch-table idiom: a CONSTANT assigned a hash/array of `method(:sym)` objects or symbols — `HANDLERS = { 'a' => method(:handlerA) }` then `HANDLERS[k].call`, or `SYMS = [:handlerA, :handlerB]` then `send(SYMS[k])`. The builder resolved `.call` only via **local** method-object bindings (`m = method(:helper)`), so a subscript receiver over a container resolved nothing and the dispatcher's targets were orphaned. Issue #299's Ruby member (5 of 7).

Fix:
- **container collection**: `name = hash/array` whose elements are `method(:sym)` calls or symbols binds the name to every referenced function name — locally in the caller's body, plus a file-scope pass (Ruby constants — the common placement, parsed once per file and cached), local bindings overriding; a rebound name is dropped rather than guessed;
- **dispatch is resolved in the walk (multi-target)**: `HANDLERS[k].call` over a **known** container resolves to every referenced function, as does `send`/`public_send`/`__send__` whose argument element-references a known container. Unknown bases fall through to the existing (abstaining) resolution — no invented edges (tested).

**T3 differential** (real corpora: **ruby-ffi** 123 files + **rack** 50 files; 470 edges): base vs fix, **LOST 0 / GAINED 0** — strict monotonicity.

## Test plan

5 hermetic tests through the real pipeline: the method-object container dispatch (`.call` over a subscript; the caller set contains the dispatcher specifically); the symbol-container `send` form; a direct-call control; a function-local container; the unknown-subscript abstention negative.

<details>

| check | command | result |
|---|---|---|
| new tests | `PYTHONPATH=<core> pytest tests/parsers/ruby/test_issue299_ruby_container_dispatch.py -q` | 5 passed (3 failed RED on pristine) |
| Ruby suite | `PYTHONPATH=<core> pytest tests/parsers/ruby/ -q` | 75 passed |
| mutation smoke | builder stashed → new file | 3 failed; restored → 5 passed |
| T3 differential | ruby-ffi + rack, base vs fix | 470 edges / 470 edges; **LOST 0**; GAINED 0 (strict monotone) |

</details>

Refs #299 (5 of 7)

### DIFF
--- a/libs/openant-core/parsers/ruby/call_graph_builder.py
+++ b/libs/openant-core/parsers/ruby/call_graph_builder.py
@@ -32,6 +32,7 @@ Output (JSON):
 """
 
 import json
+import os
 import posixpath
 import re
 import sys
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review — HOLDING. HIGH: _file_containers walks the WHOLE file tree (no method-body exclusion) — a method-local container leaks file-wide into every other method's dispatch (the exact recurring scope-leak class from the C parser review); docstring claims 'File-scope CONSTANT containers' — false on both counts. Fix: top-level-only walk skipping method/class/module bodies (as the PHP sibling does). MED: element-name walk extends into pair children, so rocket-syntax symbol KEYS (:foo => method(:x)) are added as targets. LOW: unused containers parameter on _resolve_call_node; rebind 3rd-assignment re-add; the negative test uses data[k] without .call so it never exercises the dispatch path.

PR #421, opened 2026-08-30
fix(runtime): deterministic installs — requirements.txt (the exact CI pins) before the editable install; deps hash covers both files

### LINKED ISSUE #59: End-user runtime installs use floor pins while CI uses exact pins (dev/CI/prod drift)
## Problem

The runtime `installOpenant` in `apps/openant-cli/internal/python/runtime.go` runs:

```go
exec.Command(pythonPath, "-m", "pip", "install", "-e", corePath)
```

That resolves dependencies from `pyproject.toml`'s `[project.dependencies]`, which uses **floor pins** (`anthropic>=0.40.0`, `pydantic>=2.0.0`, etc.). pip will pull whatever the latest version on PyPI is that satisfies the floor.

CI (`.github/workflows/test.yaml`) runs a different command:

```yaml
pip install -r requirements.txt && pip install ".[dev]"
```

Where `libs/openant-core/requirements.txt` has **exact pins** (`anthropic==0.75.0`, `pydantic==2.12.5`, etc.).

### What this means in practice

- A contributor running tests locally or in CI uses `anthropic==0.75.0`.
- An end user who runs `openant scan` today might get `anthropic==0.78.0` — whatever PyPI currently serves as the latest matching `>=0.40.0`.
- Three months from now, a fresh `openant scan` could pull `anthropic==1.2.0` if upstream has had releases.

This is a recipe for "works on my machine" / "works in CI" / "breaks for users" gaps that are hard to reproduce. SDK constructor changes (cf. the `AuthenticationError(message=, response=, body=)` shift in #56), default-model swaps, response-shape changes — any of these can land for end users without ever being exercised in CI.

### Where this came up

While reviewing #36 (auto-detect dep changes and reinstall), the question came up of whether the staleness check should also hash `requirements.txt`. The answer was no — `installOpenant` doesn't read `requirements.txt`, so hashing it would trigger reinstalls that don't actually apply the new pins, making the UX worse. PR #36's `pyproject.toml`-only scope is correct given today's `installOpenant` behavior. The divergence itself is a separate, pre-existing problem.

## Proposed fix

Make `installOpenant` mirror what CI does:

1. `pip install -r requirements.txt` first (apply exact pins).
2. Then `pip install -e .` (install the package itself).
3. Have #36's `CheckDepsStale` hash both files so a change to either triggers reinstall.

That makes end-user installs deterministic and matches what's exercised in CI.

### Open questions

- Keep the dual-file system (`pyproject.toml` floor + `requirements.txt` lock), or move to a single source of truth (e.g., generate `requirements.txt` from `pyproject.toml` via `pip-compile` or `uv lock`)?
- The `.[dev]` extras (pytest, etc.) in CI — survive into end-user installs, or stay CI-only?

## Why this isn't in #36

PR #36's scope was "auto-detect dep changes". Changing how `installOpenant` behaves is a behavior change that deserves its own review. Filing this so it doesn't get lost.

## References

- #36 — auto-detect dep changes (the PR that surfaced this)
- `apps/openant-cli/internal/python/runtime.go` — `installOpenant`
- `.github/workflows/test.yaml` — CI's install command
- `libs/openant-core/requirements.txt` vs `libs/openant-core/pyproject.toml`

gadievron:
Hi @ar7casper — your analysis was right (the floor-vs-exact divergence is a real reproducibility gap), and this implements your three-part proposal as written.

## Summary

`installOpenant` ran `pip install -e <corePath>`, resolving `pyproject.toml`'s **floor pins** — whatever PyPI currently serves. CI runs `pip install -r requirements.txt` (the **exact pins**). A user running `openant scan` today gets `anthropic==1.2.1` while CI tested `==1.2.0`; three months from now a fresh install could pull an SDK the project never tested — SDK constructor shifts, default-model swaps, response-shape changes landing on users without ever being exercised in CI.

Fix (the issue's proposal, all three parts):
- `installOpenant` installs **requirements.txt first** (the exact CI pins), then the editable install — the mirror of what CI does. The editable pass does not upgrade what was pinned (pyproject's floors are compatible with the pins by construction). A dev layout without requirements.txt **degrades** to the old behavior (the editable install alone);
- `depsHash` covers **both** `pyproject.toml` AND `requirements.txt`: a pin change triggers a reinstall the same way a pyproject change does. The `corePath` key component is preserved (the two-worktrees hazard), and a missing requirements.txt is not an error;
- `installOpenant` extracted into `installOpenantCmds` (the command-builder seam) so the composed install is directly testable.

**Pre-submission experimentation** (the new gate): the real fixed binary driven end-to-end (`openant parse` on a live fixture → units extracted, full artifacts written) against the accumulated merged state — no previous fix's regression tests hurt (the #295/#312/#303 suites green).

## Test plan

5 Go tests: the composed install (requirements first, then editable — the command sequence); the degraded path (no requirements.txt → 1 command); the deps-hash covers requirements.txt (a pin change triggers); corePath still in the key (the two-worktrees hazard preserved); a missing requirements.txt abstains (not an error every run).

<details>

| check | command | result |
|---|---|---|
| new tests | `go test ./internal/python/ -run 'TestInstallCommand|TestDepsHash' -count=1 -v` | 5 PASS |
| Go all packages | `go test ./... -count=1` | all ok |
| hygiene | `go vet ./internal/python/` + `gofmt` | clean |
| merged-state Python suite | the full pytest run | 3340/3340 green |
| step-10.5 gate | the real fixed binary `parse` end-to-end | 2 units extracted, full artifacts written |
| hermeticity | the env-scrubbed run | green |

</details>

Fixes #59

### DIFF
--- a/apps/openant-cli/internal/python/runtime.go
+++ b/apps/openant-cli/internal/python/runtime.go
@@ -294,7 +294,16 @@ func depsHash(corePath string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	sum := sha256.Sum256(append([]byte(corePath+"\x00"), pyproject...))
+	// #59: the staleness key covers requirements.txt too — installOpenant
+	// now installs it (the exact CI pins), so a pin change must trigger a
+	// reinstall the same way a pyproject change does. A MISSING file is not
+	// an error (the dev-layout degrade path): hash the corePath+pyproject
+	// alone, matching the old key.
+	reqs, reqsErr := os.ReadFile(filepath.Join(corePath, "requirements.txt"))
+	if reqsErr != nil {
+		reqs = nil // degrade: the dev layout without requirements.txt
+	}
+	sum := sha256.Sum256(append(append([]byte(corePath+"\x00"), pyproject...), reqs...))
 	return hex.EncodeToString(sum[:]), nil
 }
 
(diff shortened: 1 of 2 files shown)


gadievron:
Panel review (full analysis): the install ordering is right and the depsHash now covers both files — verified. Two findings worth tracking: (1) the determinism claim overstates — pyproject floors are still re-resolved by the editable install and only the floors<=pins invariant (currently hand-maintained, no check) keeps them from drifting; boto3/botocure from the bedrock extra is unpinned. Filed as issue for a consistency check. (2) TestInstallCommandRunsRequirementsThenEditable opens with a dead cmds build on an empty TempDir (`_ = cmds`) — duplicate scaffolding of TestInstallCommandDegradedWithoutRequirements. No blocking issue: CI green, hash covers both files, ordering correct.

Issue #425, opened 2026-08-30
stage1_consistency rewrites verdict without validating it and never updates finding — an upgrade is inert for every finding-first consumer, and an unrecognized should_be lands in results verbatim


gadievron:
## Summary

`apply_stage1_consistency` rewrites a Stage-1 result's `verdict` from the consistency model's
unconstrained `should_be` string — **after** verdict normalization has run — and never updates
`finding`. Two consequences, both executed at current master:

1. **An unrecognized verdict lands in results verbatim.** `stage1_consistency.py:265` applies
   only `.strip().upper()`; the downgrade guard at `:287-296` blocks only values in
   `_DISCLOSURE_DROPPED_UPPER`. Anything else — `"MAYBE VULNERABLE"`, `"probably fine"` — is
   written straight into `result["verdict"]` (`:298`). This is the same escape #316/#324 just
   closed at `analysis_core._normalize_result` / the JSON corrector, surviving at a producer
   that runs *after* those gates.
2. **A legitimate upgrade is inert end-to-end.** The loop never touches `result["finding"]`,
   and every downstream selector is finding-first: `str(r.get("finding") or r.get("verdict", ""))`
   (`core/verifier.py:118`, `:357`, `core/reporter.py:348`). A SAFE→VULNERABLE consistency
   upgrade therefore leaves `finding == "safe"`, the selector sees `"safe"`, and the upgraded
   unit never reaches Stage 2 or disclosure. The re-run `_count_verdicts` (analyzer.py:772)
   is finding-first too, so the metrics still count the old verdict.

## Executed

```
post-consistency row {"finding": "safe", "verdict": "VULNERABLE"} (an upgrade):
  selector sees 'safe'   -> selected for Stage-2: False
post-consistency row {"finding": "vulnerable", "verdict": "MAYBE VULNERABLE"} (garbage should_be):
  selector sees 'vulnerable' -> selected for Stage-2: True   (the verdict is never consulted)
_count_verdicts([both]) -> {'safe': 1, 'vulnerable': 1, ...}   (both keyed on finding, not verdict)
```

## Suggested fix

Apply the same producer discipline #316/#324 established: validate `new_verdict` against the
canonical verdict set (reject/ERROR unrecognized values), and update **both** keys together
(`verdict` + `finding`, the one-shape rule) so finding-first consumers see the upgrade. The
F-KB-1a downgrade guard stays as is — it is correct for its purpose; it is just not a
*validity* gate.

## What I am not claiming

[...]

Issue #293, opened 2026-08-19
Progress/summary counters use a two-branch `error` vs `else` split over three states, so incomplete work is counted as completed (5 sites)


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

Several per-unit counters branch on `error` versus *everything else*, but the pipeline has three
outcomes, not two: completed, incomplete (the loop never reached a verdict), and errored. The
incomplete case falls into the `else` and is counted as completed. On our run the verify checkpoint
summary reported **175 completed** where only **41** verifications actually produced a verdict.

The same two-branch shape appears at five sites. Two correct three/four-way handlers already exist in
the repository to copy — one of them in the same module as one of the broken sites.

### Runtime proof

```
RUN=~/.openant/projects/gadievron/raptor-e2e-20260818/scans/7dbf9c691d7/python
cat $RUN/verify_checkpoints/_summary.json
```

```json
{
  "step": "Verify", "phase": "done", "total_units": 223,
  "completed": 175,
  "errors": 48,
  "error_breakdown": { "api": 48 }
}
```

Ground truth, by enumerating the 223 per-unit checkpoint files and keying on the shape of their
`verification` object (`{"incomplete": true}` alone = errored; `incomplete: true` with a verdict =
incomplete; otherwise complete):

```
python3 - <<'PY'
import json, glob, os
cps = [c for c in glob.glob(os.path.expanduser("$RUN/verify_checkpoints/*.json"))
       if os.path.basename(c) not in ("_summary.json", "_fingerprint.json")]
inc = err = comp = 0
for c in cps:
    v = json.load(open(c)).get("verification") or {}
    if set(v.keys()) == {"incomplete"}: err += 1
    elif v.get("incomplete"):           inc += 1
    else:                               comp += 1
print("genuinely complete:", comp, "| incomplete:", inc, "| errored:", err, "| total:", comp+inc+err)
PY
```

```
genuinely complete: 41 | incomplete: 134 | errored: 48 | total: 223
```

[...]

PR #426, opened 2026-08-30
fix(analyze): an unrecognized or absent Stage-1 verdict routes to the error accounting — not a silently uncountable one (#316, #324)

### LINKED ISSUE #316: JSON corrector synthesises a verdict from free text and reports it as a successful correction, producing rows no bucket counts
## Summary

The JSON corrector's verdict-recovery path **synthesises** a verdict string from model-supplied free
text, and its success gate then checks only that the key is *present* — not that the value is a verdict the
rest of the pipeline recognises. A recovered record can therefore carry something like
`"MAYBE EXPLOITABLE?"` and be returned as a successful correction.

That an unrecognised verdict is dropped from every bucket downstream is already known and documented
(see below). What this issue reports is a **producer** of such verdicts.

## The synthesis

`libs/openant-core/utilities/json_corrector.py:268-274`:

```python
                    mapping = {
                        "vulnerable": "VULNERABLE", "safe": "SAFE",
                        "protected": "PROTECTED", "bypassable": "BYPASSABLE",
                        "inconclusive": "INCONCLUSIVE",
                        "insufficient_context": "INSUFFICIENT_CONTEXT",
                    }
                    extracted["verdict"] = mapping.get(finding.lower(), finding.upper())
```

The `mapping.get(..., finding.upper())` default is the issue: any text that is not one of the six
known findings becomes a verdict by upper-casing it.

**Scope of the affected path.** This block is reached only in vuln mode — `:230-232`:

```python
        # Non-verdict schemas skip the vuln-specific finding->verdict
        # normalization; their success gate is their own required_keys.
        vuln_mode = schema is None
```

so callers passing an explicit `schema=` skip it entirely, and it additionally requires `"finding"`
to be present in the extracted object. The input is model-supplied free text on that path, not on
every path.

**Updated 2026-08-22 — which callers those are.** `/usr/bin/grep -rn "attempt_correction("` returns
exactly one non-test, non-utility production call site on the Stage-1 analysis route, and it passes
**no** schema — `core/analysis_core.py:292`. So `vuln_mode` is True there and this block is
unconditionally live on the default analysis path; it is not an edge case. The adoption gate two
lines later (`:294`) tests only `not in ("ERROR", None)`, which a synthesised verdict passes, and the
intervening `_normalize_result` (`analysis_core.py:33-69`) only uppercases the verdict and coerces
non-string findings — there is no enum check between the corrector and `_count_verdicts`.

## The gate does not catch it

`libs/openant-core/utilities/json_corrector.py:276-280`:

```python
            # Validate the extracted data has the caller's required fields
            if all(k in extracted for k in required_keys):
                extracted["json_corrected"] = True
                print(f"      JSON correction successful! keys={list(extracted.keys())}", file=sys.stderr)
                return extracted
```

The comment says "Validate", and the check is membership of the *key*. The synthesised value is never
compared against the known verdict set.

Executed at `b5019628`, replicating those two steps:

```
model text 'vulnerable'            -> verdict='VULNERABLE'            passes gate=True
model text 'maybe exploitable?'    -> verdict='MAYBE EXPLOITABLE?'    passes gate=True
model text 'I could not determine' -> verdict='I COULD NOT DETERMINE' passes gate=True
```

All three are reported as `JSON correction successful!`.

## Where such a record lands

`libs/openant-core/core/analyzer.py:330-335`:

```python
    for r in results:
        finding = r.get("finding", r.get("verdict", "error").lower())
        if finding in counts:
            counts[finding] += 1
        elif r.get("verdict") == "ERROR":
            counts["errors"] += 1
```

A synthesised verdict matches neither branch: it is not a known bucket, and it is not the literal
`"ERROR"`. It is counted nowhere. `libs/openant-core/core/verdict_taxonomy.py:74-81` likewise does not contain it, so
it is not disclosure-eligible.

**This downstream behaviour is already documented, and I am not re-reporting it.**
`libs/openant-core/tests/report/test_results_bucket_reconciliation.py` states it in its own docstring — that
`_count_verdicts` "drops any row whose verdict is unrecognized … from ALL buckets", explicitly out of
that test's scope. The gap is known; what was not stated is that the corrector manufactures rows of
exactly that shape, and reports doing so as a success.

## Suggested fix

1. Validate the synthesised value, not just the key's presence: check membership against the known
   verdict set before returning, and treat a non-member as a failed correction rather than a
   successful one.
2. Alternatively, drop the `finding.upper()` default so an unrecognised finding does not become a
   verdict at all — return the failure path instead.
3. Either way, the log line at `:279` should not read `JSON correction successful!` for a record the
   pipeline cannot count.

The direction matters here: a failed correction is visible (it takes the error path, which *is*
counted), whereas a synthesised verdict is silent. Failing loudly is the safer outcome for a scanner.

## What I am not claiming

- I am **not** claiming this fired on any real run. The execution above replicates the two cited
  steps directly; I have not observed a production record carrying a synthesised verdict.
- I am **not** re-reporting the unrecognised-verdict drop itself — that is documented, as quoted
  above.
- I have **not** measured how often the corrector's recovery path runs, nor what fraction of its
  inputs miss the mapping.

---

## Correction 2026-08-21 — this issue was scoped too narrowly, and its fixes are insufficient

I filed this against the JSON corrector's recovery path. **The identical synthesis is on the primary
parse path**, where no corrector is involved, and both fixes above leave that live.

`libs/openant-core/core/analysis_core.py:49-57`:

```python
            finding_to_verdict = {
                "vulnerable": "VULNERABLE",
                "safe": "SAFE",
                "protected": "PROTECTED",
                "bypassable": "BYPASSABLE",
                "inconclusive": "INCONCLUSIVE",
                "insufficient_context": "INSUFFICIENT_CONTEXT",
            }
            result["verdict"] = finding_to_verdict.get(finding.lower(), finding.upper())
```

Executed at `b5019628` on a **well-formed** reply, no correction involved:

```
_normalize_result({"finding": "maybe exploitable"})
  -> {'finding': 'maybe exploitable', 'verdict': 'MAYBE EXPLOITABLE'}
_count_verdicts([that])
  -> {'vulnerable': 0, 'bypassable': 0, 'inconclusive': 0, 'protected': 0, 'safe': 0, 'errors': 0}
     sum = 0
```

**The maintainers already identified this exact escape and closed it for one branch.**
`analysis_core.py:42-47`, immediately above:

> if not isinstance(finding, str):
>     # A non-string finding (list/dict/null/number) is a malformed model reply,
>     # not a verdict — map it to ERROR so the error / manual-review accounting
>     # counts it, instead of a garbage verdict (e.g. "['VULNERABLE']" from
>     # str(finding).upper()) that silently escapes that accounting.
>     result["verdict"] = "ERROR"

That is precisely the failure this issue describes — "a garbage verdict that silently escapes that
accounting" — and the fix pattern is already chosen: map to `ERROR`. It was applied to the
non-string branch and not to the string branch one line below.

**Two further corrections to the text above:**

- The scoping paragraph is incomplete. The corrector's synthesis also requires `"verdict" not in
  extracted` (`json_corrector.py:266`), and there is a **second** synthesis site at `:260-263` using
  `correct_finding.upper()` with no mapping at all — executed, `'Not sure, maybe'` yields
  `verdict='NOT SURE, MAYBE'`, also uncounted. Suggested fix 2 does not reach it.
- "A synthesised verdict matches neither branch: it is not a known bucket" mis-states the mechanism.
  `_count_verdicts` tests `finding`, which is present on this path; the verdict's role is passing the
  gate at `:277`. The row goes uncounted because `finding` is unrecognised, not because the verdict is.

**Revised fix.** Apply the `:42-47` pattern to the string branch: an unrecognised `finding` should
become `ERROR` (counted, manual-review-visible) rather than an upper-cased passthrough. Then the two
corrector sites inherit the same treatment. If the passthrough is wanted for forward-compatibility
with new verdicts, the accounting needs a bucket for unrecognised values so they cannot vanish —
but the maintainers' own comment argues against silent passthrough.

**Citation fix (applied above):** the fence was cited `:330-336` while quoting only `:330-335`;
`:336` is `return counts`. The heading now reads `:330-335`.

### The harm is worse than this issue originally described

I framed the consequence as "counted nowhere". The row is also **excluded from the findings set and
never retried**, so it disappears completely rather than merely going uncounted:

- **Excluded from the vulnerable set.** `core/reporter.py:313-316` and `core/verifier.py:105-108`
  both select on `str(r.get("finding") or r.get("verdict","")).lower() in ("vulnerable","bypassable")`.
  A synthesised verdict matches neither, so the record never reaches disclosures.
- **Not treated as an error, so never retried.** `core/analyzer.py:237-239`:

```python
    def _cp_is_error(cp_data):
        res = cp_data.get("result", {}) if cp_data else {}
        return res.get("verdict") == "ERROR" or res.get("finding") == "error"
```

  `"MAYBE EXPLOITABLE"` is neither `"ERROR"` nor `"error"`, so the checkpoint counts as a success and
  is adopted on resume. The unit is never re-analysed.

That completes the chain: a unit the model gave an unrecognised verdict for is silently dropped from
the findings, absent from every counter, and permanently checkpointed as done. It is exactly the
outcome `analysis_core.py:42-47` was written to prevent for the non-string case — which is the
strongest argument for applying the same `ERROR` mapping to the string branch.

---

**Cross-link to #324 (added 2026-08-22).** The corrected section above reaches
`core/analyzer.py:237-239` (`_cp_is_error`) and `:330-335` (`_count_verdicts`). **#324 reaches those
same two sites from a different producer** — a parsed object carrying *neither* `verdict` nor
`finding`, where this issue's producer emits an unrecognised finding *string*.

The producers do not overlap: `core/analysis_core.py:40` gates on
`if "verdict" not in result and "finding" in result`, so the branch this issue's revised fix targets
runs no code for #324's shape.

**A consumer-side fix closes both; a producer-side fix closes neither.** If you fix this issue at the
producer, #324 stays live — and vice versa. Worth scheduling the two functions once, together.

### LINKED ISSUE #324: A Stage-1 result with neither `verdict` nor `finding` is counted in no bucket and adopted as complete on resume, so `units_analyzed` overstates and the unit is never retried
## Summary

A Stage-1 result that carries **neither** `verdict` **nor** `finding` is produced whenever the model
returns a valid JSON object without those keys — a JSON-shaped refusal, a `{"reasoning": ...}` reply,
a truncated-schema reply. Such a result:

1. is **counted in no verdict bucket** by `_count_verdicts`, including `errors`, and
2. is **adopted as successfully-complete on resume**, because `_cp_is_error` does not recognise it.

Both guards test two specific error shapes. This is a third.

## The producer

`libs/openant-core/core/analysis_core.py:90-95` returns `_normalize_result(result)` for any top-level
JSON object. `_normalize_result` (`:33-69`) synthesises `verdict` *from* `finding`; with neither
present it adds `cwe_id`/`cwe_name` and returns. `analyze_unit:286-295` attempts JSON correction but
keeps the original whenever the correction also yields `ERROR`/`None`, and `_process_unit:149-152`
sets `finding` only when `finding` or `verdict` is truthy — so neither key is ever written.

## Executed at `b5019628`

```
parse_response('{"reasoning": "cannot determine", "confidence": 0}')
  -> keys: ['confidence', 'cwe_id', 'cwe_name', 'reasoning']
  -> has verdict: False   has finding: False

_count_verdicts([r])                  -> {vulnerable:0, bypassable:0, inconclusive:0,
                                          protected:0, safe:0, errors:0}
_count_verdicts([{"verdict":"ERROR"}]) -> {... errors:1}      # control
_count_verdicts([{"finding":"safe"}])  -> {... safe:1}        # control

cp_is_error({"result": r})                     -> False   # adopted as complete
cp_is_error({"result": {"verdict":"ERROR"}})   -> True     # control
```

Both controls behave correctly, so the zero is specific to this shape.

## Why the guards miss it

`core/analyzer.py:237-239`:

```python
    def _cp_is_error(cp_data):
        res = cp_data.get("result", {}) if cp_data else {}
        return res.get("verdict") == "ERROR" or res.get("finding") == "error"
```

Neither arm matches, so `:244-246` pre-populates the unit as a completed result and it is never
re-analysed. The comment directly above (`:230-232`) states the intended contract — *"Pre-populate
results from checkpoints, but ONLY for successfully-completed units"* — which this violates.

`core/analyzer.py:330-335`:

```python
    for r in results:
        finding = r.get("finding", r.get("verdict", "error").lower())
        if finding in counts:
            counts[finding] += 1
        elif r.get("verdict") == "ERROR":
            counts["errors"] += 1
```

With neither key present the default is the string `"error"` — but the bucket is named `"errors"`,
so `finding in counts` is False; and `r.get("verdict") == "ERROR"` is False. The unit increments
nothing.

## Downstream effect

`_count_verdicts` feeds `AnalysisMetrics` (`analyzer.py:730-738`), and `core/reporter.py:556`
computes `units_analyzed = total_units - metrics["errors"]`. Because such units never reach
`errors`, `units_analyzed` **overstates** what was actually analysed, and the bucket-reconciliation
invariant documented at `reporter.py:562-570` breaks silently.

There is no in-run retry either: `analyzer.py:622-625` keys on
`is_retryable_error(r.get("error"))`, and this result has no `error` key.

## Not a hypothetical class

`libs/openant-core/tests/test_checkpoint_singleshot_error_key.py` is a regression test for the same
class — an error shape the classifier missed, causing a unit to be counted as completed. That one was
fixed. This is a second instance of it.

## Scope

Stage 1 only. The other phases fail closed correctly:

- `utilities/finding_verifier.py:603-611` treats a missing or empty `verification` as errored.
- `utilities/dynamic_tester/__init__.py:193-216` retries `status == "ERROR"`, and `SKIPPED` results
  are never checkpointed (`:229-233`), so there is no adoption hole.

## Suggested fix

1. Make `_cp_is_error` fail closed: treat a result lacking **both** `verdict` and `finding` as
   errored, so it is retried rather than adopted.
2. Fix the bucket-name mismatch in `_count_verdicts` — the default `"error"` can never match the
   `"errors"` key, so make the fallback explicit and route unrecognised shapes to `errors`.
3. Consider having `_normalize_result` stamp an explicit `verdict = "ERROR"` when the parsed object
   contains neither key, so every downstream consumer sees one shape instead of three.

Direction: units currently adopted as complete would be re-analysed, and the error count would stop
under-reporting — more units examined, which is the safe direction for a scanner.

## What I am not claiming

- I am **not** claiming this fired in a specific real run. The execution above uses the real
  `parse_response` and the real `_count_verdicts`, but the triggering model reply is constructed.
- I have **not** measured how often a model returns JSON with neither key.
- I am **not** re-reporting the unrecognised-verdict drop already documented in
  `tests/report/test_results_bucket_reconciliation.py`; this is about a result carrying no verdict
  field at all, and about the resume-adoption half, which that note does not cover.

---

### Grounding curve (added 2026-08-22)

| commit | producer (`_normalize_result`) | `_cp_is_error` misses shape | `_count_verdicts` misses shape | verdict |
|---|---|---|---|---|
| `d710b90` (PR #23) | *does not exist* | true | true | PARTIAL — consumers blind, no producer |
| `8d31c32` (PR #199) | **added** | true | true | FULL |
| `b5019628` (HEAD) | present | true | true | FULL |

**Attribution, and it sharpens the fix.** `git blame` shows `_cp_is_error` (`analyzer.py:237-239`)
and `_count_verdicts` (`analyzer.py:330-335`) have been **byte-identical since `d710b90`
(PR #23, 2026-04-14)** — the two-arm blind spot is four months old and was never touched by the later
reporter PRs that adjust "errors" accounting nearby. The producer that can actually yield the
gap-shape, `_normalize_result`, is newer: `8d31c32` (PR **#199**, 2026-07-30).

So the consumer-side defect lay **dormant for months waiting for a producer that could hit it**, and
PR #199 supplied one. That is why this has not been noticed: neither commit is wrong in isolation.

---

> **Corrected 2026-08-22 — prior art I should have cited, and a re-centering.**
>
> **The counting half of this issue is already documented as a known gap.** Merged **PR #230**
> (*"fix(report): include errors in results so buckets reconcile to total"*) states in its own body
> that it *"documents, as an explicitly out-of-scope pre-existing gap, that `analyzer._count_verdicts`
> drops an unrecognized verdict"*. I filed the bucket-miss above as though it were new. It is not,
> and a maintainer would be right to say so.
>
> **What is not covered by that admission, and is the point of this issue:**
> 1. **The resume-adoption half.** `_cp_is_error` (`analyzer.py:237-239`) returns False for a result
>    carrying neither key, so `:244-246` pre-populates it as a *successfully completed* unit and it is
>    never re-analysed. PR #230 is about report buckets; it does not touch checkpoint adoption.
> 2. **The `"error"` / `"errors"` mismatch.** `analyzer.py:331` defaults to the singular
>    `r.get("finding", r.get("verdict", "error").lower())` while the bucket key is the plural
>    `"errors"`, so that fallback cannot match its own default.
>
> **Relationship to #316.** #316 reaches the same two sinks (`analyzer.py:237-239` and `:330-335`) from
> a different producer — an *unrecognised* finding string, where this issue's producer emits an object
> with **no** `finding` key at all. `_normalize_result`'s branch is gated on
> `if "verdict" not in result and "finding" in result` (`analysis_core.py:40`), so #316's proposed fix
> runs zero code for this input. Different producers, one shared sink:
>
> **A consumer-side fix closes both; a producer-side fix closes neither.** Making `_cp_is_error` fail
> closed on a result carrying neither key — plus bucketing unrecognised/absent verdicts in
> `_count_verdicts` — is the change I would ask for first. Without this note, whoever fixes #316 may
> reasonably believe this issue is covered, or patch the same two functions twice.

gadievron:
## Summary

A Stage-1 verdict the pipeline does not recognize used to vanish silently, three different ways:

- **#316** — `_normalize_result`'s string branch mapped an unrecognized finding string
  (`"maybe exploitable"`) to an upper-cased passthrough verdict (`"MAYBE EXPLOITABLE"`): not a
  counter bucket, not `"ERROR"` (so never retried on resume), not in the vulnerable set (so
  never disclosed). The non-string branch ten lines above already mapped to ERROR with a
  comment naming exactly this failure; the string branch never got the same treatment. The JSON
  corrector had two twin synthesis sites feeding the same escape.
- **#324** — a parsed object with **neither** verdict **nor** finding (a JSON-shaped refusal)
  kept neither key: `_cp_is_error` adopted it as complete on resume, and `_count_verdicts`
  counted it in no bucket (the singular `"error"` default can never match the `"errors"` key),
  so `units_analyzed` overstated.

Both issues reach the same sinks and conclude *"a consumer-side fix closes both; a producer-side
fix closes neither"* — this is the joint fix.

## The fix

**Producers stamp the one error shape** (`verdict="ERROR"`, `finding="error"`, raw value
preserved under `raw_finding` for manual review): an unrecognized finding string, a non-string
finding, a neither-key refusal, and a null/empty verdict (a `{"verdict": null}` refusal is the
#324 shape with the key present). The corrector's two synthesis sites map through their enums
with an ERROR default, and a vuln-mode correction that recovers no recognizable verdict is
reported as a failed correction, not "successful!". The `context_corrector`'s private
`_normalize_result` twin inherits the full treatment.

**Consumers fail closed, once**: a new shared predicate `checkpoint.analyze_result_is_error`
now backs all four copies that had drifted — `analyzer._cp_is_error` (hoisted to module level),
`_count_verdicts`, `checkpoint.load_ids`, `checkpoint.status` — plus the `_summary.json` seed
loop. A legacy neither-key checkpoint is re-analyzed, not adopted; the counter routes it to
`errors` so the partition does not lose the row; `load_ids` keeps its four-phase contract (the
arm is scoped to analyze rows — verify/enhance/dynamic-test checkpoints carry no `result` key).

**The resume arithmetic is exact**: the summary seed counts only what adoption keeps (an
errored row is re-analyzed and owned by `_summary_callback` — seeding it too double-counted;
`completed + errors == total_units` now holds at completion where the pre-existing
`verdict=="ERROR"` over-count did not).

## Design decisions (deliberate, documented)

- A well-formed reply with a garbage verdict now triggers one JSON-correction LLM call — the
  same treatment a parse failure gets: the corrector may legitimately recover a real verdict
  from the same text, and the result is re-gated on ERROR and rejected.
- No in-run retry for these rows (no `error` key): a semantic refusal may be deterministic;
  the retry is resume-owned.
- The new errors are bucketed under the summary's existing `"api"` breakdown label.

## Residuals (pre-existing, pinned by tests, deliberately out of scope)

- An unrecognized **verdict** string (`{"verdict": "SOMETHING_WEIRD"}`) is still dropped by
  `_count_verdicts` — the documented F13 partition gap, unchanged.
- `insufficient_context` is mapped but has no counter bucket (#293 rules it completed — a
  results-partition question, not a coverage loss).
- Both-keys-disagreement rows (`{"verdict": "SAFE", "finding": "garbage"}`) are adopted and
  dropped — not produced by the fixed producers; the live producer of that family
  (`stage1_consistency`'s verdict-only rewrite) is filed separately as #425.
- Legacy pre-fix garbage-verdict checkpoints are still adopted (escape hatch: delete the
  checkpoint dir).

<details>

| check | command | result |
|---|---|---|
| new tests | the issue-316/324 test file (26 tests) | 26 green (RED on pristine: the file cannot even collect — the extracted seed does not exist there) # proc-ok |
| full suite | the full pytest run on the fix branch | 3400 green / 1 failed — that 1 fails byte-identically on pristine master (a macOS `/private/var` tmp artifact, environment-only) |
| mutation smoke | pristine sources restored, the test file re-run | the defect shapes fail; restored: green # proc-ok |
| hermeticity | the env-scrubbed run (fake HOME) | 26 green |
| the real binary | `openant parse <fixture> --output ... --level all`, built from this branch | 2 units extracted, all outputs written — no regression |
| the integrated analyze flow | the real `parse_response` → `_count_verdicts` → the shared predicate, on the two issues' exact reply shapes | the malformed replies land in `errors` (partition closes), are retried on resume, and the checkpoint-completed set excludes them |
| lint | the CI lint scope over the seven touched files | green |
| review | a three-round adversarial wave (e2e + test-validity + bug-hunt lenses) | round 1: 6 findings (the unscoped `load_ids` arm — all three reviewers independently; the null-verdict crash; the seed's fourth predicate copy; the context-corrector partial mirror) — all fixed; round 2: the seed double-count + the unstamped break — fixed; round 3: no findings |

</details>

Fixes #316
Fixes #324

### DIFF
--- a/libs/openant-core/core/analysis_core.py
+++ b/libs/openant-core/core/analysis_core.py
@@ -36,14 +36,24 @@ def _normalize_result(result: dict) -> dict:
     Handles cases where the model returns 'finding' instead of 'verdict',
     or uses different casing/naming conventions.
     """
-    # Normalize finding -> verdict
-    if "verdict" not in result and "finding" in result:
+    # Normalize finding -> verdict. A verdict only counts as present when it
+    # is an EFFECTIVE string — {"verdict": null} is the #324 refusal shape
+    # with the key present (a null verdict crashed _count_verdicts' fallback).
+    verdict = result.get("verdict")
+    has_verdict = isinstance(verdict, str) and verdict.strip() != ""
+    if not has_verdict and "finding" in result:
         finding = result["finding"]
         if not isinstance(finding, str):
             # A non-string finding (list/dict/null/number) is a malformed model reply,
             # not a verdict — map it to ERROR so the error / manual-review accounting
             # counts it, instead of a garbage verdict (e.g. "['VULNERABLE']" from
             # str(finding).upper()) that silently escapes that accounting.
+            # #316: stamp BOTH keys (verdict + finding) — the finding-keyed
+            # consumers (_summary_callback, _count_verdicts) would otherwise
+            # disagree with the verdict-keyed ones. Raw value preserved for
+            # manual review.
+            result["raw_finding"] = finding
+            result["finding"] = "error"
             result["verdict"] = "ERROR"
         else:
             finding_to_verdict = {
@@ -54,7 +64,24 @@ def _normalize_result(result: dict) -> dict:
                 "inconclusive": "INCONCLUSIVE",
                 "insufficient_context": "INSUFFICIENT_CONTEXT",
             }
-            result["verdict"] = finding_to_verdict.get(finding.lower(), finding.upper())
+            verdict = finding_to_verdict.get(finding.lower(), "ERROR")
+            if verdict == "ERROR" and finding.lower() != "error":
+                # #316: an unrecognized finding string is a malformed model
+                # reply, not a verdict. Map it to ERROR — counted in `errors`,
+                # retried on resume, manual-review-visible — instead of the
+                # upper-case passthrough that escaped every accounting (the
+                # non-string branch above chose the same direction).
+                result["raw_finding"] = finding
+                result["finding"] = "error"
+            result["verdict"] = verdict
+    elif not has_verdict:
+        # #324: a parsed object with neither an effective verdict NOR a
+        # finding (a JSON-shaped refusal, e.g. {"reasoning": ...}) is not an
+        # analysis. Stamp the one error shape so it is counted in `errors`
+        # (units_analyzed stops overstating) and re-analyzed on resume
+        # instead of adopted as complete.
+        result["verdict"] = "ERROR"
+        result["finding"] = "error"
 
     # Ensure verdict is uppercase
     if "verdict" in result and isinstance(result["verdict"], str):
--- a/libs/openant-core/core/analyzer.py
+++ b/libs/openant-core/core/analyzer.py
@@ -22,7 +22,7 @@ from pathlib import Path
 
 from core.schemas import AnalyzeResult, AnalysisMetrics, UsageInfo
 from core import tracking
-from core.checkpoint import StepCheckpoint
+from core.checkpoint import StepCheckpoint, analyze_result_is_error
 from core.progress import ProgressReporter
 
 # Import existing analysis machinery
@@ -234,10 +234,7 @@ def _run_detection(units, binding: PhaseBinding, json_corrector, app_context, wo
     code_by_route = {}
     units_to_process = []
 
-    def _cp_is_error(cp_data):
-        res = cp_data.get("result", {}) if cp_data else {}
-        return res.get("verdict") == "ERROR" or res.get("finding") == "error"
-
+    # Hoisted to module level (testable; mirrors checkpoint.py's predicates).
     for i, unit in enumerate(units):
         uid = unit.get("id", f"unit_{i}")
         cp_data = checkpointed.get(uid)
@@ -337,6 +334,54 @@ def _interrupt_report(results, total):
         f"({not_started} not started); progress saved to checkpoints")
 
 
+def _cp_is_error(cp_data):
+    """Is this checkpointed unit an error (must be re-analyzed, not adopted)?
+
+    Delegates to ``checkpoint.analyze_result_is_error`` — the one shared
+    predicate (load_ids / status / the summary seed below all use it; four
+    hand-copies drifted within one PR).
+    """
+    res = cp_data.get("result", {}) if cp_data else {}
+    return analyze_result_is_error(res)
+
+
+def _seed_summary(existing: dict) -> dict:
+    """Seed the _summary.json counters from checkpointed rows.
+
+    Counts as completed ONLY the rows adoption will keep: an errored row
+    (``analyze_result_is_error``) is re-analyzed and its outcome is owned by
+    ``_summary_callback`` — seeding it here (as completed OR as an error)
+    double-counts on resume (completed + errors > total; the pre-existing
+    verdict=="ERROR" over-count, which #316/#324 was extending to the
+    neither-key shape). Usage tokens accumulate over ALL rows — the spend
+    happened regardless of the row's fate.
+
+    Returns: completed, input_tokens, output_tokens, cost_usd,
+    unpriced_models (the #216 marker).
+    """
+    completed = 0
+    input_tokens = 0
+    output_tokens = 0
+    cost_usd = 0.0
+    unpriced: set = set()
+    for _cp in existing.values():
+        if not analyze_result_is_error(_cp.get("result") or {}):
+            completed += 1
+        _usage = _cp.get("usage", {})
+        input_tokens += _usage.get("input_tokens", 0)
+        output_tokens += _usage.get("output_tokens", 0)
+        cost_usd += _usage.get("cost_usd", 0.0)
+        # #216: restore the incomplete-cost marker from per-unit records.
+        unpriced.update(_usage.get("unpriced_models") or [])
+    return {
+        "completed": completed,
+        "input_tokens": input_tokens,
+        "output_tokens": output_tokens,
+        "cost_usd": cost_usd,
+        "unpriced_models": unpriced,
+    }
+
+
 def _count_verdicts(results):
     """Count verdict categories from a results list."""
     counts = {
@@ -353,11 +398,27 @@ def _count_verdicts(results):
         # checkpoint-resume semantics own those units).
         if r is None:
             continue
-        finding = r.get("finding", r.get("verdict", "error").lower())
-        if finding in counts:
-            counts[finding] += 1
-        elif r.get("verdict") == "ERROR":
+        # #324/null-verdict: an INEFFECTIVE finding (absent/None/non-string/
+        # empty) falls back to the verdict, which must itself be an effective
+        # string — the eager ``.lower()`` default crashed post-LLM-spend on
+        # ``{"verdict": null}``.
+        finding = r.get("finding")
+        verdict = r.get("verdict")
+        if not isinstance(finding, str) or not finding.strip():
+            finding = verdict if isinstance(verdict, str) and verdict.strip() else None
+        if finding is None:
+            # Neither an effective finding nor an effective verdict — the
+            # malformed shape (#324; null/empty verdicts): an error, not a
+            # silent drop.
             counts["errors"] += 1
+        elif finding.lower() in counts:
+            counts[finding.lower()] += 1
+        elif verdict == "ERROR" or finding.lower() == "error":
+            # finding=="error" without verdict=="ERROR" (a legacy
+            # half-stamped row): agree with analyze_result_is_error, which
+            # classifies it as an error.
+            counts["errors"] += 1
+        # else: an unrecognized verdict — the documented F13 gap (dropped).
     return counts
 
 
@@ -587,32 +648,20 @@ def run_analysis(
     # Initialize summary tracking for _summary.json
     # Count checkpointed units to seed the counters and sum existing usage
     _existing = checkpoint.load()
-    _summary_completed = 0
-    _summary_errors = 0
     _summary_error_breakdown = {}
-    _summary_input_tokens = 0
-    _summary_output_tokens = 0
-    _summary_cost_usd = 0.0
-    _summary_unpriced: set[str] = set()
     # #293 adjudication: analyze has NO third state — "inconclusive" is a
     # first-class COMPLETED verdict (verdict_taxonomy FINDING_VERDICT_ORDER;
     # the Stage-1 prompt's own enum), not a degenerate no-verdict marker like
     # verify's verification.incomplete or enhance's INCOMPLETE_CLASSIFICATION.
     # The third bucket stays 0 here but is still emitted for shape consistency.
     _summary_incomplete = 0
-    for _uid, _cp in _existing.items():
-        _r = _cp.get("result", {})
-        if _r.get("verdict") == "ERROR" or _r.get("finding") == "error":
-            _summary_errors += 1
-            _summary_error_breakdown["api"] = _summary_error_breakdown.get("api", 0) + 1
-        else:
-            _summary_completed += 1
-        _cp_usage = _cp.get("usage", {})
-        _summary_input_tokens += _cp_usage.get("input_tokens", 0)
-        _summary_output_tokens += _cp_usage.get("output_tokens", 0)
-        _summary_cost_usd += _cp_usage.get("cost_usd", 0.0)
-        # #216: restore the incomplete-cost marker from per-unit records.
-        _summary_unpriced.update(_cp_usage.get("unpriced_models") or [])
+    _seed = _seed_summary(_existing)
+    _summary_completed = _seed["completed"]
+    _summary_errors = 0  # errored rows are re-analyzed; _summary_callback owns them
+    _summary_input_tokens = _seed["input_tokens"]
+    _summary_output_tokens = _seed["output_tokens"]
+    _summary_cost_usd = _seed["cost_usd"]
+    _summary_unpriced: set[str] = _seed["unpriced_models"]
 
     def _usage_dict():
         usage = {"input_tokens": _summary_input_tokens,
--- a/libs/openant-core/core/checkpoint.py
+++ b/libs/openant-core/core/checkpoint.py
@@ -41,6 +41,43 @@ SUMMARY_FILE = "_summary.json"
 _RESERVED_FILES = frozenset({SUMMARY_FILE, FINGERPRINT_FILE})
 
 
+def analyze_result_is_error(res) -> bool:
+    """Is an analyze-style ``result`` an error (retried, never adopted as complete)?
+
+    The single shared predicate for every resume-facing consumer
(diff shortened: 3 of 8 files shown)


gadievron:
Panel review finding (full analysis on the issue): the unrecognized **verdict**-key case still fails open at every accounting sink (dropped from _count_verdicts, adopted-complete on resume, counted completed, never disclosed) — the exact failure mode #316 fixed for the finding key. Your comments document it as the deliberate F13 gap, which is honest — but the PR title ('unrecognized or absent Stage-1 verdict routes to the error accounting') is literal only for findings + absent/null/empty verdicts. Filed as issue for the close-or-retitle decision: the close is a small enum whitelist in the passthrough branch. Also: json_corrector's non-string branch stamps only verdict=ERROR (no finding rewrite / raw preservation) — the third divergent copy of the 'one error shape'; and three unused function-local imports survive at test :287-289 (json as _json, write_json, os).

Issue #431, opened 2026-08-30
python.Invoke reads its subprocess pipes manually — a descendant-held pipe wedges the CLI until the 30m deadline, and envelope recovery waits for the deadline even when the run finished long before it


gadievron:
## Summary

`python.Invoke` reads the subprocess's stdout/stderr through `StdoutPipe`/`StderrPipe` +
its own `io.Copy`/`streamStderr` readers. That shape has two structural costs the sibling
implementation (`invoke_ctx.go`, the web-UI path) already diagnosed and solved with managed
writers + `WaitDelay` — quoted from its own comment (`invoke_ctx.go:100-104`):

> Managed line-writer for stderr instead of StderrPipe: os/exec runs the copy goroutine, so
> WaitDelay can force-close the pipe if a killed **OR a naturally exited** process leaves a
> detached child holding the write end. A StderrPipe scanner read before Wait() would instead
> deadlock there — the read never sees EOF, so Wait (which starts WaitDelay) is never reached.

Both costs are live for `Invoke`:

1. **The descendant-held-pipe wedge (pre-existing at master).** The engine spawns subprocesses
   that inherit the Go-side pipes — e.g. `core/reporter.py:762` / `:799` and
   `core/parser_adapter.py:715` / `:847` run `subprocess.run(..., stdout=sys.stderr,
   stderr=sys.stderr, ...)`. If such a descendant outlives the Python parent while holding only
   the stderr write-end, `Invoke`'s normal path (`<-stderrDone` before `cmd.Wait()`, the
   ordering at master) blocks until the **30-minute invoke deadline** — `Wait`, the only thing
   that would close the read-end and start `WaitDelay`, is never reached. A scan that finishes
   in two minutes stalls for thirty, then surfaces the deadline note for what was only a held
   pipe. (`waitdelay_test.go:20-54` is the executed repro of this shape for the sibling path.)
2. **Recovery waits for the deadline even when the run finished long before it.** With the
   pipes read manually, a held stdout write-end delays the envelope recovery until the
   watchdog closes the read-end at the deadline — a run that completed in 40 seconds is
   returned 30 minutes late.

The fix pattern exists in-repo (`invoke_ctx.go:18-53`, `lineWriter` + a captured stdout
buffer): route `cmd.Stdout`/`cmd.Stderr` through managed writers so os/exec's own copy
goroutines are drained by `Wait` and bounded by `WaitDelay` (5s). That simultaneously removes
[...]

Issue #313, opened 2026-08-21
Ctrl+C during analyze raises `AttributeError` on the `None` placeholders, and the resulting error envelope makes the CLI record an interrupt as a failed scan


gadievron:
## Summary

`Ctrl+C` during Stage-1 analyze is caught and handled — pending work is cancelled and checkpoints are
reported saved — and then the run raises `AttributeError`, because the results list still holds
`None` placeholders for units that never ran.

The consequence is not only a traceback. The error envelope that Python emits fills stdout, and the
Go CLI's interrupt short-circuit fires **only** on empty stdout. So the interrupt is never recognised
as one: the scan is recorded as **failed** and exits 2, rather than interrupted with 130.

Work is not lost — checkpoints are written before the failure.

## The placeholders

`libs/openant-core/core/analyzer.py:233`:

```python
    results = [None] * total
```

Each slot is replaced as its unit completes. On a clean run every slot is filled.

## Both interrupt handlers return the partially-filled list

Parallel path, `:306-317`:

```python
    except KeyboardInterrupt:
        print("[Detect] Interrupted — cancelling pending work...",
              file=sys.stderr, flush=True)
        executor.shutdown(wait=False, cancel_futures=True)
        print("[Detect] Progress saved to checkpoints",
              file=sys.stderr, flush=True)
    else:
        executor.shutdown(wait=False)

    progress.finish()

    return results, code_by_route
```

Sequential path, `:279-283`, has the same shape:

```python
        except KeyboardInterrupt:
            print("[Detect] Interrupted — progress saved to checkpoints",
                  file=sys.stderr, flush=True)
        progress.finish()
        return results, code_by_route
```

Neither re-raises. Every unit that had not completed is still `None`.

## The consumer does not tolerate `None`

`libs/openant-core/core/analyzer.py:330-331`:

```python
    for r in results:
        finding = r.get("finding", r.get("verdict", "error").lower())
```

Executed at `b5019628`, from the repository root with `libs/openant-core` on `sys.path`:

```
$ python3 -c "import sys; sys.path.insert(0,'libs/openant-core'); \
from core.analyzer import _count_verdicts; _count_verdicts([{'finding':'safe'}, None])"
Traceback (most recent call last):
  ...
  File ".../libs/openant-core/core/analyzer.py", line 331, in _count_verdicts
    finding = r.get("finding", r.get("verdict", "error").lower())
AttributeError: 'NoneType' object has no attribute 'get'
```

## The path to the user

[...]

Issue #161, opened 2026-07-15
Enhance is failing with local Ollama Model after x entries


jblu42:
I tried OpenAnt with the latest changes to possibly use a local model. I use Ollama with a Quen3.5 Model with the OpenAI compatible endpoint manually configuring the models in config.json.

During Enhance, after about 20-40 entries it stops with an error:

Error: failed to read stdout: read |0: file already closed

Any idea what could be the issue or how to debug this?

jblu42:
Update: I had Claude Fable analyze this issue and here is the bug report: 

The error wasn't coming from ollama at all — it came from the Go CLI's subprocess watchdog. python.Invoke ([apps/openant-cli/internal/python/invoke.go](https://github.com/jblu42/openant/blob/claude/ollama-stdout-error-enhance-h15o2i/apps/openant-cli/internal/python/invoke.go)) wrapped every Python subprocess in an absolute 30-minute wall-clock timeout. Agentic enhance against a local ollama model runs roughly a minute per unit, so after ~20–40 entries the deadline fired, the CLI killed the Python process and closed its pipes, and the in-flight stdout read surfaced as the cryptic failed to read stdout: read |0: file already closed. That's why the failure point varied between 20 and 40 entries — it was always exactly 30 minutes of elapsed time, not a particular entry count.

gadievron:
Thanks for the report — it's been triaged into the queue (deferred under the current fixes-only scope, but tracked).

gadievron:
Hi @jblu42 — your diagnosis was **correct**, and this fix makes the error say so at the point of failure.

**What you hit:** enhance against a local Ollama model runs ~1 min/unit, so the 30-minute invoke deadline (the Go CLI's subprocess watchdog in `python.Invoke`) fires after ~20-40 entries — your elapsed-time observation was the key discriminator; it was always exactly 30 minutes, not a particular entry count.

**What the error said:** `Error: failed to read stdout: read |0: file already closed` — the watchdog killing the process and closing the pipes, with nothing naming the deadline or how to proceed.

**What it says now** (PR #413, CI green): when the deadline fires, the error names the invoke deadline with its effective value, the `OPENANT_INVOKE_TIMEOUT` override (PR #237, already merged — e.g. `OPENANT_INVOKE_TIMEOUT=2h` for a slow local model: ~1 min/entry × your corpus size, with headroom), and the checkpoint-resume path (completed units are checkpointed; a re-run resumes from them).

[...]

PR #433, opened 2026-08-30
fix(cli): a complete result envelope wins over the invoke deadline on every kill surface — and the diagnosis stops claiming kills that did not happen (#161 follow-up)

gadievron:
## Summary

A follow-up hardening of #413 (Refs #161 — the issue stays open for the reporter). A
7-round adversarial review of the merged fix — with every load-bearing finding verified by
execution before fixing, and every test proven RED against the pre-fix code — found the
diagnosis was gated on **how the process died** instead of **whether a usable result exists**:

- A child that wrote a COMPLETE result envelope — with a descendant holding a pipe write-end
  past the deadline — had the envelope **discarded** and was told "killed mid-run and its
  output discarded" while it sat in the buffer (two executed shapes).
- The exit-0 window (`Wait` returning the context error, not an `ExitError`) and the Windows
  kill (`TerminateProcess` yields exit code 1, never negative) never saw the diagnosis at all.
- The checkpoint-resume hint was false for 7 of 12 subcommands — including `parse`, the very
  path (a hung parser) the deadline exists for — and a test pinned the false case.
- A natural crash (traceback, exit 1) with a held pipe was diagnosed as a deadline kill —
  "raise the budget" advice for a crash, masking the real exit status (executed).

## The fix

- **Envelope-wins uniformly** (the #313 principle): the stdout buffer is parsed first on every
  path; a complete envelope is returned with a quiet-aware stderr note — a recovered envelope
  is a **completed run** (the Python CLI emits its envelope as the run's final act; SIGKILL
  mid-`json.dump` leaves unparseable truncation, and valid JSON that is not an envelope —
  `null`, `{}` — is rejected by a status-field guard).
- **The killed-gate**: the diagnosis fires only when NO usable envelope exists AND the child
  was actually killed — a signal death on Unix; on Windows (where a kill's exit 1 is
  indistinguishable from a natural exit 1) the fired deadline is accepted as the kill signal,
  a documented trade-off; the exit-0 window is classified via `ProcessState` (os/exec
  substitutes the context error only for an already-successfully-exited child), so a natural
  exit keeps its own failure.
- **Artifact suppression**: the watchdog's read-end close (`os.ErrClosed`) and the
  context-substitution wait error, under a fired deadline, never surface — the cryptic
  `read |0: file already closed` headlines nothing, and the diagnosis wraps the honest
  `signal: killed`.
- **Interrupt precedence**: a Ctrl+C inside the deadline's final seconds wins —
  interrupted/130, not failed/2 (pinned by a real self-SIGINT test).
- **The hint, gated and conditional**: only the checkpointing subcommands
  (`scan`/`enhance`/`analyze`/`verify`/`dynamic-test`, table-tested over all 12), worded
  "under the run's output dir (a re-run with the same `--output` or project resumes)" —
  an ad-hoc run's checkpoints land in a fresh `mkdtemp` dir a re-run will not see.
- **Ordering**: stderr is drained before `Wait` on the normal path (StderrPipe's contract —
  Wait-first drops the child's final stderr lines) and after on the read-error path.

## Residuals (documented in code and tests)

- On Windows a kill's exit-code artifact is indistinguishable from a legitimate exit 1.
- A genuine wait failure under a fired deadline reads as the kill (the cause is still wrapped).
- A recovered envelope from a signal-killed child keeps `normalizeExit`'s conservative exit 2
  — a completed result, an anomalous invocation (pinned as the contract).
- `invoke_ctx.go` — the web-UI sibling — has its own envelope-discard-on-cancellation gap;
  both it and the descendant-held-pipe wedge (a pre-existing 30-minute stall shape) are filed
  as #431 with the `lineWriter`/`WaitDelay` refactor that resolves them.

<details>

| check | command | result |
|---|---|---|
| new tests | the invoke envelope/round2/timeout test files (15 deadline-dependent tests) | all green # proc-ok |
| full suite | `campaign/bin/go-suite.sh` (the wrong-workdir-proof wrapper) | 10 ok / 0 fail, five consecutive runs (three stability + commit + pre-ship) |
| RED proof | each key test against the pre-fix `invoke.go` (worktrees at 60c4c40 / e9ed4c0) | every one fails for its named defect; the recovery ×3, natural-crash, exit-0-window, interrupt, hint-gating, warn-once, wrapped-cause |
| the race, fixed at source | the deadline-dependent tests were flaky under the full suite (child-start lag past 2s and 6s budgets under the suite's spawn storms; CPU hogs alone do not reproduce it) — budgets sized to 30s, the mechanism documented at `writeScript` | four consecutive green full-suite runs; the standalone package 225s # proc-ok |
| lint | `gofmt -l` empty; `go vet ./...` clean | green # proc-ok |

The review loop that shaped it: 7 adversarial rounds (3-persona → 2-axis → Windows/test-axis →
a lost-fixes incident caught because the commit message claimed them → confirmation →
deep-refute → final polish), including two rounds that found defects in earlier rounds' fixes
(a stderr-tail truncation regression; a vacuous "strengthened" test) — each fix re-verified by
execution.

</details>

Refs #161


gadievron:
Panel verdicts applied (fable: APPROVE with 4 cosmetic nits; sonnet: FIX with 3 items, adjudicated from source and live-probed):

- Sonnet item 1 (kill-artifact exit code): REAL, fixed in 72f751f..+1 — a killed Windows run with a recovered envelope surfaced exit 1 (the TerminateProcess artifact, same code as a legitimate vulns-found exit). exitErrIsKillArtifact now routes every platform to the conservative 2.
- Sonnet item 2 (stderr-drain bound): REAL, confirmed by an executed probe before fixing — a child that printed a complete envelope and exited in 0.1s hung the invocation 15.77s of a 15s budget (a stderr-only descendant kept the drain blocked until the deadline watchdog); post-fix probe 5.40s, clean return, no spurious recovery note.
- Sonnet item 3 (envelope finality contract test): not applied — fable verified from the Python source that _output_json is the single writer and each invocation emits exactly one envelope as its final act (incl. the analyze --verify chain), so the assumption the item guards is currently true; noted as optional hardening on the Python side, out of this Go-side PR's scope.

Issue #215, opened 2026-08-03
Confirmed findings carry almost no structured metadata, and no severity field exists


NahumKorda:
## Summary

Of the findings upheld by Stage 2 in our run, fewer than one in ten carried a CWE classification, an impact statement, or reproduction steps. None carried a suggested fix. There is no severity field in the output schema at all.

## Evidence

Across the 82 Stage 2-upheld findings in `pipeline_output.json`:

| Field populated | Count |
|---|---|
| `description` | 82 |
| `cwe_id` (non-zero) | **6** |
| `impact` | **6** |
| `steps_to_reproduce` | **6** |
| `suggested_fix` | **0** |
| any severity field | **0 — field does not exist** |

The other 76 arrive with `cwe_id: 0`, `name: "vulnerable"`, `short_name: "VULNERABLE"`, and prose only. The six well-formed findings are richly detailed — CWE, concrete attack path, reproduction — which suggests an extraction or schema-adherence gap rather than a design limit.

## Why it matters

Without severity or CWE, findings cannot be ranked, filtered, deduplicated across runs, or mapped to a remediation SLA — and 76 of 82 are mutually indistinguishable in any structured view, forcing reviewers to read prose serially. `suggested_fix` exists in the schema and was populated zero times.

## Suggested fix

Constrain the Stage 1 and Stage 2 outputs with a schema requiring `cwe_id`, `severity` and `impact`, and treat a response missing them as a parse failure to retry rather than a finding to emit. Introducing a `severity` field at all would be the single highest-value change for downstream triage.

---

*Observed on one full-pipeline run at production scale: a private TypeScript monorepo (Angular front end, NestJS back end), 1,351 analysis units from 866 source files, run with `--verify` and no `--limit`, ~6.5 hours wall clock. The engine was invoked directly (`python -m openant scan …`, Python 3.13); the Go CLI was not built, so none of this involves the Go↔Python envelope. Stage 1 detection and Stage 2 verification ran on **Claude Opus 5**; application context, enhancement and reporting on **Claude Sonnet 5** — all seven phases routed through **OpenRouter** (`anthropic/claude-opus-5`, `anthropic/claude-sonnet-5`) via a locally-added `openrouter` provider adapter, on a build based on upstream `2ed78f6`. Cost figures are expressed as proportions of the run total; absolute amounts are omitted deliberately, as is the identity of the scanned repository.*

gadievron:
[...]

PR #436, opened 2026-08-30
feat(report): every finding carries a rankable severity — model-supplied with a conservative derived fallback, threaded to every deliverable surface (#215)

gadievron:
## Summary

The reporter's population: **82 Stage-2-upheld findings, 6 with cwe_id/impact/steps, 0 with any
severity** — "introducing a severity field at all would be the single highest-value change for
downstream triage." Traced to a schema mismatch re-derived first-hand: the finding record's
rich fields read from `finding["vulnerabilities"][0]` — an array **no prompt asks for** (the
Stage-1 contract is flat); its only producer is the JSON corrector's legacy `_VULN_SCHEMA`,
which carries a 4-level severity enum nested where nothing downstream reads it. The reporter's
well-formed 6 are most plausibly the corrector-repaired subset (checkable on their data via
the `json_corrected` transit field from #366).

The reporter's GOAL — rank/filter/SLA — is what this delivers; their hard-retry MEANS is
deliberately not adopted (a model that never emits the field would lose its findings to
ERROR — the false-negative direction).

## The fix (through 3 adversarial review rounds + a deep-refute, every finding verified and every test RED-proven)

- **The Stage-1 prompt asks for severity** (`critical`/`high`/`medium`/`low` — the enum the
  corrector's legacy schema already used); the corrector schema carries it top-level.
- **FINDING-ONLY**: only vulnerable/bypassable rows carry severity. Safe/protected/
  inconclusive/ERROR rows carry none, everywhere — a `severity=low` filter must not return
  mostly non-findings.
- **The producer** (`_normalize_result`): a model-supplied enum is kept
  (`severity_source: model`); anything else derives conservatively (`vulnerable→high`,
  `bypassable→medium`, never a derived critical). Severity never produces the error shape and
  never touches the error key — the #426 accounting is undisturbed.
- **The read-time contract** (unified and pinned at all three sites): ONLY a model value
  keeps its rank; JSON-repaired (`corrected`) and stale `derived` stamps re-derive from the
  **final** verdict — Stage 2 reclassifies rows, and a Stage-1 stamp must not outrank it.
  The "Max iterations reached" downgrade strips severity (no CRITICAL badge on an unverified
  row). OLD artifacts (pre-PR scans) get read-time derivation, so every scan ranks.
- **The surfaces**: the `pipeline_output.json` finding record + `severity_source`; the CSV's
  `severity` + `severity_source` columns; the summary-table column (the summary LLM now
  receives the values); HTML badges on both templates (with a severity-source tooltip — a
  derived HIGH must not read as an assessment); SARIF.
- `report/schema.py`: Optional on Finding — old artifacts validate.

## Known limits (disclosed)

- **GitHub Code Scanning shows ONE severity per verdict**: the SARIF `security-severity`
  lives on the rule (per-verdict, its findings' max — stable rule ids, so a severity flip
  never re-keys alerts and discards triage state). Per-finding severity reaches SARIF
  consumers via `result.properties` (`severity`, `severity_source`), which GitHub's UI does
  not render. A batch of 9 low + 1 critical vulnerable findings shows as Critical in
  GitHub's UI.
- On old artifacts (the reporter's own 82), read-time derivation yields 82/82 `derived` —
  the model-supplied values land on scans made with this prompt.
- The prompt change invalidates in-flight checkpoint adoption via the analyze fingerprint
  (I2, by design): a partially complete pre-upgrade scan resumed after upgrade re-runs from
  zero.
- Stage-2 verify does not yet revise severity — a follow-up direction if the reporter
  confirms this shape.

<details>

| check | command | result |
|---|---|---|
| new tests | the issue-215 test file (23 tests) + 4 Go tests | green; RED-proven at each round (10 on pristine master, 6 on round-1, 6 on round-2 — via worktrees) # proc-ok |
| the corrected-adoption behavior | the real `analyze_unit` driven with a stubbed adapter/corrector | valid extraction → `corrected`; garbage → `derived` |
| full suites | the Python suite + the Go suite wrapper | 3432 green + the one byte-identical-on-pristine macOS tmp failure · the Go side 10 ok / 0 fail | # proc-ok
| review | 3 adversarial wave rounds + the deep-refute; two rounds found real defects in the round before (the stale-derived escape, the downgrade leak, a rule-fanout revert) | convergence; the deep-refute could not refute the composition chain |

</details>

Refs #215 — the issue stays open for the reporter's confirmation (the schema-mismatch
reading of their 6 well-formed findings is a hypothesis they can check on their data via the
`json_corrected` transit field)

### DIFF
--- a/apps/openant-cli/internal/report/report_test.go
+++ b/apps/openant-cli/internal/report/report_test.go
@@ -121,3 +121,35 @@ func excerpt(out, key string) string {
 	}
 	return out[start:end]
 }
+
+// #215: a severity-bearing finding renders its badge; a finding without one
+// (old artifacts) renders none.
+func TestRenderReskinSeverityBadge(t *testing.T) {
+	data := baseReportData()
+	// the reskin template iterates the GROUPED structure the report-data
+	// projection builds (cli.py: findings_by_verdict) — not .Findings.
+	data.FindingsByVerdict = []FindingGroup{{
+		Verdict: "vulnerable",
+		Findings: []Finding{{
+			Number:         1,
+			Verdict:        "vulnerable",
+			File:           "src/app.py",
+			Function:       "render",
+			Severity:       "critical",
+			SeveritySource: "model",
+		}},
+	}}
+	var buf bytes.Buffer
+	if err := RenderReskin(data, &buf); err != nil {
+		t.Fatalf("RenderReskin: %v", err)
+	}
+	out := buf.String()
+	if !strings.Contains(out, ">critical<") {
+		t.Error("expected the severity badge to render for a severity-bearing finding")
+	}
+	// the color method backs the badge style
+	f := Finding{Severity: "critical"}
+	if f.SeverityColor() != "#721c24" {
+		t.Errorf("SeverityColor(critical) = %q, want #721c24", f.SeverityColor())
+	}
+}
--- a/apps/openant-cli/internal/report/sarif.go
+++ b/apps/openant-cli/internal/report/sarif.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 )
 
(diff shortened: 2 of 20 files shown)


gadievron:
Panel verdicts applied (fable: APPROVE, 8 attack points verified from source, 2 non-blocking warts noted; sonnet: FIX x3 — item 1 real, items 2-3 doc-only; all applied):

- Sonnet item 1 (REAL, confirmed live on the verify-out artifact before fixing): report-data downgraded a 'Max iterations reached' verification to inconclusive and stripped its severity, while the reporter and CSV sites gated on the raw preserved vulnerable verdict — the same row ranked in pipeline/CSV while the HTML/SARIF displayed it as unrankable. Fixed by factoring the downgrade into severity_display_verdict (verdict_taxonomy) and gating all three sites on it; post-fix probes on the real artifact show the rows stripped uniformly. 3 regression tests added (26 total in the file).
- Sonnet items 2-3 (doc-only): the model-exemption rationale and the old-artifact vs resumed-checkpoint coverage distinction — comments added at the named sites.
- Fable's two warts (raw results_verified keeps a stale severity on downgraded rows — gated surfaces are the contract; the corrected-label wart) — test-pinned by the PR's own tests, noted, not changed.

Gate: Run A 3438 passed / 0 failed (incl. the 23-test PR suite); CI green pre-push; live arms: model severity stamped on a live vulnerable finding (severity=high, source=model), read-time derivation proven on a pre-#215 artifact, finding-only gate proven (safe + Stage-2-downgraded rows carry none), Go SARIF rule security-severity 7.5 for high with one rule per verdict.

Issue #317, opened 2026-08-21
`safe_filename` preserves case, so two units differing only in case share one checkpoint file on a case-insensitive filesystem


gadievron:
## Summary

`safe_filename` preserves case, so two unit IDs differing only in case produce two distinct filename
*strings* — but the same file on a case-insensitive filesystem (macOS default, and Windows). The
second checkpoint overwrites the first. The lost unit is re-analysed on every resume, because its
checkpoint is overwritten again each time.

## The sanitiser preserves case

`libs/openant-core/utilities/safe_filename.py:12-16` (the sanitiser):

```python
    safe = (unit_id
            .replace("/", "__")
            .replace("\\", "__")
            .replace(":", "_")
            .replace(" ", "_"))
```

The hash suffix that would disambiguate is applied only on the long-name branch, `:21-23`:

```python
    if len(safe) > max_len:
        h = hashlib.sha256(unit_id.encode()).hexdigest()[:16]
        safe = safe[:max_len] + "_" + h
```

So for ordinary-length IDs there is no uniqueness mechanism at all. The docstring is accurate about
this — it scopes the hash to *long* filenames — so the gap is in the design, not in a mismatch
between code and comment.

## Executed against the real checkpoint path

Using the shipped `StepCheckpoint.save()` / `.load()` at `b5019628`, on macOS:

```
saved 2 units: 'pkg/Foo.py:run' and 'pkg/foo.py:run'
checkpoint files on disk: 1 -> ['pkg__Foo.py_run.json']
load() returns 1 unit(s): ['pkg/foo.py:run']

=> unit 'pkg/Foo.py:run' is LOST
```

Note the surviving file is named with the **first** ID's casing while holding the **second** ID's
data, so the `id` recorded inside no longer matches the filename.

The underlying filesystem behaviour, for completeness:

```
$ printf 'A' > Foo.json ; printf 'B' > foo.json ; ls
Foo.json
$ cat Foo.json
B
```

## Consequences

1. **One unit's analysis is silently lost.** Nothing errors; the second save simply replaces the
   first, and `load()` reports one unit where two were saved.
2. **A resume never converges for the pair.** The unit whose checkpoint was overwritten is absent
   from `load()`, so it is re-analysed on the next run — and overwritten again. It is re-paid on
   every resume.

## Exposure — honest bounds

> ⚠️ **SUPERSEDED — read this before the section below.** The "no observed instance" bounds here were
[...]

PR #438, opened 2026-08-31
fix(checkpoint): case-only unit IDs no longer silently lose checkpoints — save-time disambiguation + id-keyed restore (#317)

gadievron:
## Summary

`safe_filename` preserves case, so two unit IDs differing only in case (`pkg/Foo.py:run` vs `pkg/foo.py:run` — 593 real pairs in ordinary Go dependencies via the exported/unexported convention) computed the **same** filename. On a case-insensitive filesystem (macOS default, Windows) the second checkpoint save silently overwrote the first: one unit's analysis lost, and — `load()` reporting one unit where two were saved — re-analyzed and re-**paid** on every resume. The run never converged for the pair. The issue's executed repro:

```
saved 2 units: 'pkg/Foo.py:run' and 'pkg/foo.py:run'
checkpoint files on disk: 1
```

The issue also pre-analyzed the migration hazard that shaped the fix: completion **detection** is content-based (listdir + each file's id) but **restoration** was filename-computed — any naming change alone strands pre-existing checkpoints into degraded analysis (empty contexts, fewer findings — the false-negative direction).

## The fix

- **Save disambiguates on collision**: the bare name when free or holding the unit's own data; a different unit's checkpoint there → the case-sensitive 16-hex hash suffix (the existing long-name mechanism, capped so name+`.json` fits 255). **Home stability**: once a unit lives at its suffixed home, later saves keep writing there — even when the bare name frees up (the sibling's file deleted, the dir moved to a case-sensitive fs) and **including the corrupt-file branch** — otherwise the fresh result orphans the home and two same-id files resolve by arbitrary listdir order, resurrecting the non-convergence. A corrupt/foreign `.json` reads as no-id (overwrite the garbage) instead of crashing every later save of that unit.
- **The >233 truncated regime is injective by construction** (the name already carries the full id's hash) — an explicit early-out states the invariant rather than leaving it to the 233-constant coupling between two files.
- **A module-level lock serializes resolve+write** — saves run in ThreadPoolExecutor workers (the enhancer *and* the analyzer); the TOCTOU (two siblings in flight both writing the bare name) is pinned deterministically (two threads rendezvous inside the hooked exists check; proven by mutation: the lock deleted → red in 0.07s, restored → green in 2.03s).
- **Restore is id-keyed everywhere**: the enhancer's three filename-computed restore sites (agent_context, llm_context, the usage summary) now look up by the id *inside* the file (one listdir+read pass). This fixes the migration stranding (old-scheme files restore by content), the B-restores-A hazard, and the detection/restore asymmetry itself — and makes every future naming change safe. Behaviorally pinned on **both** paths (the single-shot loop and `enhance_dataset_agentic`).
- **Bonus, disclosed**: the sanitizer aliases `:` and ` ` both to `_` — `a:b:c` vs `a b:c` produce byte-identical names on **every** filesystem. The exists-with-different-id check cures that class too (pinned as the fs-independent guard — the load-bearing test on Linux where the case-only repro cannot fire).
- On case-**sensitive** filesystems the exists-with-different-id check never fires for a case-only pair: two real files, the naming scheme byte-identical — zero behavior change there.

## Residuals (documented)

- A third unit occupying a unit's hash-suffixed home requires a ~2⁻⁶⁴ prefix collision or hand-forging — out of scope.
- **Downgrade note**: rolling back this change strands suffixed-home files the same way the issue's migration hazard describes (detected complete, restore no-ops). Inherent to any forward migration here.

<details>

| check | command | result |
|---|---|---|
| new tests | the issue-317 test file (16 tests) | all green; RED 5 on pristine via stash; the lock pin red in 0.07s with the lock mutated out, green 2.03s restored # proc-ok |
| full suite | the full pytest run on the fix branch | 3425 green + the one byte-identical-on-pristine macOS tmp failure |
| review | need-check ("ship the design — strictly dominates the issue's options") + 3 adversarial wave rounds + deep-refute | round 1: the orphaned twin + the dead-by-accident coupling + a leaky test; round 2 caught round 1's fixes over-claiming (a lock test with no threads, the corrupt-branch bypass, one-of-three sites covered); round 3: no functional findings; deep-refute: no refutation — the branch matrix, lock, and a full consumer audit held |
| hermeticity + lint | the env-scrubbed run; ruff over the touched files | green # proc-ok |

</details>

Refs #317

### DIFF
--- a/libs/openant-core/core/checkpoint.py
+++ b/libs/openant-core/core/checkpoint.py
@@ -20,9 +20,11 @@ Usage:
     cp.cleanup()                          # remove dir on success
 """
 
+import hashlib
 import json
 import os
 import shutil
+import threading
 import sys
 from datetime import datetime, timezone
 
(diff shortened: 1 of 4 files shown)


gadievron:
Panel verdicts applied (fable: APPROVE — 8 attack points verified from source incl. direct computation of the sanitizer-alias collision mechanism and code-unreachability of the duplicate-id shapes; sonnet: FIX x3 — all applied):

- Item 1 (self-heal): the bare-holds-mine branch now consolidates a dual-stale-file state (bare + suffixed home both holding the id) to the NEWER file, deleting the superseded twin — the id-keyed restore's last-wins can no longer serve the stale copy. Regression test added (17 in the file).
- Item 2 (scope): _save_lock documented as SINGLE-PROCESS — concurrent CLI invocations on one checkpoint dir are an unsupported shape not serialized by a threading lock.
- Item 3 (coupling): the 233 literal is now SAFE_FILENAME_MAX_LEN exported from safe_filename.py and imported by the disambiguator.

Gate: Run A 3429 passed / 0 failed (incl. the 17-test file); CI green pre-push.

Issue #440, opened 2026-08-31
parser: typed-receiver base resolution is same-file-anchored and FIFO — imported in-repo bases and the MRO on diamonds stay unresolved (the #318 residuals)


gadievron:
## Summary

Two FN residuals from the #318 fix (the same-file base walk for typed receivers), filed
separately per that PR's triage:

1. **Imported in-repo bases stay unresolved for typed receivers.** The walk is anchored to
   the CALLER's file, so `a.py: class Base` imported into `b.py` (`class Sub(Base)`) never
   resolves `s = Sub(); s.inherited()` — while `_resolve_super_call` walks the SAME class
   cross-file and the import map (`self.imports['b.py']['Base'] == 'a.Base'`) already
   identifies it. This is the most common real layout (a base module + subclasses
   elsewhere) and likely the bulk of the #318 census's remaining gap. The walk should
   follow the import map (and the cross-file step-3-identified class's own file) — the C
   "same-file" floor doesn't transfer to a language with an import map the resolver
   already holds. Related: `PARSER_UPGRADE_PLAN.md:170` (cross-file class inheritance).
2. **The FIFO walk is not the C3 MRO — a diamond picks a single genuine-but-wrong
   ancestor.** `class C(A, B)` with `A(X)`, `X.m` and `B.m`: the FIFO picks `B.m`; Python
   runs `X.m` — so `X.m` keeps an empty caller set and is pruned (the FN direction
   #318 was filed to remove). Shared with `_resolve_self_call`, so the fix unifies both
   walks: compute C3 (the extractor's base lists suffice) or, per the over-seed rule,
   emit every first-hit-per-branch candidate when multiple base branches define the
   method (a reachability-safe FP beats the FN).

## What I am not claiming
- Not blocking #318's merge: both are disclosed there as the FN floor of the declared
  same-file scope.
- The subscripted-base case (`Sub(Base[int])`) is FIXED in #318's PR; not re-asked here.

gadievron:
Fixed in PR #474 (CI green, all 14 checks; leave-for-review — not merged). Closing per the campaign protocol; the PR body carries the full root-cause + evidence.

Issue #318, opened 2026-08-21
Inherited methods resolve through `self`/`super()` but not through a typed receiver or the class name, so same-file inherited calls emit no edge


gadievron:
## Summary

The Python call-graph builder resolves an inherited method when the receiver is `self` or `super()`,
and drops the same call when the receiver is a locally-constructed instance or the class name. The
base-class walk exists, the base data is recorded, and the receiver's type is already inferred — the
typed-receiver lookup simply never consults the bases.

Both cases fail **same-file**, with no cross-file resolution involved.

## Executed — same inherited target, five receiver shapes

One file, one class hierarchy, real `parse_repository()` at `b5019628`:

```python
class Base:
    def inherited_method(self): return 1
    @classmethod
    def inherited_cm(cls): return 2

class Sub(Base):
    def own_method(self): return 3
    def via_self(self):   return self.inherited_method()
    def via_super(self):  return super().inherited_method()
    def via_self_cm(self): return self.inherited_cm()

def via_local_instance():
    s = Sub()
    return s.inherited_method()      # M1

def via_class_name():
    return Sub.inherited_cm()        # M2

def control_own():
    s = Sub()
    return s.own_method()            # control: same receiver shape, non-inherited
```

```
self.inherited_method()          -> ['app.py:Base.inherited_method']   RESOLVES
super().inherited_method()       -> ['app.py:Base.inherited_method']   RESOLVES
self.inherited_cm()              -> ['app.py:Base.inherited_cm']       RESOLVES
s = Sub(); s.inherited_method()  -> NO EDGE                            M1
Sub.inherited_cm()               -> NO EDGE                            M2
s = Sub(); s.own_method()        -> ['app.py:Sub.own_method']          control
```

The control isolates the variable: same receiver construction, same call syntax — the only difference
is whether the method is declared on `Sub` or inherited from `Base`.

## Where the lookup stops

- **M1** — `_resolve_class_method` (`parsers/python/call_graph_builder.py:604-626`) never reads
  `class_data['bases']`. It stops at `:614`, a flat scan of the class's declared methods.
- **M2** — never reaches that function. `:488` routes to `_resolve_module_call`, which scans only the
  named class's declared methods (`:645-648`) and then `return None` (`:650`).
- The receiver's type **is** known: `_collect_local_types` records `{'s': 'Sub'}` for `s = Sub()`, and
  the typed-receiver path works — the control above proves it. It is only base-unaware.

[...]

PR #441, opened 2026-08-31
fix(parsers): typed receivers resolve inherited methods — the same-file base walk the C suite already declared the floor (#318)

gadievron:
## Summary

The call-graph builder resolved an inherited method when the receiver was `self` or `super()`
(two BFS walks over the class bases existed) but dropped the same call when the receiver was a
**locally-constructed instance** (M1: `s = Sub(); s.inherited()`) or the **class name** (M2:
`Sub.inherited_cm()`) — the typed-receiver path never consulted `class_data['bases']`. A method
whose only callers used a typed receiver got an empty caller set and was **pruned while its
caller was kept** — false negatives (~12% of typed-receiver calls lost on the issue's stdlib
census). The project had already decided the required behaviour in C (the dispatch suite's Bug
[30], the sound floor): walk up the base chain to the first ancestor defining the method,
same-file, own-first, no derived-override fan-out, no ancestor => no edge.

## The fix

- **`_same_file_base_walk`** (the `_resolve_self_call`'s own-first FIFO, cycle-guarded order,
  restricted to the caller's file) wired into every typed-receiver path: `_resolve_class_method`
  (M1 — before the cross-file simple-name match), `_resolve_module_call` (M2 — the gate fixed
  per the need-check to key on the class index so method-less subclasses, the most common M2
  shape, reach it), the constructor fallback, the `__call__` dunder, and the property read.
- The wave's guards (7 rounds + the deep-refute, each catching real seams in the prior):
  function-local classes can't hijack imported namesakes (the extractor marks them
  `function_local`; the defining caller's own exception wired into every path); a receiver the
  caller's scope **binds** is a variable, not a class reference — with the scope rules each
  pinned (params, def/class names, walrus, for/with, except-aliases, match-captures,
  comprehensions, class bodies, `global`, imports); subscripted bases (`Sub(Base[int])`)
  no longer sever the chain; the merge reconciles `function_local` and keeps the self/super
  walks' union in `all_bases` while the file-scope bases stay module-side.

## Disclosed limits

- **FIFO, not C3 MRO** (shared with the self walk): a diamond picks a single genuine ancestor,
  not the MRO-correct one — pinned exactly; the C3 unification filed as #440.
- **Imported bases/subclasses stay unresolved** for typed receivers (the caller-file-anchored
  walk can't see them — cross-file inheritance is `PARSER_UPGRADE_PLAN.md:170`): filed as #440.
- The merged-bases over-seed: two same-scope function-local namesakes' bases union (the
  `defined_here` exception walks the union inside either defining function) — the safe
  direction, pinned as CURRENT semantics.
- The try/except-ImportError shim: a module-level fallback class + an import gives M1 the
  fallback and M2 the import (two answers for one name; both-hit-emit-both is the deferred
  refinement).
- For a bound-variable receiver, the own-method scan is skipped too (only the TYPED resolution
  applies) — decided and pinned.

<details>

| check | command | result |
|---|---|---|
| new tests | the issue-318 test file (46 tests) | green; the worktree smokes RED at each round boundary (5 on pristine master, 3 on the pre-r1 commit, 3 on pre-r2, 1 on pre-r5, 1 on pre-r6) # proc-ok |
| full suites | the parser suite; the full suite | the parser side 215 green · the full side 3500 green + the one byte-identical-on-pristine macOS tmp failure # proc-ok |
| the corpus differential | master vs fix on anyio/dateutil/annotated_types | 34 gained / **0 same-file lost / 0 unredirected / 0 newly orphaned**; the PRUNE-DELTA: **11 functions rescued** (their first caller gained — AsyncFile.read/write/seek… no longer prunable) |
| the e2e (the real binary) | the shipped CLI's parse on the issue's five-receiver fixture | M1=True M2=True — both shapes resolve through the real pipeline |
| review | the need-check ("ship the design — strictly dominates the issue's options") + **7 wave rounds + the deep-refute** | each round caught real seams in the prior (the M2 gate, the hijack, the block-wrapped escape, the merged bases, the collector's scope rules, the defining-caller exception, the wiring completeness); the deep-refute caught the merge-reset severing the self/super walks — fixed and pinned |

</details>

Refs #318

### DIFF
--- a/libs/openant-core/parsers/python/call_graph_builder.py
+++ b/libs/openant-core/parsers/python/call_graph_builder.py
@@ -1008,6 +1008,11 @@ class CallGraphBuilder:
             return None
 
         # Check if obj_name is an imported module/class
+        # (panel r4 note): receiver_is_local_var deliberately does NOT gate
+        # this import branch — a local variable whose name matches an
+        # IMPORTED symbol still resolves through the import (the pre-
+        # existing over-seed direction; a local-var gate here would drop
+        # real imported-symbol calls that only happen to share a name).
         file_imports = self.imports.get(caller_file, {})
         if obj_name in file_imports:
             import_path = file_imports[obj_name]
--- a/libs/openant-core/parsers/python/call_graph_builder.py
+++ b/libs/openant-core/parsers/python/call_graph_builder.py
@@ -186,6 +186,11 @@ class CallGraphBuilder:
         # Local variable -> constructor-type map, so that `v = ClassName(); v.method()`
         # dispatches to the bound type's method.
         local_types = self._collect_local_types(tree)
+        # #318 (wave r5): the caller's own bound names AND the classes it
+        # defines at its own level (a receiver named for one of those is a
+        # class reference in this caller, not a variable).
+        bound_names = self._collect_bound_names(tree)
+        defined_classes = self._collect_defined_classes(tree)
         # #309 (item 4): a type-annotated parameter carries its receiver
         # type in the signature — consult it (a local Assign binding still
         # wins, matching Python's own scoping).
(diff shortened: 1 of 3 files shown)


gadievron:
Panel verdicts applied (fable: FIX x3; sonnet: FIX x1 — all four items landed on-branch):

- Sonnet 1 + fable 1 (both REAL, merged into one fix): the scope-mismatch branch of _merge_class_data dropped bases two ways — (a) aliasing: existing['bases'] was reset before local_side read it in the module-after-local order (sonnet); (b) all_bases was rebuilt from bases alone, discarding the prior all_bases on a local→module→local sequence (fable). all_bases now captures local bases before the reset AND unions the pre-existing all_bases. Three regression tests added, two RED-confirmed pre-fix.
- Fable 2 (doc): the 'never drops data' docstring claim replaced with the reset/union split.
- Fable 3 (doc + consistency): the _resolve_module_call import branch's deliberate non-gating noted; function_local now forwards through the nested-class recursion and class-body block descent — a class nested inside a function-local class was mis-keyed as module-level.

Gate: Run A green twice (3377 pre-panel, 3506 post-fix, the 49-test file included); parse-level differential probe: master emits no edge for Child.helper(self) typed-receiver inherited call (the #318 bug live), PR tree resolves it, no fabricated extras; CI green pre-push.

PR #443, opened 2026-08-31
fix(server): the web-UI invoke path honors the operator deadline — a wedged subprocess no longer holds a scan slot forever (#320)

### LINKED ISSUE #320: webui scan subprocesses run with no deadline, so a hung parser holds one of four scan slots indefinitely and `OPENANT_INVOKE_TIMEOUT` has no effect (PR #237's stated follow-up)
## Summary

Scans started from the web UI run their Python subprocess with **no deadline at all**. The job
context is cancel-only, and the `InvokeCtx` / `InvokeCtxCapture` entry points never consult
`resolveInvokeTimeout()`, so `OPENANT_INVOKE_TIMEOUT` has no effect on this path. The CLI path is
bounded at 30 minutes by default; the server path is unbounded.

Because a running scan holds one of four server slots for its whole duration, a wedged subprocess
permanently consumes a quarter of the server's scan capacity with no automatic recovery.

This is the follow-up merged PR #237 said it was leaving for later.

## The job context carries no deadline

`apps/openant-cli/internal/server/server.go:655`:

```go
	ctx, cancel := context.WithCancel(context.Background())
```

`context.WithCancel` gives cancellation but no timeout. Repo-verified: `server.go` contains exactly
two `context.WithTimeout` call sites, and neither covers the Python subprocess —

```
1520:	rctx, cancel := context.WithTimeout(ctx, 5*time.Second)     // DNS resolution
1545:	ctx, cancel := context.WithTimeout(ctx, 15*time.Minute)     // git clone
```

## The invoke path never applies the configured timeout

`apps/openant-cli/internal/python/invoke_ctx.go` contains **zero** occurrences of
`resolveInvokeTimeout` (`/usr/bin/grep -c resolveInvokeTimeout` → `0`). It takes the caller's context
and hands it straight to `exec.CommandContext`. The CLI path does the opposite —
`apps/openant-cli/internal/python/invoke.go:91`:

```go
	ctx, cancel := context.WithTimeout(context.Background(), resolveInvokeTimeout())
```

with `defaultInvokeTimeout = 30 * time.Minute` and the `OPENANT_INVOKE_TIMEOUT` override documented
alongside it.

## Complete site enumeration (CODE-TRACED, not executed)

All three server call sites pass a deadline-free context:

```
server.go:1152   python.InvokeCtxCapture(job.ctx, ...)   // the scan itself
server.go:1620   python.InvokeCtxCapture(ctx, ...)       // report-data
server.go:1669   python.InvokeCtx(ctx, ...)              // summary (makes LLM calls)
```

I have **not** executed a hung scan against a live server — that needs a wedged parser and a running
`openant serve`. The evidence here is the call graph, and I am labelling it as such.

## Operational consequence

`server.go:221` sizes the scan semaphore at four:

```go
		sem:            make(chan struct{}, 4),
```

and `:1096-1097` holds a slot for the entire run:

```go
	case s.sem <- struct{}{}:
		defer func() { <-s.sem }()
```

So an unbounded subprocess does not merely hang its own job — it removes 1 of 4 slots until someone
cancels it by hand. The CLI cannot get into this state.

## Why this is not the "cancel-only by design" case

`server.go:1541-1545` is worth quoting in full, because read partially it looks like a ratification
of the missing deadline and it is the opposite:

```go
	// Bound the clone in time so a hostile remote that streams forever can't hold a
	// scan slot indefinitely (the parent ctx is cancel-only). Disk size is not
	// bounded here — a huge working tree is a documented limitation for a
	// local, user-chosen scan target.
	ctx, cancel := context.WithTimeout(ctx, 15*time.Minute)
```

"The parent ctx is cancel-only" is the *justification for adding a local deadline*, not an argument
that none is needed. The same pattern appears at `:1520` for DNS. The maintainers' approach is
evidently: leave the job context cancel-only, and bound each subprocess that could run away. The
Python invoke is the one that never got that treatment.

## Prior art — this is the maintainers' own stated follow-up

Merged **PR #237**:

> that webui path does not honor `OPENANT_INVOKE_TIMEOUT`... unifying the timeout handling across
> both is a reasonable follow-up once both are merged.

Both are merged. No issue tracks it — a search across all open and closed issues for
`OPENANT_INVOKE_TIMEOUT` / `InvokeCtx` / webui-timeout returns nothing.

## Suggested fix

1. Apply `resolveInvokeTimeout()` in `InvokeCtx` / `InvokeCtxCapture`, or wrap the three server call
   sites in `context.WithTimeout(ctx, resolveInvokeTimeout())` — matching the local-bound pattern
   already used at `:1520` and `:1545`.
2. Keep the job context cancel-only. This asks for parity with the CLI's operator escape hatch, not
   for a job-level deadline that would truncate legitimately long scans; `OPENANT_INVOKE_TIMEOUT`
   already exists precisely so an operator can raise it.
3. On timeout, surface a distinct status so the UI can show "timed out" rather than a generic failure.

## What I am not claiming

- I am **not** claiming a hang has occurred in a real deployment. The evidence is the call graph;
  the consequence is code-traced, not executed.
- I am **not** asking for a deadline on the job context itself — see fix item 2.
- I have **not** measured how often a Python subprocess wedges in practice.

---

### Grounding curve (added 2026-08-22)

| commit | job ctx cancel-only | `invoke_ctx` skips timeout | 3 call sites | sem=4 | env var exists | verdict |
|---|---|---|---|---|---|---|
| `ad17f3f` (PR #133) | no server | file absent | — | — | partial (30 min fixed) | FAILS |
| `b8844d4` (PR #235) | true | true | true | true | not yet | PARTIAL |
| `a601f82` (PR #237) | true | true | true | true | true | FULL |
| `b5019628` (HEAD) | true | true | true | true | true | FULL |

**Attribution:** every server-side leg entered in one squashed commit — **PR #235** (`b8844d4`,
"feat(webui): add a local web UI for the scan pipeline"). `git log -- internal/server/server.go`
returns **exactly that one commit**, so there has been no drift since. **A fix belongs against #235.**
PR #237 (`a601f82`) added `resolveInvokeTimeout` and `OPENANT_INVOKE_TIMEOUT` to `invoke.go` only —
it never touched the server. Note the headline phrasing is only *sayable* from #237: before it there
was no env var to have no effect, and the honest statement would have been "the server path has no
deadline while the CLI is fixed at 30 minutes".

gadievron:
## Summary

The web-UI path ran its Python subprocess with **no deadline at all**: the job context was
cancel-only, and `InvokeCtx`/`InvokeCtxCapture` never consulted `resolveInvokeTimeout` — so
`OPENANT_INVOKE_TIMEOUT` had no effect on this path while the CLI path was bounded at 30
minutes by default. A running scan holds one of four server slots for its whole duration, so
a wedged subprocess permanently consumed a quarter of the server's scan capacity with no
automatic recovery. The maintainers' own pattern bounds each runaway subprocess locally (DNS
5s, git clone 15m); PR #237 named this the follow-up.

## The fix (through 4 wave rounds + a deep-refute, every finding fixed and pinned)

- `invokeCtxInner` derives `context.WithTimeout(ctx, resolveInvokeTimeout())` from the
  **caller's** context — the job cancel still propagates, the deadline bounds, and the
  override works on both paths.
- On the deadline: the distinct `ErrInvokeDeadline` error (the effective value + the override
  hint); the server maps it to a NEW **`timeout` job status** — `setTimeout` (idempotent),
  persisted into `meta.json` on the transition, and **restored by `recoverJobs`** (a timed-out
  job stays "Timed out" across a restart). The UI gets a distinct amber **"Timed out"**
  badge and a scan-page line naming the escape hatch; the slot is recovered.
- **Windows**: the kill gate branches on GOOS (mirroring the CLI's #433 fix —
  `TerminateProcess` yields exit code 1, never negative, which would have surfaced the kill as
  "vulnerabilities found").
- **The envelope-wins parity** (the #433 rule, now on the server path): a COMPLETE captured
  envelope beats the deadline — the envelope's own status drives the exit code (success → 0;
  else → 2 — the server's own contract, stated in the comment) plus a recovered note; the
  kill's exit artifact never leaks. The zombie-kill window (the child exited 0, a descendant
  held a pipe) keeps its success.
- The job context stays **cancel-only** (the issue's explicit scope); the cancel contract
  ("", -1, nil) pinned unchanged.

## Known limits (disclosed)

- The report/summary invokes also inherit the deadline (each a fresh 30m budget — worst-case
  job wall clock is 3× the budget); a report-step deadline surfaces as a `[report] Warning`
  (the job still ends `done`/`error`, not `timeout`).
- A legitimate exit racing the deadline on Windows is mis-attributable — the same documented
  trade-off the CLI's GOOS branch carries.

## The record

<details>
<summary>The checks (commands + results)</summary>

| check | command | result |
|---|---|---|
| new tests | the invoke_ct timeout + the server persist test files (11 tests) | green; RED on pristine # proc-ok |
| full suite | the campaign go-suite wrapper | 10 ok / 0 fail # proc-ok |
| lint | the CI scope over the touched files | green # proc-ok |

</details>

**The real-binary e2e**: a real `openant serve` job with
`OPENANT_INVOKE_TIMEOUT=200ms` and a wedged scan — the live `[timeout] **The real-binary e2e**: a real `openant serve` job with
`OPENANT_INVOKE_TIMEOUT=200ms` and a wedged scan — the live `[timeout]` log line with the
named cause + the override hint, the `meta.json` carrying `id/repo/started_at + status:
timeout`, and the index badge rendering "Timed out".

Fixes #320


gadievron:
Full acceptance gate (openant-pr + ship-fix, T3) — review verdict:

**Verdict: mergeable after the on-branch fixes.** Evidence:
- **RED (re-derived)**: the branch's timeout tests on pristine base 5024b418 fail to BUILD (`undefined: ErrInvokeDeadline`, `undefined: StatusTimeout / setTimeout`) — the machinery does not exist at base; GREEN on branch (python 219s ok, server ok, twice).
- **Wave (4 axes)**: state-machine + consumers-drift-twin FINDINGS (converging on the scan-page red-vs-amber contradiction + two report-consumer items — one re-derived NOT a bug, the non-fatal warning path), regression CLEAN, adversarial-races CLEAN. The drift-twin enumeration verified the CLI-vs-ctx divergences are intentional and documented (success->0 vs the CLI's conservative 2; GOOS gating present on both).
- **Deep-refute (fable)**: 5 findings — the vacuous regression test (MEDIUM: `exit 0` before `sleep 30 &` — the descendant never spawned, test vacuously green), the report-deadline fall-through (a SUCCEEDED scan marked error when only the report invoke timed out), the whole-buffer-JSON envelope-recovery strictness (a noisy stream with a complete envelope reported a false kill), the amber-dot cross-consumer inconsistency, and two INFO notes (stale-diff caveat; the per-subprocess slot budget ≈1h45m at defaults — documented limitation).
- **All fixed on-branch** (the just-pushed commit): the test spawns the descendant before exit; both report steps map ErrInvokeDeadline to the timeout state; deadlineOutcome scans bottom-up for the last well-formed envelope (the server's own noise tolerance); the scan page renders the amber dot + timeout-specific banner.
- **Live experiment ($0)**: the wedged-stub probe through the real server binary — OPENANT_INVOKE_TIMEOUT=2s → job status `timeout`, SSE done event, slot recovered, status persisted to meta.json, and **restored as Timed out after a full server restart**. The whole #320 contract, live.
- **Real E2E ($1.62 < $2 cap)**: a full 13-unit via-sonnet scan through the web-UI on the PR binary — job done, green badge, honest scan.report (success).
- Collision: re-verified merge-tree vs CURRENT master (c92c6bc) exit 0; base staleness nil (all evidence re-derived at the branch head).

PR #456, opened 2026-09-01
fix(stage1): a consistency correction writes finding too — gated on the Stage-1 finding verdicts (#331)

### LINKED ISSUE #331: Stage-1 consistency corrections are shadowed by a stale `finding`, so a corrected VULNERABLE is never verified and never disclosed
## Summary

When the Stage-1 consistency pass corrects a unit's verdict it writes `result["verdict"]` and does not
write `result["finding"]`. Ingestion has already set a lowercase `finding`, and the canonical
downstream read is finding-first — `str(r.get("finding") or r.get("verdict", "")).lower()` — so the
stale `finding` short-circuits the `or` and the correction is invisible.

A `safe → VULNERABLE` correction leaves `{finding: "safe", verdict: "VULNERABLE"}`. The unit is counted
safe, filtered out of Stage 2 entirely, and absent from the disclosure document.

**There is a second edge, and it points the other way.** Because the stale `finding` also survives a
*downgrade* to a verdict the block-list does not cover, it is currently acting as an accidental safety
net. The obvious one-line fix removes that net and introduces a different false negative. The fix
below is gated accordingly.

## The lost write

`libs/openant-core/utilities/stage1_consistency.py:297`:

```
297|                            if old_verdict_norm != new_verdict:
298|                                result["verdict"] = new_verdict
299|                                result["stage1_consistency_update"] = {
```

`/usr/bin/grep -nE '\["(verdict|finding)"\] *=' libs/openant-core/utilities/stage1_consistency.py`
returns one line — `:298`. The file writes `verdict` once and `finding` never.

Ingestion guarantees the shadowing value. `libs/openant-core/core/analyzer.py:149`:

```
149|        if result.get("finding"):
150|            result["finding"] = str(result["finding"]).lower()
151|        elif result.get("verdict"):
152|            result["finding"] = str(result["verdict"]).lower()
```

Upstream of that, `libs/openant-core/core/analysis_core.py:33-56` (`_normalize_result`) synthesises
`verdict` from `finding`, so both keys are present and agreeing before the consistency pass runs — and
that pass runs later, at `libs/openant-core/core/analyzer.py:686`.

## Where the stale value is read

| consumer | effect |
|---|---|
| `libs/openant-core/core/verifier.py:107` — Stage-2 **input** gate | the corrected unit is not in `vulnerable_results`, so **Stage 2 does not verify it at all** |
| `libs/openant-core/core/verifier.py:351-356` — `confirmed_findings` | same canonical read; the unit is absent on the Stage-2 path |
| `libs/openant-core/core/reporter.py:313-316` | the fallback filter, used when `confirmed_findings` is absent (`:307-308`) — same outcome when Stage 2 did not run |
| `libs/openant-core/core/analyzer.py:331` — `_count_verdicts` | tallies it as `safe` |

`_count_verdicts` is worth separating: its expression is
`r.get("finding", r.get("verdict", "error").lower())` — a default-on-**absent**, not an `or`-on-falsy.
It reaches the same wrong answer here by a different route, so a fix targeting only the `or` idiom
would miss it.

## Executed at `b501962`

Harness attached in the comment below; it reads only and makes no API calls.

```
A. _count_verdicts (core/analyzer.py:320)
   corrected-only  -> vulnerable=0  safe=1
   CONTROL vuln    -> vulnerable=1  safe=0
   CONTROL safe    -> vulnerable=0  safe=1
   mixed           -> vulnerable=1  safe=2

B. canonical finding-first read (core/reporter.py:315)
   corrected (finding=safe, verdict=VULNERABLE) -> 'safe'  in confirmed? False
   CONTROL   (finding=vulnerable)              -> 'vulnerable'  in confirmed? True

C. writes in utilities/stage1_consistency.py
   result["verdict"] : 1
   result["finding"] : 0
```

`CONTROL vuln` is the discriminating control: the same function, on a row whose `finding` agrees with
its verdict, counts correctly. The zero in A is an effect of the disagreement, not a broken harness.

An independent reviewer drove the **real** `run_stage1_consistency_check` with a stubbed model call
and confirmed the emitted shape is `{'finding': 'safe', 'verdict': 'VULNERABLE'}` — i.e. the rows
above are what the pass actually produces, not only what it is read to produce.

## The negation, executed

The claim fails if anything re-derives `finding` after the pass. Enumerated (unedited output):

```
$ /usr/bin/grep -rn '\["finding"\] *=' --include="*.py" libs/openant-core | grep -v /tests/
libs/openant-core/core/analyzer.py:150:            result["finding"] = str(result["finding"]).lower()
libs/openant-core/core/analyzer.py:152:            result["finding"] = str(result["verdict"]).lower()
libs/openant-core/core/verifier.py:169:    # checkpoint and ``finding_verifier.py`` ``r["finding"] = cp_data["finding"]``
libs/openant-core/utilities/json_corrector.py:251:                    extracted["finding"] = extracted["finding"].lower()
libs/openant-core/utilities/finding_verifier.py:434:                # sets result["finding"] = correct_finding, and the report
libs/openant-core/utilities/finding_verifier.py:624:                    r["finding"] = cp_data["finding"]
libs/openant-core/utilities/finding_verifier.py:747:                result["finding"] = verification.correct_finding
libs/openant-core/utilities/finding_verifier.py:918:                                result["finding"] = new_verdict
libs/openant-core/experiment.py:558:                    r["finding"] = r["verdict"].lower()
libs/openant-core/experiment.py:595:                        result["finding"] = verification.correct_finding
```

Two are comments (`:169`, `:434`). The `analyzer.py` pair runs at ingestion, before the pass. The
`finding_verifier.py` writes are Stage 2 — which the corrected unit does not reach, per
`libs/openant-core/core/verifier.py:107`. `experiment.py` is not the production pipeline. No production
write re-derives `finding` between the pass and the consumers.

## The naive fix is unsafe — measured

The obvious fix is to write both keys at `:298`. **Do not ship that.** The downgrade guard at
`libs/openant-core/utilities/stage1_consistency.py:287` blocks only
`old ∈ {VULNERABLE, BYPASSABLE} AND new ∈ DISCLOSURE_DROPPED`, and `DISCLOSURE_DROPPED` is
`{inconclusive, protected, rejected, safe}` (`libs/openant-core/core/verdict_taxonomy.py:86-91`).
`new_verdict` is unvalidated model output — the code's own comment at `:281` says so, and
`_resolve_stage1_inconsistency` (`:334-342`) `json.loads` it with no enum check.

Executed:

```
DISCLOSURE_DROPPED = ['inconclusive', 'protected', 'rejected', 'safe']

  VULNERABLE -> INSUFFICIENT_CONTEXT   blocked_at_287=False
      today     canonical='vulnerable'             in confirmed? True
      UNDER FIX canonical='insufficient_context'   in confirmed? False   <-- REMOVED
  VULNERABLE -> NOT_VULNERABLE         blocked_at_287=False
      today     canonical='vulnerable'             in confirmed? True
      UNDER FIX canonical='not_vulnerable'         in confirmed? False   <-- REMOVED
  VULNERABLE -> SAFE                   blocked_at_287=True
```

`INSUFFICIENT_CONTEXT` is not hypothetical — it is a first-class verdict in `_normalize_result`'s map
at `libs/openant-core/core/analysis_core.py:55`, and it is not in `DISCLOSURE_DROPPED`. So the stale
`finding` is presently masking unrecognised-verdict downgrades, and an ungated write of `finding`
would drop a disclosed vulnerability in the two cases measured above.

## Suggested fix

Gate the write on the new verdict being disclosure-eligible, so the correction lands for upgrades
without removing the accidental protection for unrecognised downgrades:

```python
                            if old_verdict_norm != new_verdict:
                                result["verdict"] = new_verdict
                                # Canonical downstream reads are finding-first, and ingestion
                                # always sets `finding` (core/analyzer.py:149-152), so writing
                                # only `verdict` leaves the correction shadowed. Gate on
                                # disclosure-eligibility: `new_verdict` is unvalidated model
                                # output (see the comment at :281), and writing an unrecognised
                                # value here would drop a disclosed finding.
                                if str(new_verdict).lower() in DISCLOSURE_ELIGIBLE:
                                    result["finding"] = str(new_verdict).lower()
```

The alternative — widen `:287`'s block-set from `DISCLOSURE_DROPPED` to "anything not
disclosure-eligible", then write both keys without the gate — is the more complete fix, and it is the
open question PR #245 deferred. Either way `_count_verdicts` at `core/analyzer.py:331` should be
brought onto the same canonical read.

Regression tests should assert (1) after a `safe → VULNERABLE` correction the unit is in
`_count_verdicts`' `vulnerable` bucket and in `confirmed_findings`, and (2) after a
`VULNERABLE → INSUFFICIENT_CONTEXT` correction it is **still disclosed**. I found no existing test
covering a correction whose `finding` and `verdict` disagree.

## Prior art, checked

`libs/openant-core/tests/test_verifier_verdictonly_confirmed_drop.py` exists and passes at `b501962`
(3 passed). Its subject is a result with **no `finding` key**, fixed by adding `or r.get("verdict")`
fallbacks. Those are not reached here: the `or` short-circuits on a present-but-stale value, as the executed
runs above show.
`_resolve_stage1_finding`'s docstring states the absent-key intent explicitly at
`libs/openant-core/utilities/finding_verifier.py:209-211`, and the same assumption appears at
`libs/openant-core/experiment.py:557` (`if finding is None`).

PRs #193, #195, #243 and #245 all touch this function and all left `:298` writing one key.

## Relation to existing issues

- **#324** — a Stage-1 result with **neither** key; this is one with **both**, disagreeing.
- **#316** — the JSON corrector synthesising an *unrecognised* verdict string. Adjacent in a way that
  matters here: #316's mechanism is what makes the naive fix above dangerous.

## What I am not claiming

- Not that a specific vulnerability was missed in a real scan. The evidence is the production sinks
  and the canonical reads executed at `b501962`, plus an independent drive of the real consistency
  pass with a stubbed model call. No end-to-end CLI run against a real repository was performed.
- Not that every unit is affected — only units the consistency pass actually corrects. I did not
  measure how often it fires on a real corpus.
- `finding` is present only for units arriving with either key; `core/analyzer.py:149-152`
  sets nothing when both are absent, which is #324's case and where this mechanism does not fire.
- The suggested fix is reasoned from the executed downgrade table above; I have not run the gated
  variant against the full suite.

gadievron:
The Stage-1 consistency pass corrected a unit's verdict by writing `result["verdict"]` and never `result["finding"]`. Ingestion always sets a lowercase `finding` (core/analyzer.py:149-152), and every canonical downstream read is finding-first — `str(r.get("finding") or r.get("verdict", "")).lower()` (core/verifier.py:118) — so the stale `finding` short-circuited the `or` and the correction was invisible: a safe→VULNERABLE correction left `{finding: "safe", verdict: "VULNERABLE"}`, counted safe, filtered out of Stage 2 entirely, absent from the disclosure document (the issue's executed A/B/C probe).

**The second edge points the other way:** the stale finding also survives a downgrade the `:287` block-list does not cover (DISCLOSURE_DROPPED is {inconclusive, protected, rejected, safe}; VULNERABLE→INSUFFICIENT_CONTEXT passes unblocked) — today the stale finding is an ACCIDENTAL SAFETY NET keeping the row disclosed. The naive write-both-keys fix REMOVES that net (the issue measured both INSUFFICIENT_CONTEXT and NOT_VULNERABLE moving out of confirmed under it) — do not ship the ungated write.

**The fix (base commit):** the finding write is gated on the new verdict being a Stage-1 finding verdict — upgrades land for every finding-first reader; unrecognised downgrades keep the stale finding (still disclosed) with the correction recorded in `stage1_consistency_update`. The alternative (widen the block-set, write ungated) is PR #245's open question; this is the smaller safe slice. `_count_verdicts` (the issue quoted its pre-#324 expression) is already finding-first post-#324 — pinned by test, not changed. 4 tests, RED 1 on pristine; the stage1/verifier neighbors 16/0; the full suite 3520/1 (the known macOS artifact).

**The wave-r1 round (three axes, one root cause):**
- the gate checked DISCLOSURE_ELIGIBLE — which admits Stage-2 vocabulary (unverified, confirmed, agreed, error) that every finding-first reader REJECTS: a VULNERABLE→UNVERIFIED correction wrote `finding="unverified"` and dropped a disclosed vulnerability from Stage-2 input, confirmed_findings, and disclosure — the net's own failure mode reintroduced through the gate. The gate is now SEVERITY_FINDING_VERDICTS ({vulnerable, bypassable} — exactly the pair the readers accept);
- a promoted ERRORED row kept its stale `error` key — one row counted as BOTH a confirmed finding and an error; the gated write clears it.

**Evidence:** 6 tests (the four Stage-2 tokens each keep the stale-finding net; the promoted-ERROR row carries neither the error key nor its old verdict; the originals); the stage1 families green; the full suite 3522/1 (the known macOS artifact); ruff clean.

Fixes #331

### DIFF
--- a/libs/openant-core/tests/test_issue331_finding_shadow.py
+++ b/libs/openant-core/tests/test_issue331_finding_shadow.py
@@ -132,3 +132,57 @@ def test_verdict_only_row_counts_via_verdict_fallback():
     ])
     assert counts["vulnerable"] == 1
     assert counts["errors"] == 1
+
+
+def test_stage2_vocabulary_keeps_the_net():
+    """Wave r1 (three axes, one root cause): DISCLOSURE_ELIGIBLE admitted
+    Stage-2 vocabulary the finding-first readers REJECT — a
+    VULNERABLE -> UNVERIFIED correction wrote finding="unverified" and
+    dropped a disclosed vulnerability from Stage-2 input, confirmed_findings,
+    and disclosure: the net's own failure mode reintroduced through the gate."""
+    def resolver(binding, group, code_by_route, tracker):
+        return s1.Stage1ConsistencyResult(
+            "p", "UNVERIFIED",
+            [{"route_key": VULN_RK, "original_verdict": "VULNERABLE",
+              "should_be": "UNVERIFIED", "reason": "x"}], "g")
+    out = _run(resolver, [INGESTED_VULN, INGESTED_SAFE])
+    row = _by_rk(out, VULN_RK)
+    assert row["verdict"] == "UNVERIFIED"
+    assert row["finding"] == "vulnerable", (
+        "the Stage-2 vocabulary must keep the stale-finding net "
+        f"(got {row['finding']!r})"
+    )
+    for token in ("CONFIRMED", "AGREED", "ERROR"):
+        def r2(binding, group, code_by_route, tracker, _tok=token):
+            return s1.Stage1ConsistencyResult(
+                "p", _tok,
+                [{"route_key": VULN_RK, "original_verdict": "VULNERABLE",
+                  "should_be": _tok, "reason": "x"}], "g")
+        out = _run(r2, [INGESTED_VULN, INGESTED_SAFE])
+        assert _by_rk(out, VULN_RK)["finding"] == "vulnerable", token
+
+
+def test_promoted_errored_row_clears_the_stale_error_key():
+    """Wave r1 (fable): a promoted ERROR row kept its stale `error` key —
+    one row counted as BOTH a confirmed finding (finding-first) and an error
+    (r.get("error")). The gated write clears it."""
+    errored = dict(INGESTED_SAFE)
+    errored["verdict"] = "ERROR"
+    errored["finding"] = "error"
+    errored["error"] = "LLMConnectionError: DNS lookup failed"
+    errored["route_key"] = VULN_RK
+
+    def resolver(binding, group, code_by_route, tracker):
+        return s1.Stage1ConsistencyResult(
+            "p", "VULNERABLE",
+            [{"route_key": VULN_RK, "original_verdict": "ERROR",
+              "should_be": "VULNERABLE", "reason": "pattern matches sibling"}],
+            "g")
+    out = _run(resolver, [errored, INGESTED_SAFE])
+    row = _by_rk(out, VULN_RK)
+    assert row["verdict"] == "VULNERABLE"
+    assert row["finding"] == "vulnerable"
+    assert "error" not in row, (
+        f"the stale error key makes the promoted row count as both a finding "
+        f"and an error: {row}"
+    )
--- a/libs/openant-core/tests/test_issue331_finding_shadow.py
+++ b/libs/openant-core/tests/test_issue331_finding_shadow.py
@@ -1,0 +1,134 @@
+"""#331: a Stage-1 consistency correction writes BOTH keys — gated on eligibility.
+
+When the Stage-1 consistency pass corrects a unit's verdict it wrote `result["verdict"]`
+and never `result["finding"]`. Ingestion (core/analyzer.py:149-152) always sets a
+lowercase `finding`, and every canonical downstream read is finding-first —
+`str(r.get("finding") or r.get("verdict", "")).lower()` (core/verifier.py:118) — so the
+stale `finding` short-circuited the `or` and the correction was invisible: a
+safe -> VULNERABLE correction left {finding: "safe", verdict: "VULNERABLE"}, counted
+safe, filtered out of Stage 2 entirely, absent from the disclosure document.
+
+The second edge points the other way: the stale finding also survives a downgrade to
+a verdict the block-list does not cover — `DISCLOSURE_DROPPED` is
+{inconclusive, protected, rejected, safe}, so `VULNERABLE -> INSUFFICIENT_CONTEXT`
+passes unblocked, and today the stale `finding` is an ACCIDENTAL SAFETY NET keeping the
+row disclosed. An ungated write of `finding` would remove that net and drop a disclosed
+vulnerability (measured in the issue: VULNERABLE -> INSUFFICIENT_CONTEXT /
+NOT_VULNERABLE both move out of `confirmed` under the naive fix).
+
+So the write is GATED: `finding` is updated only when the new verdict is
+disclosure-eligible (DISCLOSURE_ELIGIBLE); unrecognised downgrades keep the stale
+finding (still disclosed via the net) with the correction recorded in
+`stage1_consistency_update`.
+"""
+from core.analyzer import _count_verdicts
+from core.verdict_taxonomy import DISCLOSURE_ELIGIBLE
+from utilities import stage1_consistency as s1
+
+VULN_RK = "libs/a/client_utils.py:build_async_httpx_client"
+SAFE_RK = "libs/b/client_utils.py:get_async_httpx_client"
+
+
+def _run(resolver, results):
+    orig = s1._resolve_stage1_inconsistency
+    s1._resolve_stage1_inconsistency = resolver
+    try:
+        return s1.run_stage1_consistency_check(
+            [dict(r) for r in results], {}, binding=object(), tracker=None)
+    finally:
+        s1._resolve_stage1_inconsistency = orig
+
+
+def _by_rk(out, rk):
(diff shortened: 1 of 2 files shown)


gadievron:
Superseded by #466 — the same fix through the stronger wholesale-reject gate (the panel found the two PRs collide line-for-line in utilities/stage1_consistency.py; #466's verdict: 'the correct, stronger successor to #456'). Re-landing any #456-unique content there if the diff shows any.

Issue #420, opened 2026-08-30
An interrupted step's on-disk report claims status: success


gadievron:
Panel finding during #418's review (fable): when an interrupt propagates through a step (as #418 makes enhance do), `step_context`'s `finally` still writes the step report with the default `status="success"` and an empty summary — the on-disk artifact of an interrupted run lies (stdout envelope unaffected: no envelope is emitted on the KI path, so the Go exit-130 contract still works; this is the stderr/file channel only).

Suggested: map a propagating BaseException (not caught by `except Exception`) to `status: "interrupted"` in step_context, so the resume path and any artifact reader can see the run was cut short. Applies to analyze (#411) and verify once those fix their handlers too.

gadievron:
Fixed in PR #464 (CI green, all 14 checks; leave-for-review — not merged). Closing per the campaign protocol; the PR body carries the full root-cause + evidence.

Issue #417, opened 2026-08-29
SIGINT during the ENHANCE stage is silently swallowed — the run cannot be emergency-stopped mid-enhance


gadievron:
Live-run finding (raptor core/sandbox, 160-unit directed scan, sonnet, 2026-08-29): a SIGINT sent at 80/160 enhance units was **completely absorbed** — enhance processed all 160, transitioned to analyze, and finished normally. No interrupt report, no exit 130, no checkpoint-visible stop. #411 fixed the ANALYZE stage's interrupt (the None-placeholder AttributeError); the ENHANCE stage's KeyboardInterrupt path is separate and does not propagate. Consequence: the budget-stop mechanism (#380's 130 path) cannot stop a runaway enhance; a cost-cap kill during enhance silently burns the remaining queue. Repro: launch `openant scan ... --limit 160`, SIGINT during `[Enhance]` progress; observe the run continue through all units and into [Detect].

Suggested fix: the enhance loop's per-unit exception handling (the same `_process_file_guarded`-style boundary that isolates unit errors) should re-raise KeyboardInterrupt after draining in-flight units (mirror #411's analyze fix: print the interrupt report with the analysed/not-started split, then re-raise).

PR #463, opened 2026-09-01
fix(stage2/dynamic-test): an interrupt propagates — it is not a completion (#419)

### LINKED ISSUE #419: SIGINT during the VERIFY and DYNAMIC-TEST stages is still swallowed (same class as #417)
Panel enumeration during #418's review (fable): the #417 interrupt-swallow class persists at two more stages —

- **verify**: `utilities/finding_verifier.py:952` (sequential: prints 'progress saved', falls through) and `:987` (parallel: `executor.shutdown(...)` then bare `return`). SIGINT landing during verify reproduces #417's exact live-run shape: the scan completes, success envelope, no exit 130.
- **dynamic-test**: `utilities/dynamic_tester/__init__.py:343` — `return results` on KI; the caller then writes DYNAMIC_TEST_RESULTS.md from partial results. (`core/reporter.py:975` already re-raises — clean.)

#418 (enhance) and #411 (analyze, unmerged) fix the other stages. These two are the remaining recurrences.

gadievron:
The #417 interrupt-swallow class persisted at two more stages (panel enumeration during #418's review): the sequential verify handler printed 'progress saved' and FELL THROUGH — a SIGINT during verify made the scan continue with partially-verified results and complete with a SUCCESS envelope, no exit 130; the parallel handler shut the executor down and returned BARE — same swallow; and the dynamic-test handler `return results` on KI, handing the caller PARTIAL results which flowed into the step summary → the report step → the success envelope (the corrected framing from the wave round — the original commit message described a DYNAMIC_TEST_RESULTS.md partial-write that never existed on that path).

**The fix (base commit):** the #417/#418 contract at all three sites — the checkpoints hold the progress (every completed unit/finding is saved as it completes — that IS the resume story), and the KeyboardInterrupt itself PROPAGATES so the CLI can exit 130 with an interrupted envelope, never a success one. `core/reporter.py:975` already re-raises (clean); #418 (enhance) and #411 (analyze) fix the other stages; no other KI handler exists in the two call paths (KeyboardInterrupt is a BaseException — no broad except-Exception re-swallow).

**The wave-r1 round (fable+opus, all confirmed, all fixed):**
- the resume story's other half was untested: a KI arriving after a finding COMPLETED — the raise must not land before the completed unit's checkpoint save (a re-nested try would regress exactly this). Pinned: the first finding's checkpoint survives the interrupt;
- the parallel-path residual documented at the raise site (opus): cancel_futures cancels only queued work; the in-flight verification loops finish in non-daemon threads (bounded by the Go CLI's 5s SIGKILL; a direct `python -m openant` invocation waits them out in interpreter shutdown), and their results are not checkpointed — the retry owns them on resume. Not a regression; now stated instead of overstated;
- the step-report layer one layer up (all three axes): that is PR #420 (except BaseException → status "interrupted") — this PR and that one are a pair.

**Evidence:** 3 base tests (RED 3 on pristine) + the completed-unit-checkpoint pin; the verify/dynamic/interrupt families 132 passed; the full suite 3520/1 (the known macOS artifact); ruff clean.

Fixes #419

### DIFF
--- a/libs/openant-core/tests/test_issue419_ki_propagates.py
+++ b/libs/openant-core/tests/test_issue419_ki_propagates.py
@@ -1,0 +1,134 @@
+"""#419: an interrupt during verify or dynamic-test propagates — it is not a
+completion.
+
+The #417 interrupt-swallow class persists at two more stages (panel enumeration
+during #418's review):
+
+- `finding_verifier.py:952` (sequential): the KeyboardInterrupt handler prints
+  'progress saved' and FALLS THROUGH — a SIGINT during verify makes the scan
+  continue to the next stage with partially-verified results: the live-run shape
+  is the #417 one — the scan completes, success envelope, no exit 130.
+- `finding_verifier.py:987` (parallel): `executor.shutdown(...)` then a bare
+  `return` — same swallow, parallel path.
+- `dynamic_tester/__init__.py:343`: `return results` on KI, and the caller then
+  writes DYNAMIC_TEST_RESULTS.md from PARTIAL results — an interrupted run
+  presented with a completion artifact.
+
+`core/reporter.py:975` already re-raises (clean); #418 (enhance) and #411
+(analyze) fix the other stages. The contract: checkpoints hold the progress
+(every completed unit is saved as it completes — that is the resume story), and
+the KI itself PROPAGATES so the CLI can exit 130 with an interrupted envelope —
+never a success one.
+"""
+import json
+import sys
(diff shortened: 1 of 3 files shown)


gadievron:
Superseded by #464 — the two PRs collide line-for-line (identical test_issue419_ki_propagates.py + the same dynamic_tester/finding_verifier raises); #464 carries the interrupt propagation AND the interrupted-status fix. Opus panel: 'Supersedes #463.'

Issue #428, opened 2026-08-30
Pip floor-vs-pin drift between pyproject.toml and requirements.txt has no consistency check


gadievron:
Panel finding during #421's review (fable): #421 installs requirements.txt (the exact CI pins) BEFORE the editable install — good. But the two files are hand-maintained side by side, and pip's editable install still re-resolves [project.dependencies] floors. Today the floors (pydantic>=2.0.0, anthropic>=0.40.0, openai>=2.2.0, the eight tree-sitter floors, PyYAML, requests) are all BELOW the pins, so nothing drifts. The moment a pyproject floor is raised above a requirements pin, the editable install silently upgrades past the pin and NOTHING — no test, no CI step — catches it. Also anthropic[bedrock]'s extra still pulls boto3/botocore unpinned (botocore releases ~daily) and requirements' 10 floor lines copy pyproject verbatim. Cheap guards: a pip install --dry-run consistency check, or a comment-anchored test that pyproject floors <= requirements pins. Also: 'EXACT pins'/'deterministic' in the test banner overstates — the LLM-SDK chain is genuinely locked; the floors+extras are not.

gadievron:
Fixed in PR #468 (CI green, all 14 checks; leave-for-review — not merged). Closing per the campaign protocol; the PR body carries the full root-cause + evidence.

PR #479, opened 2026-09-01
fix(test): the #428 test follows the live anthropic pin, not the frozen snapshot (master CI is red)

gadievron:
Renovate #475 bumped the anthropic pin 1.2.0 → 1.3.0 and broke master's CI: two assertions in the merged #428 test file hardcoded the 1.2.0 figure. The drift-message test constructs a synthetic drift (1.3.0 floor vs 1.2.0 pin) and asserts the "upgrade past the CI pin" branch by checking the LIVE pinned-rows text contains "1.2.0" — which fails the day the live pin moves (exactly what happened); the parser test asserts `pins["anthropic"] == Version("1.2.0")` verbatim.

A pinned row is a moving target by design: the guard compares the two live files, so the assertions now follow the live requirements.txt (verified against master's own tree: 2 failed before, 5 passed after; full suite green except the known macOS artifact). This un-reds CI for every open PR merged against current master (e.g. #477).

### DIFF
--- a/libs/openant-core/tests/test_issue428_floors_vs_pins.py
+++ b/libs/openant-core/tests/test_issue428_floors_vs_pins.py
@@ -153,20 +153,24 @@ def test_the_guard_reports_not_crashes_on_drift():
     """famD panel (sonnet): the old fire path CRASHED (AttributeError) the
     moment drift existed — the guard must REPORT. Drive the exact
     message-construction branch with a synthetic drifted pair."""
-    floors = {"anthropic": Version("1.3.0")}
-    pins = {"anthropic": Version("1.2.0")}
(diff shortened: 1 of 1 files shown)


gadievron:
Superseded by #478 (merged at 20:04 — three minutes after this opened): the same un-redding, a different shape. #478 keeps the message-path test self-contained (the pinned text PATCHED, both branches asserted) rather than dropping the monkeypatch and asserting one arm against a synthetic 9.9.9 pin; the live-file parse assert is presence-based on both sides. Master is green on 0431e28b with the #478 form. If the campaign session prefers the 9.9.9-synthetic shape for the message test, that's a stylistic re-landing — but the fix this PR exists for is already shipped.

Issue #332, opened 2026-08-24
HTML report loads 3 CDN scripts with no SRI; server sets 7 security headers but no CSP


gadievron:
## Summary

The HTML report that `openant serve` renders and auto-opens in a browser loads **three third-party
scripts from public CDNs with no Subresource Integrity**, and the server that serves it sets **no
Content-Security-Policy**. The report page holds the findings of a security scan.

This is not an oversight of security posture in general — the same middleware sets seven other
security headers, and the same commit that added the auto-opening server (`b8844d4`) vendored two
other JS libraries locally. CDN pinning and CSP are the two gaps in an otherwise deliberate posture.

## The scripts

`apps/openant-cli/internal/report/templates/report-reskin.gohtml:10-12` and
`apps/openant-cli/internal/report/templates/overview.gohtml:7-9` — identical three lines in each:

```html
    <script src="https://cdn.tailwindcss.com"></script>
    <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-datalabels@2"></script>
```

`chart.js` carries **no version at all**; `chartjs-plugin-datalabels@2` pins a major only;
`cdn.tailwindcss.com` is a dynamic script. None carries `integrity=`.

Repo-wide (the counts are 0 with or without `node_modules` excluded, and the same grep run
against `X-Frame-Options` returns 1 — so these are real zeros, not a broken command):

```
integrity=                 -> 0
content-security-policy    -> 0
```

## This is a served, auto-opened page, not a static file on disk

Both templates are compiled in via `//go:embed` (`apps/openant-cli/internal/report/report.go:12,15`),
served by `apps/openant-cli/internal/server/server.go`, and `apps/openant-cli/cmd/serve.go:72` calls
`openBrowser(url)` immediately after start — the page opens by itself.

## Why this reads as inconsistent rather than unaware

**The server already sets seven security headers, and not CSP** — `internal/server/server.go:363-369`,
one consecutive block, quoted in full:

```go
		w.Header().Set("X-Frame-Options", "DENY")
		w.Header().Set("X-Content-Type-Options", "nosniff")
		w.Header().Set("Referrer-Policy", "no-referrer")
		w.Header().Set("Cross-Origin-Opener-Policy", "same-origin")
		w.Header().Set("Cross-Origin-Resource-Policy", "same-origin")
[...]

PR #480, opened 2026-09-01
fix(config): renovate.json's regex customManager uses named groups only — restores Renovate PRs (#476)

gadievron:
The #332 wave-round entry used a Handlebars-style conditional `packageNameTemplate` (`{{#if matches.1}}...`) — invalid for Renovate's regex custom manager, which supports only named-capture-group substitution. Renovate halted ALL PRs for the repository as a precaution (#476).

Each package now gets its own `matchString` with a named `packageName` capture group; `packageNameTemplate` dropped. The tracked surface is unchanged: the vendored report scripts' version-bearing file names (templates, the embed test, SOURCES.txt).

Apologies for the breakage — this entry came from my #457 wave round.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,15 +6,18 @@
   "customManagers": [
     {
       "customType": "regex",
-      "description": "#332: the vendored report scripts — versions live in the file names, the go:embed list, and the template literals; nothing else would look at them.",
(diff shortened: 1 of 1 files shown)


gadievron:
Superseded by the rebased branch: this branch was based on pre-#483 master — its diff would have reverted the ruff-scope PRs (#483/#484). The real change (renovate.json) is carried to the rebase at fix/480-renovate-named-groups with the panels' fix applied (the chart.js literal never matched the real vendored filename — all three panels caught it).

Issue #481, opened 2026-09-01
The review-bot comment corpus across 128 merged PRs: 88 comments, three infra follow-ups — all retired (items 1-2 via #483; item 3 via #500)


gadievron:
Swept all 89 code-review bot comments across every PR merged in the past week (106 PRs; 44 carried comments). Adjudication of the 10 prod-file comments: 8 false-positive/stale/intentional, 1 already restructured by #478, 1 real-but-cosmetic (the unused `SEVERITIES` import at analysis_core.py:23 — cleanup follows). Of the 79 test-file comments: all idiom-debt, zero behavioral defects. **The corpus contains no unaddressed behavioral finding.**

The conceptual yield — the comments cluster into three infrastructure gaps, each one fix that retires a whole class:

1. **A shared root conftest for the issue-numbered test files** (36 comments = the import-style class): 30+ test files hand-roll `sys.path` setup independently. The #462 conftest.py pattern (scoped autouse for the parser subtree) applied at the tests/ root retires the class wholesale.
2. **Ruff over the test tree in CI** (28 comments = unused-symbol class): prod is linted, tests are not — that is the entire delta.
3. **A pathlib-write fixture helper** (11 comments = the not-always-closed class): a `write(path, text)` helper or uniform `.write_text` retires the fixture-leak idiom.

The empty-except class (2 prod) is documented-by-docstring today; a one-line `# expected: already gone` inside each would silence the checker. The unreachable-code class (5, all pytest.raises blocks) is a checker-vs-pytest semantic mismatch — no action. #410's chmod is intentional (executable fixtures). #422's Go handle-close is the defer idiom on a lock whose unlock is file removal — no data-loss class.

gadievron:
CORRECTION to the adjudication above: the 'one real-but-cosmetic finding' (the unused SEVERITIES import) was itself a FALSE POSITIVE — analysis_core.SEVERITIES is a load-bearing RE-EXPORT consumed by the #436 one-enum-everywhere pin (test_issue215_severity_field.py reads it through the module namespace). The attempted cleanup failed its full-suite run (1 failed) and was reverted before merge. Final adjudication: the 89-comment corpus contains ZERO unaddressed real findings — all 10 prod-file comments are false-positive/stale/intentional/by-design, and the three infrastructure follow-ups (root conftest, test-tree ruff, fixture write helper) remain the corpus's only actionable yield.

gadievron:
[...]

PR #482, opened 2026-09-01
chore(core): drop the unused SEVERITIES import — the one real finding in the comments sweep

gadievron:
The 89-comment sweep across all 106 merged past-week PRs (issue #481) found exactly one unaddressed real prod finding: the `SEVERITIES` import at analysis_core.py:23 (from #436) is unused — `_SEVERITIES` is the used symbol. One-line cleanup; ruff clean; full suite green.

### DIFF
--- a/libs/openant-core/core/analysis_core.py
+++ b/libs/openant-core/core/analysis_core.py
@@ -21,7 +21,7 @@ from typing import TYPE_CHECKING
 from datetime import datetime
 
 from core.verdict_taxonomy import (
-    FINDING_VERDICT_ORDER, SEVERITIES, _SEVERITIES, STAGE1_VERDICTS,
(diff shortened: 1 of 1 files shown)


gadievron:
My adjudication was WRONG — the sweep's 'one real finding' was a false positive. The test_the_one_severity_enum_everywhere pin (from #436) reads analysis_core.SEVERITIES as a RE-EXPORT — the import is load-bearing for the enum-uniformity invariant, and the full suite caught it (1 failed) before any merge. The bot comment (unused-import at the diff position) and my sweep both missed the consumer because it reads the symbol through the module namespace, not a direct reference. Corpus re-adjudication: ZERO real unaddressed findings in the 89 comments. Reverted; branch closed. #481's report corrected below.

PR #483, opened 2026-09-02
chore(tests): the corpus-cleanup sweep + the test-tree-scoped ruff gate (#481)

gadievron:
The comments-corpus sweep (#481, re-derived wholesale in the retro-scan) found the test tree carrying **93 unused imports + 6 dead locals**, invisible under the repo's bug-only ruff select (F821/F811/F823 — a deliberate choice). This PR is the cleanup half of #481's infra fix:

**The scope** (pyproject.toml): F401 + F841 added to the select, per-file-ignored for the 8 prod packages + the 2 root scripts — prod keeps the bug-rules-only choice (its 87 F401s are the parser/idiom re-export candidates, a **separate investigation**, never bulk-autofixed by a chore sweep). The **efficacy fixtures are per-file-ignored** (benchmark data — the webapp oracle scores them; a style autofix mutating a fixture changes the corpus under measurement, the fable panel's catch: ruff's autofix had touched app_b.py, reverted).

**The cleanup**: ruff --fix for the 93 (the cross-file re-export check over all 72 touched files found **0 traps** — the git-diff cross-reader sweep; the #482 lesson's executed form); the 6 dead locals by hand (1 walrus leftover, 3 pass-through/dead-reassignment locals, 1 dead intermediate dict, 1 dead rebind; one rename touched 3 sibling readers in test_reachability_prune_telemetry, all restored in-commit).

**RED→GREEN**: the new scope flags 99 on pristine master; after the fix `ruff check .` is fully green. **The full suite on the deletion: 3718 passed / 0 failed** (the re-export adjudication's executed proof — the adjudication ran the suite, not just the grep).

**Gates**: full panel (sonnet+opus+fable): opus APPROVE (its honesty caveat caught the 8th package — github_scanner, added); fable FIX → all resolved (the fixtures catch, github_scanner, the narrative); sonnet FIX → prose corrections (this body). Bug-research verification: the git-diff cross-reader sweep over all 72 files — 0 re-export traps. Collision: clean (only #480 open in tree — renovate.json, no overlap). CI below; E2E on the branch follows.

PR #484, opened 2026-09-02
chore(core): the prod import-hygiene sweep — the #481 prod half

gadievron:
The comments-corpus sweep's prod half (#481): **~90 dead imports/locals removed across 51 files**, the ruff config flipped to **global F401/F841 scope** (the 8-package + 2-script per-file-ignores dropped; the fixtures-only ignore remains — the benchmark-data rule).

**The investigation** (per the substrate directive): the 92 findings are refactor residue, NOT a parser-template convention — the wide import block lives in c/php/ruby/**python** (7 python files, trimmed in this same diff), the minimal style in go/javascript/rust/swift/zig; no in-tree canonical template exists. The substrate: rapid issue-driven refactors leave imports; the sibling class was #483's 93 test-tree findings (same phenomenon, same fix shape).

**The adjudication found THREE from-import re-export traps** (the #482 class's from-import form — invisible to attribute-access grep; the SUITE caught each): `finding_verifier.MAX_TOOL_RESULT_CHARS` (verifier's caps probe), `experiment._normalize_result` (test_cwe_tagging), `analysis_core.SEVERITIES` (#436's one-enum pin) — each restored as an explicit-alias `noqa: F401` re-export with the consumer named in the comment. The formal from-import consumer sweep over all 33 touched modules: **2 traps, both satisfied** (one deliberate survivor + one restored). The mock.patch/getattr string-reference class over all 42 removed symbols: **0 traps**.

RED→GREEN: the flipped gate flags 102 on pristine; after the fix `ruff check .` fully green. **Full suite 3718 passed / 0 failed.** Panels: sonnet FIX(5) + fable FIX(2) — independently convergent (the python-parser claim + the dead statements), both applied; opus died mid-analysis but its crux (the from-import consumer surface) became the formal sweep above. CI below; E2E follows.

### DIFF
--- a/libs/openant-core/context/threat_model.py
+++ b/libs/openant-core/context/threat_model.py
@@ -60,7 +60,6 @@ from pathlib import Path
 from typing import Any
 
 from context.application_context import ApplicationContext
-from utilities.file_io import open_utf8
 
 # --- Schema v1 constants ------------------------------------------------------
 
--- a/libs/openant-core/context/threat_model_agent.py
+++ b/libs/openant-core/context/threat_model_agent.py
@@ -23,7 +23,7 @@ import json
 from pathlib import Path
 
 from utilities.file_io import read_repo_file, repo_path_state, write_repo_file
-from utilities.llm import PhaseBinding, simple_text
+from utilities.llm import simple_text
 
 from context.repo_explorer import explore_repository
 
--- a/libs/openant-core/core/analysis_core.py
+++ b/libs/openant-core/core/analysis_core.py
@@ -21,8 +21,14 @@ from typing import TYPE_CHECKING
 from datetime import datetime
 
 from core.verdict_taxonomy import (
-    FINDING_VERDICT_ORDER, SEVERITIES, _SEVERITIES, STAGE1_VERDICTS,
+    FINDING_VERDICT_ORDER,
+    _SEVERITIES,
+    STAGE1_VERDICTS,
 )
+# RE-EXPORT, not unused: test_issue215's one-enum-everywhere pin asserts
+# `analysis_core.SEVERITIES is SEVERITIES` — the from-import re-export form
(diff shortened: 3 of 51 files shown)


gadievron:
**E2E receipt (owed — the PR body said "E2E follows")**

The mechanism = deleted dead prod imports/locals + the three explicit-alias re-exports. Here a real-binary run IS mechanism-relevant (prod deletions exercise end-to-end). Executed on current master (325976f):

1. **Prod pipeline exercised, real binary, $0**: `python -P -m openant parse <corpus> --output <dir>` (a 2-file Python corpus; the engine = the tree at 325976f, venv built from its `requirements.txt`) → **`call_graph.json`: 7 edges, `parse.report.json` `errors: []`** — the parser→merge→analyzer-output path (including this PR's import-cleaned modules) runs clean end-to-end.
2. **The three explicit-alias re-exports verified live on master** (the #482-class from-import traps this PR adjudicated):
   - `experiment.py:47` — `from core.analysis_core import _normalize_result as _normalize_result  # noqa: F401`
   - `utilities/finding_verifier.py:79` — `from .agentic_enhancer.agent import MAX_TOOL_RESULT_CHARS as MAX_TOOL_RESULT_CHARS  # noqa: F401`
   - `core/analysis_core.py:31` — `from core.verdict_taxonomy import SEVERITIES as SEVERITIES  # noqa: F401`
3. **Gate-firing positive control**: a deliberate unused `import json` in `utilities/llm/providers/openai.py` → ruff **fires** (1 error); reverted → clean.
4. **Full suite**: `python -m pytest tests/ -q --ignore=tests/report` → **3586 passed / 0 failed**.

Combined with #496 (the installed-SDK contract floor, merged): the import-hygiene surface is now machine-guarded — the floor re-derives the SDK field names the adapters depend on, on every run.

PR #500, opened 2026-09-04
test(infra): the bare-open class retired — ast conformance scan + SIM115 backstop

### LINKED ISSUE #481: The review-bot comment corpus across 128 merged PRs: 88 comments, three infra follow-ups — all retired (items 1-2 via #483; item 3 via #500)

gadievron:
## What this fixes

Closes #481's remaining test-tree half (the prod half landed in #484, the config/conftest half in #483). The review-bot corpus sweep that filed this issue found no unaddressed behavioral finding — its yield was three infrastructure classes, and this PR retires the last one: the **not-always-closed class**, i.e. hand-rolled `open()` blocks in the tests tree that sometimes missed their close. (Claim-ok: the class's 11-comment count is issue #481's own census, quoted from its body.)

Three parts:

A note on item 3's either-arm wording ("a `write(path, text)` helper or uniform `.write_text`"): neither literal arm shipped — an initially-built helper fixture was dropped as zero-adopter dead infra, and `with open(p, "w") as fh:` remains a common (managed, closed) idiom. What retired the class is its actual property — no unmanaged handles anywhere in the tree, enforced by the two arms below. The adjacent encoding sibling (`encoding=` unspecified) is deliberately out of scope here and disclosed in the residuals.

1. **Every inline bare open is converted** to the pathlib idiom — `Path.read_text(encoding="utf-8")` / `Path.write_text(..., encoding="utf-8")` — 16 files (`git diff origin/master..HEAD --name-only | grep -c 'tests/test_\|tests/parsers/'` → 17 incl. the ZipFile fix below), including the six shared parser-test `_graph()` helper sites and the two `zipfile.ZipFile(wheel).namelist()` handles in `test_installed_layout.py`, now `with`-managed. The `call_graph.json` reads are now encoding-symmetric with the utf-8-forced writes the prod side enforces.

2. **A structural conformance scan locks the retirement** (`tests/test_issue481_bare_open_conformance.py`, new): an AST-based sweep over the tests tree that fails if an unmanaged open ever comes back. Managed = the outermost call of a `with`-item's context expression (walrus/await unwrapped; argument-nested opens flag — the `with` closes its object, not the handle handed to it); `os.open` is exempt (fd-based); the project's `open_utf8` wrapper and every non-`os` `.open` are covered. AST rather than regex because the tree embeds `open(` inside string fixtures as parser input data, which a regex scan needs a scrubber to survive. 25 self-identifying teeth pin the scanner semantics, plus 4 test functions (the sweep, its extracted scope rule unit-pinned on a planted tree, and the two config pins) — `pytest -q tests/test_issue481_bare_open_conformance.py` → `29 passed`; the sweep's RED evidence: the recorded run (taken mid-conversion, with the first ten files already staged) caught the six remaining `_graph()`-helper sites — the pristine-base census is **23 non-exempt sites across 16 files** under the original scanner — **25 across 17** once the constructor gate the final scanner ships is included (the two `zipfile.ZipFile` handles it adds are the very sites this PR converts) — coverage 25 → 0, confirmed by two independent census oracles. The sweep also covers pytest-collected logic OUTSIDE `tests/` (the root `conftest.py`, the `parsers/*/test_pipeline.py` harnesses — where ruff and the tests-tree-only sweep previously had zero reach), and constructor-style openers (`zipfile.ZipFile`, `os.fdopen`, …) are the scan's teeth — ruff's SIM115 covers only the tempfile family.

3. **The lint backstop joins the config**: `SIM115` in the whole-tree ruff `select`. Its actual blind spot (from its implementation) is the return statement — which is precisely how the six `return json.load(open(...))` helpers escaped it — plus non-Call `.open` receivers; the scan closes both. The fixture data joins the per-file-ignores (`tests/efficacy/fixtures/**`, `tests/fixtures/**` — a deliberate bare open lives at `efficacy/fixtures/webapp/src/app_a.py:15`), and the scan derives its exemption globs from those same pyproject entries so the two gates cannot drift apart. A **whole-table allowlist pin** asserts the exact ruff config shape — any future config key, whatever ruff names it next, trips the pin consciously instead of silently disarming an arm.

## Why the gates are shaped this way

- **Two arms, not one:** ruff's SIM115 is real for assigned opens (`f = open(...)`) but blind to return-shapes and Name/BinOp `.open` receivers; the AST scan covers those. The pin keeps SIM115 from silently disappearing (or the ruff arm from being disarmed via `ignore`/`exclude`/`fix-only`/a new key) — all must-tripped with route-specific messages.
- **Collected logic is never exempt:** `test_*.py` / `*_test.py` / `conftest.py` are scanned even under an exempt glob — a planted test file under fixture data is logic, not data.
- **Known limits are documented in the scan's docstring** (name-based detector limits, the constructor openers invisible to both arms — zero such sites today), rather than claimed away.

## Verification

- Full suite: `3805 passed, 33 skipped` at the first commit; the final state (the scope-rule pins) re-verified `3674 passed, 168 skipped` in a venv-less worktree (the skip delta is environmental — tests that need the repo-local `.venv`; CI's matrix is the arbiter) (venv synced to the requirements pins; `pip install` churn three times this session was the #496 contract test correctly catching the local venv lagging the merged renovate bumps).
- `ruff check .` whole-tree clean under the new config; the prod tree has zero SIM115-flaggable sites (`utilities/file_io.py`'s `open_utf8` return-shape is ruff-exempt by design — the sanctioned wrapper).
- SAST (re-run over the final state, incl. every amend line): semgrep clean on the changed files; CodeQL 601/601 scanned → 0 findings in changed files (3 pre-existing intentional vulns live in the efficacy benchmark fixtures); semgrep over the changed files clean.
- Go build + vet clean (no Go surfaces touched; parity check).
- No behavioral change: the 16 conversions are I/O-idiom-only; every converted test's assertions are unchanged, and the suite green on the union base (every merged PR's regression tests included) is the union-state check.

<details>
<summary>Residuals (disclosed, not fixed here)</summary>

- CI installs ruff unpinned; `SIM115`'s return-statement exemption and opener list are what the whole-tree policy leans on — a future ruff release could shift coverage. The allowlist pin catches config edits but not ruff behavior changes; the scan arm is unaffected.
- `os.popen` and subprocess pipe handles are the remaining handle-returning openers neither arm sees (zero sites today; named in the scan's known-limits paragraph). Aliased imports (`import os as _os`, an opener imported under another name) are the detector's named name-based limits.
- `utilities/file_io.py:223`'s `return open(...)` (the `open_utf8` wrapper) is ruff-exempt by the return-statement rule — fine today; a `# noqa: SIM115` at that line is the pre-authorized remedy if a ruff release ever flags it.
- The adjacent idiom-debt classes are deliberately out of scope here and filed separately: the unspecified-encoding sibling (#504 — text I/O without `encoding=`, a Windows/cp1252 latent class) and the `tempfile.mkdtemp()` accumulation class (#505).
</details>

### DIFF
--- a/libs/openant-core/pyproject.toml
+++ b/libs/openant-core/pyproject.toml
@@ -55,12 +55,25 @@ target-version = "py311"
 # parser import blocks + the core/utilities refactor residue, re-export-
 # adjudicated with 0 cross-file traps) was cleaned in the same commit the
 # scope expanded (the #481 prod half — the #483 test half preceded it).
-select = ["F821", "F811", "F823", "F401", "F841"]
+# SIM115 (open-without-with) flags unmanaged open-family calls EXCEPT in
+# return statements, behind ExitStack/enterContext, immediately .close()d,
+# or inside __enter__ — the return shape and Name/BinOp ``.open`` receivers
+# are its blind spot; the tests-tree conformance scan in
+# tests/test_issue481_bare_open_conformance.py closes those.
+select = ["F821", "F811", "F823", "F401", "F841", "SIM115"]
 
 [tool.ruff.lint.per-file-ignores]
-# fable (prod-cleanup panel, the #483 carryover): the efficacy fixtures are
-# BENCHMARK DATA — the oracle scores them; style rules never touch them.
-"tests/efficacy/fixtures/**" = ["F401", "F841"]
+# NOTE: the SIM115 entries below are also consumed (via fnmatch) by the
+# AST conformance sweep in tests/test_issue481_bare_open_conformance.py —
+# keep slash-containing path forms; a basename-only pattern matches for
+# ruff (basename) but not the sweep (full relative path).
+# The efficacy fixtures are BENCHMARK DATA — the oracle scores them; style
+# rules never touch them (a deliberate bare open lives at
+# tests/efficacy/fixtures/webapp/src/app_a.py:15 and must stay).
+"tests/efficacy/fixtures/**" = ["F401", "F841", "SIM115"]
+# The sample repos are parser INPUT DATA — arbitrary target-repo code the
+# language parsers consume, not this project's own test logic.
+"tests/fixtures/**" = ["SIM115"]
 
 [tool.hatch.build.targets.wheel]
 packages = [
--- a/libs/openant-core/tests/parsers/javascript/test_arrow_const_collision.py
+++ b/libs/openant-core/tests/parsers/javascript/test_arrow_const_collision.py
@@ -8,6 +8,7 @@ that function and its calls. Other emit paths use a first-wins skip guard.
 import json
 import os
 import tempfile
+from pathlib import Path
 
 from core.parser_adapter import parse_repository
 
@@ -22,7 +23,7 @@ def _functions(files: dict):
     out = tempfile.mkdtemp()
     parse_repository(repo, out, language="javascript", processing_level="all",
                      skip_tests=True, name="r")
-    return json.load(open(os.path.join(out, "call_graph.json")))["functions"]
+    return json.loads((Path(out) / "call_graph.json").read_text(encoding="utf-8"))["functions"]
 
 
 def test_arrow_const_does_not_drop_block_function_sink():
--- a/libs/openant-core/tests/parsers/javascript/test_class_expression_accessors.py
+++ b/libs/openant-core/tests/parsers/javascript/test_class_expression_accessors.py
@@ -8,6 +8,7 @@ getGetAccessors()/getSetAccessors().
 import json
 import os
 import tempfile
+from pathlib import Path
 
 
 from core.parser_adapter import parse_repository
@@ -23,7 +24,7 @@ def _functions(files: dict):
     out = tempfile.mkdtemp()
     parse_repository(repo, out, language="javascript", processing_level="all",
                      skip_tests=True, name="r")
-    return json.load(open(os.path.join(out, "call_graph.json")))["functions"]
+    return json.loads((Path(out) / "call_graph.json").read_text(encoding="utf-8"))["functions"]
 
 
 def test_class_expression_getter_emitted_as_unit():
--- a/libs/openant-core/tests/parsers/javascript/test_constructor_reachability.py
+++ b/libs/openant-core/tests/parsers/javascript/test_constructor_reachability.py
@@ -9,6 +9,7 @@ constructor only within the same file; a cross-file `new X()`, `new Map()`, and 
 import json
 import os
 import tempfile
+from pathlib import Path
 
 from core.parser_adapter import parse_repository
 
@@ -23,7 +24,7 @@ def _cg(files: dict):
     out = tempfile.mkdtemp()
     parse_repository(repo, out, language="javascript", processing_level="all",
                      skip_tests=True, name="r")
-    return json.load(open(os.path.join(out, "call_graph.json")))
+    return json.loads((Path(out) / "call_graph.json").read_text(encoding="utf-8"))
 
 
 def _edges(cg, caller_suffix):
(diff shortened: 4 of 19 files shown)


gadievron:
**Gate-session review receipt (the /pr-review gate — run by the maintainer session, independent of the authoring session)**

**Re-derived independently:** the RED replay on the pristine base (`ab18c3f^` = 4219011) fails exactly the 3 load-bearing tests (the sweep + both config pins) — the author's mid-conversion RED evidence replays; a planted bare open in `test_agreement_filter.py` flips the sweep RED on the head (gate-firing control, reverted); full venv suite **3625/0** on the merge-commit head; `ruff check .` clean under the new config; the fable census independently reconciles (23 sites/16 files → 25/17 with the ZipFile pair → **0**); leak scrub over diff+bodies 0; the encoding conversions verified error-preserved (`errors="ignore"` sites), newline-preserved (CSV `newline=""` sites untouched), no binary-site conversions.

**Fable gate: NOT-REFUTED** (no code blocker; no evasion class with live sites). Opus gate: died on a tooling permission mid-run — its completed checks were all confirmations (whole-tree ruff clean; its isolated SIM115 probe caught exactly the deliberate fixture open at `efficacy/fixtures/webapp/src/app_a.py:15` that the per-file-ignores correctly exempt).

**Folded by the gate session** (commit 9ea1cfd, clearly labeled): `venv`/`env` joined `_ENV_OR_BUILD_DIRS`; the false sibling-scanner-parity docstring sentence corrected; the two gate-surfaced zero-site blind classes (first-class `open` references; the fd-returning openers) added to the known-limits paragraph.

**Body-accuracy corrections (the body stays the author's — these are deltas for the reader):**
1. The file-count arithmetic: the cited grep pattern also matches the new scan file — expect **18**, not 17, at the head (16 converted + `test_installed_layout` + the scan file).
2. "The two gates cannot drift apart" overstates — the **exact-map pin** prevents drift, not the derivation (the scan's own docstring says the sets differ; drift would be a conscious edit tripping the pin).
3. "No behavioral change / I/O-idiom-only" — assertion-observable behavior is unchanged, but 23 sites now decode **explicit utf-8** where the locale decided before: a real (positive) change on the `windows-latest` leg.
4. The body's "SIM115 covers only the tempfile family" is the pre-`d3a7a46` wording — the shipped docstring (lzma/fileinput too) is the corrected one.

**Written SKIPPEDs (per the process):** RED→GREEN analog = the conformance test's pristine-base replay (executed, above) + the gate-firing control; real-data differential = **SKIPPED(tests-only diff; no prod path touched — the suite green on the union base is the differential)**; $2 e2e scan = **SKIPPED(tests-only; no prod mechanism changed — the #483/#484 receipts already carry the live-scan legs for the tree)**.

CI 17/17 on the final head (incl. the explicit `windows-latest` legs — the leg this PR's encoding pinning most affects). Merge-ready under the standing directive.

Issue #509, opened 2026-09-05
Verify-stage disagreement corrected to 'inconclusive' is counted as 'safe' in the aggregate metrics (scan.report.json / CLI envelope) — the ->inconclusive arm of the #374/#381 family


gadievron:
Found by this window's glm-5.3 full-stack e2e (the `--verify` run) — a live cross-artifact inconsistency with a false-clean shape on the summary channel.

## The artifact proof (one run, `--verify --llm-reachability`)

Same run, three artifacts disagree:

- `results_verified.json`: findings **safe 4, inconclusive 2** (two command-injection sinks Stage-2 reclassified to `inconclusive`), verdicts SAFE 4 / VULNERABLE 2
- `pipeline_output.json`: safe 4, inconclusive 2 (consistent)
- `scan.report.json` metrics: **safe 6, inconclusive 0**, `stage2_disagreed: 2`

Two unmitigated command-injection sinks that Stage 2 itself called "one caller away from being critical, remotely exploitable" read as *safe* in the aggregate consumers (the CLI summary + the envelope) — anyone consuming only the summary sees a clean report.

## The mechanism

- `core/verifier.py` `_count_verification_outcomes` (the `agree=False` branch): a corrected finding of `vulnerable`/`bypassable` → `confirmed_vulnerabilities` (the #374/#381 fix — pinned by `tests/test_famreport_disagreed_vuln_counted_safe.py`); **every other corrected finding — including `inconclusive` — lands in `counts["disagreed"]`**.
- `core/scanner.py:1111`: `safe = analyze_result.metrics.safe + verify_result.disagreed` — so a unit the verifier explicitly could NOT confirm (corrected to `inconclusive`) inflates `safe`, and `inconclusive` stays at the analyze-stage count (the reclassification escapes the counters).

`inconclusive` is not `safe` — it is the "kept for triage" state (disclosure-eligible in the taxonomy). Counting it safe is the same under-reporting class the vulnerable→bypassable arm fixed; this is the uncovered `->inconclusive` arm.

## The fix direction

Bucket the disagreement by its corrected finding: `inconclusive` → its own counter threaded into `metrics.inconclusive` at the scanner site (a corrected-to-`safe` downgrade keeps folding into `safe` — the PR #69 F5 rationale — and corrected-to-`protected` is recorded as its own open question for the maintainers). Extend the pinned test with the `->inconclusive` arm.

gadievron:
[...]

Issue #268, opened 2026-08-18
llm-reachability stage: wrap the post-read stage body so non-LLM failures degrade instead of aborting the scan


gadievron:
The `llm-reachability` stage's `try/except` guards only the dataset read (`libs/openant-core/core/scanner.py:467-485`); the rest of the stage body runs inside `step_context`, which re-raises — so a non-LLM failure there aborts the whole scan instead of degrading the way `enhance` and `verify` do.

Concretely, the exposed paths are:
- a corrupt/truncated `call_graph.json` → unguarded `read_json` at `core/parser_adapter.py:472`, inside the post-LLM re-filter (the same `JSONDecodeError`/`UnicodeDecodeError` class the dataset-read guard was added for);
- a failed `write_json` (disk-full / permissions) at `core/scanner.py:527` / `:704`.

Provider errors are **not** affected: `analyze_reachability` already skips failed batches and continues, re-raising only `LLMAuthError` by design (`core/llm_reachability.py:391-404` (excerpt — 9 interior lines elided, 3 of them comments)).

**Fix:** wrap the whole stage body the way `enhance` (`scanner.py:744`) and `verify` (`scanner.py:859`) do — record a skip and continue — and correct the stale comment at `scanner.py:473-481`, which currently mis-describes this as "a provider error there still aborts the scan" (provider errors are exactly what's already handled).

Follow-up to #17 (the stage itself shipped and works).

---

## Prior art (added to the body 2026-08-22): PR #257 deferred exactly this, and named the wrong failure class

This report should have opened by citing **PR #257** (merged), which lists under *"Known follow-ups,
out of scope for this PR"*:

> - **the llm-reachability LLM call (`analyze_reachability`) is not wrapped**, so a
>   provider error (429/529/empty completion) there still aborts the whole scan with no
>   artifacts — the higher-frequency failure. The right fix wraps the whole stage body in
>   the same warn→`_record_skip`→continue pattern `enhance`/`verify` use; deferred to keep
>   this PR scoped to the file-read + skip-recording change.

So the fix direction proposed above is not a new suggestion — **it is the one #257 already chose and
deferred**, which should make it cheaper to accept.

**But #257's description of the failure is wrong, and that matters for prioritisation.** It names
provider errors (429/529/empty completion) as the exposed class. They are not exposed:
`core/llm_reachability.py:391-404` re-raises only `LLMAuthError` and swallows everything else —

```python
        except LLMAuthError:
            ...
            raise
[...]

PR #514, opened 2026-09-05
chore(core): the two re-adjudicated dead-code findings — the reachability handler's dead store + the coverage alias wiring

### LINKED ISSUE #507: Two real-but-cosmetic dead-code findings from the review-bot corpus (re-adjudicated): scanner.py's dataset dead store; _TEST_FILES_SKIPPED_KEY unused
Re-adjudication of the #481 corpus (against current master) found two true positives the original sweep had filed as false-positive/stale. Both are non-behavioral no-ops (consequence-triaged; the "no unaddressed behavioral finding" headline stands) — this issue is the cleanup rider:

1. `core/scanner.py:907` — `dataset = None` is the last statement of the outer #268 wave-catch handler; no read of the local follows (the sibling inner handler at :595 IS read at :608 — that one is live; only the outer is dead). Flagged by the review bot on PR #371.
2. `_TEST_FILES_SKIPPED_KEY` (defined `core/scanner.py:1388`) — `git grep` finds exactly one line in the whole tree: its own definition. Zero readers, and :1392 still duplicates the literal it was meant to replace. Flagged on PR #405.

Commands: `git grep -n _TEST_FILES_SKIPPED_KEY origin/master --` → 1 hit (the definition); token scan of the scanner function shows no `dataset` read after :907.

gadievron:
## What this fixes

Closes #507 — two dead-code findings re-adjudicated from the review-bot corpus (both behavior-neutral, verified):

1. **The dead store** — the outer llm-reachability exception handler's trailing `dataset = None` (`core/scanner.py`): the last live read of the variable is the pre-exception `write_json` call; nothing after the handler reads it. The LIVE sibling store (inside the dataset-read guard, read by the `if dataset is not None:` check) is untouched. AST receipt at the pre-fix state: every remaining store has a load after it; the deleted one had none.
2. **The orphan constant** — `_TEST_FILES_SKIPPED_KEY` had zero readers while `_TEST_FILES_SKIPPED_ALIASES` duplicated its literal. The alias tuple now reads the constant (same value, same position; the snake-case-first order the consumers' comment requires is preserved). The report's public output key — `"test_files_skipped"` in `_read_coverage_stats` — is a **separate role** that happens to share the spelling and deliberately stays a literal: it is the report schema consumed by the summary prompt and pinned by `tests/test_issue307_coverage_fields.py` (8/8 green).

## Verification

- `pytest tests/ -q -p no:cacheprovider` → `3648 passed, 168 skipped` (worktree without the repo-local `.venv`; the skip delta is environmental — CI's matrix is the arbiter), including the coverage-fields pin above.
- `ruff check .` → clean (note: F841 cannot catch a rebinding, which is exactly why this dead store survived lint and needed the review-bot finding).
- The diff is exactly two hunks in one file: `git diff --stat` → `1 file changed, 1 insertion(+), 2 deletions(-)`.

## Notes for the reviewer

- Zero behavioral delta by construction: one deleted constant-RHS store with no readers; one same-value literal→constant substitution.
- The `#509`-shaped inconclusive-aggregation lines that appear near the handler on this branch's base are the concurrent upstream fix, not part of this diff.

### DIFF
--- a/libs/openant-core/core/scanner.py
+++ b/libs/openant-core/core/scanner.py
@@ -1498,7 +1498,7 @@ def _collect_coverage(result: ScanResult) -> dict:
         **counts,
         **examples,
         # #307: the dominant exclusion, per-language attributed
-        **({"test_files_skipped": test_files} if test_files else {}),
+        **({_TEST_FILES_SKIPPED_KEY: test_files} if test_files else {}),
         "languages_without_coverage_data": sorted(set(without_data)),
         # #307 (review finding): the languages whose test-file exclusion is
         # UNCOUNTERED — the reader must see that "no entry" means unknown,
--- a/libs/openant-core/core/scanner.py
+++ b/libs/openant-core/core/scanner.py
@@ -904,7 +904,6 @@ def scan_repository(
                 # Record the crash so the degraded reachability pass is
                 # visible in the artifacts, matching the dataset-read guard.
                 _record_skip(result, "llm-reachability", "failed")
-                dataset = None
 
         collected_step_reports.append(
             _load_step_report(output_dir, "llm-reachability")
@@ -1395,7 +1394,7 @@ _TEST_FILES_SKIPPED_KEY = "test_files_skipped"
 # JavaScript's scanner still writes camelCase testFilesSkipped (the same
 # naming drift the 2026-08 CHANGELOG fixed for symlinks_skipped) — read it
 # as an alias until the parser is renamed.
-_TEST_FILES_SKIPPED_ALIASES = ("test_files_skipped", "testFilesSkipped")
+_TEST_FILES_SKIPPED_ALIASES = (_TEST_FILES_SKIPPED_KEY, "testFilesSkipped")
 _COVERAGE_EXAMPLE_KEYS = ("symlink_examples", "unreadable_examples")
 # Parsers disagree on the filename: the in-process Python parser writes
 # scan_result.json (singular); the subprocess parsers write scan_results.json.


gadievron:
**Gate-session review receipt (the full process — re-derivation + fable gate + opus hunt + the astra/evidence-tier adjudication; the purpose lens: reachability safety and finding-visibility)**

**Re-derived independently:** AST store/load walk on `scan_repository` pre-fix finds exactly ONE dead store (:907, the deleted one; the three live stores each have subsequent loads) and post-fix NONE; the use-before-def hazard cleared by dominance analysis (both inner-try arms bind `dataset` before the outer handler can run; a `BaseException` path exits before any read); suite green + `ruff check .` clean on the fix tree; the pin test 8/8; the diff verified as exactly two hunks (+1/−2); #510's merged hunk sits ~200 lines away (no interaction); leak scrub clean.

**Fable gate: NOT-REFUTED** — the deletion matches the function's own convention (the app-context/enhance handlers don't null their locals; downstream reads disk via `active_dataset_path`); the #268 restore-invariants intent is carried by the units_count/status/summary/skip lines, not the dead store.

**Astra/evidence-tier adjudication: MERGE-AS-IS + the amendment (applied as gate-commit 56cd26d).** The tier-grade table's key cells:
- deadness: **proven** (AST def-use, no escapes, empty orelse/finalbody, no enclosing loop)
- over-seed safety: **proven** — the degraded pass leaves the scan over-seeded via the DISK dataset (:936/:1005); the pruned rebuild never persists; the atomic write_json covers the persist window
- the "absence ≠ zero" attribution doctrine: **proven intact** (a byte-identical `_collect_coverage` differential shows `languages_without_test_skip_data` preserved)
- the deleted store's "fail-loud protection" value: **half-wrong** (the file's dominant `is not None` idiom yields silent-skip, not loud failure — corrected in the record)
- the traceback/frame-introspection caveat: **immaterial** (the handler swallows; stdlib prints no locals; zero introspection hits in prod)
- the "report schema, pinned by test_issue307" rationale for the emit-site literal: **REFUTED** (the pin is on the emitted VALUE; the sibling keys :1458→:1493 are already constant-derived) → the **amendment applied**: the emit site now reads `_TEST_FILES_SKIPPED_KEY` (byte-identical output, pin 8/8, +44 tests green on the conformance/file_io pair) — retiring the orphan-constant finding completely (the canonical name flows end to end, two readers)

**Receipt corrections (the body stays the author's):** (1) the emit-site literal sits in `_collect_coverage` (:1500 pre-fix), not `_read_coverage_stats` — the body's role-attribution is off by one function; (2) the schema-pin rationale for the literal is superseded by the amendment; (3) "zero behavioral delta by construction" — proven for *semantics*, overclaimed as worded: a marginal GC/retention delta exists on the degraded path (the rebuilt dict survives to function exit) — non-semantic, disclosed here; (4) a local suite count is not a valid receipt in this repo (~135 tests swing 168/33 with flag/order-dependent skip behavior) — CI is the arbiter.

**A new out-of-scope finding (present identically on master, filed as #516):** the narrow post-`write_json` units_count divergence window — pre-existing #268-class, for the maintainer to triage.

CI green on the final head incl. both windows-latest legs. Merge-ready under the standing directive.

Issue #504, opened 2026-09-05
Tests-tree text I/O without explicit encoding= (787 sites) — a Windows/cp1252 latent class; prod's own open_utf8 reader unused by tests


gadievron:
Found by the #481 deep investigation (adjudicated census, counts cited with commands):

- `grep -rn` + AST census over the tests tree: **787 text-I/O sites without `encoding=`** (494 `write_text` + 220 `read_text` + 73 `open(`) vs 71 with it. Sharpest slice: **59 write-mode `open(p, "w")` sites across 49 files** have no `encoding=` on the open line (command: `grep -rn 'open(' --include='*.py' tests/ | grep -E '"w"' | grep -vc 'encoding='` → 59; with → 1).
- CI runs the Python matrix on **windows-latest** (`.github/workflows/test.yaml:37`), where text-mode I/O without `encoding=` resolves to the ANSI code page (cp1252) — any fixture round-tripping non-ASCII content is platform-dependent by construction. Today's fixtures are ASCII, so this is latent, not live.
- Production has the canonical reader — `utilities/file_io.py:214 open_utf8` + `read_json` (33 prod files use it) — but the tests tree consumes it in exactly two files (its own unit test + the conformance test's name reference). The canonical-read idea never propagated to tests.
- Related asymmetry: #500's conversions pin `encoding="utf-8"` on the reads they convert while the same helper functions keep `with open(p, "w")` (no encoding) for writes.

Candidate direction (the maintainers' call): either propagate `open_utf8`/explicit encoding to tests, or add an encoding tooth/gate (PLW1514 or an AST conformance extension — #500's `test_issue481_bare_open_conformance.py` is the natural home; note its prod-side sibling `test_file_io.py` already enforces encoding= for non-test code, so tests are currently the uncovered cell of that map).

gadievron:
Closing as not-planned, with the reasoning on record (a two-axis adjudication with independent re-derivation):

1. **The hazard is latent, not live**: no bare-written non-ASCII payload exists in the tests tree today (every non-ASCII payload path goes through explicit encoding or the utf-8-forced json helpers).
2. **It self-retires**: CI runs Python 3.14; PEP 686 makes UTF-8 the default at 3.15 — the next CI bump dissolves the class.
[...]

PR #517, opened 2026-09-06
test(parsers): the shared mkdtemp helper family migrated to TemporaryDirectory (#505)

### LINKED ISSUE #505: tempfile.mkdtemp() in tests: 116 unmanaged sites across 89 files (~3 with cleanup) — permanent tmp accumulation; tmp_path already the majority idiom
Found by the #481 deep investigation (AST census over the tests tree, counts with commands):

- **116 unmanaged `tempfile.mkdtemp()` sites across 89 files**; cleanup census: only **3 files** mention rmtree/addFinalizer/atexit (`git grep -n rmtree -- libs/openant-core/tests` → 3 hits, one per file). ~113 sites leak a directory permanently — and in this tree they are not empty: the parser-test helpers write a synthetic repo **plus a full `parse_repository` output** into each one, so every local test run grows $TMPDIR by ~one parsed repo per case.
- `ruff`'s SIM115 does not cover mkdtemp (open-family rule), and #500's conformance scan does not either — this class has **zero gates**; the review-bot corpus produced 0 comments on it.
- The managed alternative is already the tree's majority idiom: **198 files** use pytest's auto-cleaned `tmp_path` (`git grep -l tmp_path -- libs/openant-core/tests | wc -l` → 198).
- Same structural shape as the bare-open class #500 retired, one notch worse (never cleaned at any scope).

Candidate direction: migrate the `parse_repository`-helper family (16 files share the copy-pasted helper) to `tmp_path`, and/or gate future mkdtemp with an AST conformance tooth. Deliberately filed as a follow-up for triage, not bundled into #500.

gadievron:
## What this fixes

Closes #505 (stage 1 of its ruling): the 13 copy-pasted `parse_repository` test helpers that allocated a synthetic repo **plus** a parse-output dir via `tempfile.mkdtemp()` — and never cleaned either — now scope both under `with tempfile.TemporaryDirectory()`, so every test case cleans up deterministically instead of leaking a full parsed repo into `$TMPDIR` per run.

**Scope** (per the issue's adjudication): the helper family selected by `rg -l '^(\s*)out = tempfile\.mkdtemp\(\)' tests/parsers/` → **13 files** (5 js, 2 zig, 2 python, php/ruby/rust/swift ×1). Deliberately excluded: `python/test_function_extractor_robustness.py` (its `_repo()` returns a live path that callers build against after return — a `with` would delete the files first), the ~100 remaining in-body mkdtemp sites, and any ban-gate (a stage-2 tooth was explicitly ruled out — this is dev-machine hygiene, not correctness).

## Shape details

- `os.path.realpath` preserved where it existed (the macOS `/var`→`/private/var` fix); `Path(_repo)` for the two pathlib-dialect helpers; no `ignore_cleanup_errors` (zero of the 31 existing `TemporaryDirectory` sites in-tree use it, including a parser-run precedent that is green on `windows-latest`).
- `encoding="utf-8"` folded into the same write lines the conversion rewrites (11 `open(p, "w", encoding="utf-8")` + 2 `write_text(content, encoding="utf-8")`) — the #504-adjacent slice, riding here because these exact lines were being rewritten anyway.
- The read-side `Path.read_text` lines in the 6 js/zig files are #500's already-merged idiom, untouched here.

## Verification

- **The fix works**: a probe running one converted helper then comparing `$TMPDIR` before/after → `LEAKED_DIRS: 0` (pre-fix, each helper left 2 dirs).
- **Behavior parity**: full suite `3813 passed, 33 skipped` — identical result pre/post (the diff is cleanup-scope only; every assertion unchanged). Note for reviewers: running only `tests/parsers/` as a subset reproduces a **pre-existing** suite-order pollution failure set (12 tests, identical pre/post, recorded since 2026-09-01 as the known dir-ordering artifact) — it does not appear in the full run or CI.
- `ruff check .` clean.

### DIFF
--- a/libs/openant-core/tests/parsers/javascript/test_arrow_const_collision.py
+++ b/libs/openant-core/tests/parsers/javascript/test_arrow_const_collision.py
@@ -14,16 +14,16 @@ from core.parser_adapter import parse_repository
 
 
 def _functions(files: dict):
-    repo = os.path.realpath(tempfile.mkdtemp())
-    for rel, content in files.items():
-        p = os.path.join(repo, rel)
-        os.makedirs(os.path.dirname(p), exist_ok=True)
-        with open(p, "w") as fh:
-            fh.write(content)
-    out = tempfile.mkdtemp()
-    parse_repository(repo, out, language="javascript", processing_level="all",
-                     skip_tests=True, name="r")
-    return json.loads((Path(out) / "call_graph.json").read_text(encoding="utf-8"))["functions"]
+    with tempfile.TemporaryDirectory() as _repo, tempfile.TemporaryDirectory() as out:
+        repo = os.path.realpath(_repo)
+        for rel, content in files.items():
+            p = os.path.join(repo, rel)
+            os.makedirs(os.path.dirname(p), exist_ok=True)
+            with open(p, "w", encoding="utf-8") as fh:
+                fh.write(content)
+        parse_repository(repo, out, language="javascript", processing_level="all",
+                         skip_tests=True, name="r")
+        return json.loads((Path(out) / "call_graph.json").read_text(encoding="utf-8"))["functions"]
 
 
 def test_arrow_const_does_not_drop_block_function_sink():
--- a/libs/openant-core/tests/parsers/javascript/test_class_expression_accessors.py
+++ b/libs/openant-core/tests/parsers/javascript/test_class_expression_accessors.py
@@ -15,16 +15,16 @@ from core.parser_adapter import parse_repository
 
 
 def _functions(files: dict):
-    repo = os.path.realpath(tempfile.mkdtemp())
-    for rel, content in files.items():
-        p = os.path.join(repo, rel)
-        os.makedirs(os.path.dirname(p), exist_ok=True)
-        with open(p, "w") as fh:
-            fh.write(content)
-    out = tempfile.mkdtemp()
-    parse_repository(repo, out, language="javascript", processing_level="all",
-                     skip_tests=True, name="r")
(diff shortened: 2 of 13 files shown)


gadievron:
**Gate-session receipt (full process; the orchestrator + fable + the opus trio-hunt; e2e legs run)**

**Re-derived independently:** the leak probe executed on a clean TMPDIR — pre-fix helper shape leaks **2 dirs/call**, post-fix **0** (matches the body's claim, different method: direct dir-counting around a replicated helper pair); full suite green ×2; ruff clean; with-scope escape audited (all 13 helpers return materialized json INSIDE the with — nothing escapes); subprocess-cleanup race checked (parse_repository's parser step is `subprocess.run` = waits; the only Popen waits and targets the JS parser dir, not the temp dirs); helper-family completeness re-derived (the body's selector over ALL of tests/ yields the same 13; the one ≥2-mkdtemp file outside is the deferred in-body class).

**E2E (mechanism-tied):** parse **PARITY** — the same corpus through master's tree vs this tree: `call_graph.json` and `analyzer_output.json` **byte-identical**; `functions.json` differs ONLY in `extraction_time` (timestamp; verified key-by-key). The conversion is cleanup-scope-only, proven at the artifact level.

**Opus round folded (gate commit da32e29):** the write side went explicit utf-8 but **7 of the rewritten helpers' `call_graph.json` reads stayed encoding-default** — the locale/cp1252 asymmetry on the windows-latest leg this repo's own `run_utf8` names. All 7 patched; residual `grep` = 0; full suite green.

CI 17/17 (incl. both windows legs). Merge-ready.

Issue #302, opened 2026-08-20
Stage 2 only examines Stage-1 positives (4.1% of units) and can only downgrade — the denominator is printed to stderr but absent from the step reports and SUMMARY_REPORT.md


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

Stage 2 only ever examines Stage 1's positives. On the run that prompted this work that was 223 of
5,475 analyzed units (4.1%); the other 5,252 never receive a second *adjudicating* opinion.

> **Corrected 2026-08-21.** This sentence originally read "the other 5,252 were never re-examined by
> anything." That is false as written. `core/analyzer.py:686` runs
> `run_stage1_consistency_check(results, code_by_route, binding, get_global_tracker())` over **all**
> results, including those 5,252. What that pass does is narrower than adjudication: it groups units
> by signature pattern, skips any group of fewer than two, and sends only groups whose verdicts
> *disagree* to the LLM. So a unit that is alone in its pattern, or agrees with its pattern-mates, is
> never looked at a second time — but "never re-examined by anything" over-states it, and the
> consistency check is a second mechanism this issue should have accounted for. The denominator
> argument below is unaffected. Every completed
Stage-2 disagreement was a downgrade — 34 `vulnerable → safe`, 6 `vulnerable → protected`, 0 upgrades
— which is structural, not incidental: a unit Stage 1 called safe cannot be upgraded because it never
enters Stage 2.

That is a legitimate design (it is a false-positive filter), but nothing in the step
reports or `SUMMARY_REPORT.md` states it, and the product framing "Stage 1 detects. Stage 2 attacks.
What survives is real." reads as though surviving findings were adjudicated against the whole
codebase.

## Evidence

`libs/openant-core/core/verifier.py:104-108` (HEAD `b501962`):

```python
    # Filter to vulnerable/bypassable only
    vulnerable_results = [
        r for r in all_results
        if str(r.get("finding") or r.get("verdict", "")).lower() in ("vulnerable", "bypassable")
    ]
```

On the live run:

```
[...]

PR #518, opened 2026-09-06
fix(verify): an incomplete verification never vetoes a correction — and never drops out of disclosure (#296)

### LINKED ISSUE #296: `_has_conclusive_exploit_path` tests a free-text explanation string, so an incomplete verification suppresses a correction — its mirror guard must NOT be tightened (PR #195/#243)

gadievron:
## What this fixes

Closes #296 — the direction-aware fix the issue was re-scoped to by its own 2026-08-21 correction (**only the first helper changes**; the PR #195/#243 mirror guard stays untouched, per the issue's own constraint).

**The window** (re-derived at the fix site): `_parse_finish_result`'s finish paths — agree-missing and self-contradictory finishes — return `incomplete: True` **while preserving a populated `exploit_path` and a model-authored explanation**. The conclusive-path guard tested only the free-text marker string, so that unfinished evidence counted as a conclusive analysis and **vetoed verdict corrections** at the consistency call site. The structured signal (`verification["incomplete"]`) was already on the same object.

## The fix (two arms, per the adjudication)

1. **The veto lift**: an incomplete verification skips the conclusive-path check — unfinished evidence no longer suppresses corrections. The marker constant stays (it covers rows that predate the structured flag). A *complete* verification's conclusive broken path still vetoes, unchanged.
2. **The disclosure guard**: a correction of an **incomplete** verification into any `DISCLOSURE_DROPPED` verdict (`{safe, protected, inconclusive, rejected}` — the canonical set, imported) is blocked with a `consistency_downgrade_blocked` record carrying `incomplete: True`. Without this arm, lifting the veto alone would open the exact hole the adjudication found: the conclusively-exploitable mirror **cannot catch broken-path shapes** (its return requires a reached sink, attacker control, and no break), so an incomplete broken-path result would pass straight through into a disclosure drop.
3. The `incomplete` flag survives an allowed correction — a consistency opinion is not a completed verification (pinned by test).

## Verification

- **RED→GREEN on the true window** (incomplete + populated **broken** path + model text; route keys following the `*Msg` grouping the pattern detector actually uses):
  - test 1 (a `vulnerable→bypassable` correction, both disclosure-eligible) — RED on pristine master: the veto suppressed it. Green with the fix, and the `incomplete` flag survives (pinned).
  - test 2 (an `inconclusive` proposal, in `DISCLOSURE_DROPPED`) — RED on pristine master (silently blocked, **no audit record**) and RED on fix-arm-1-alone (the disclosure-drop hole). Green only with the full fix, and the block record carries `incomplete: True` (pinned).
- Full suite `3815 passed, 33 skipped` (includes the pre-existing characterization pins for both helpers and the e2e mirror-survival test); `ruff check .` clean.

## Notes for the reviewer

- The consistency apply-loop only fires on **non-conformant** model replies (the shipped prompt never requests `findings_to_update`) — this fix hardens that existing defense layer; it does not activate the unused prompt field.
- A characterization docstring in the test file was updated: it described this defect as open ("explicitly NOT the direction-aware behavior change"); post-fix the helpers are shape-only by design and the incomplete-awareness lives at the call site.

### DIFF
--- a/libs/openant-core/tests/test_issue296_conclusive_guards.py
+++ b/libs/openant-core/tests/test_issue296_conclusive_guards.py
@@ -1,21 +1,19 @@
-"""Regression tests for issue #296 — the conclusive-exploit-path guards
-test a free-text explanation string; the structured signal
-(``verification["incomplete"]``) sits unused on the same object.
-
-Scope of this change (the issue's lower-risk items, explicitly NOT the
-direction-aware behavior change the maintainer raised as an open question):
+"""Regression tests for issue #296 — the conclusive-path guards and the
+incomplete-verification signal, at the call site (the direction-aware fix,
+landed in this PR after the issue's own 2026-08-21 re-scoping):
 
 1. The matched marker is a NAMED constant (``INCOMPLETE_VERIFICATION_MARKER``)
-   with the match set UNCHANGED — widening it is the false-negative direction
-   for BOTH guards and must wait for the direction-awareness analysis.
-2. The ASYMMETRY is pinned (the issue's suggested test 3): an INCOMPLETE
-   verification carrying an INTACT exploit path must still leave
-   ``_has_conclusive_exploitable_path`` True — the downgrade into
-   DISCLOSURE_DROPPED stays BLOCKED. Whatever future fix makes the guards
-   signal-aware, it must not let an unfinished verification become the
-   reason a finding is downgraded out of disclosure.
-3. The guards' current behavior on the marker and on the constructed
-   defect input is characterized so any change forces a conscious decision.
+   — it stays as the legacy arm covering rows that predate the structured
+   flag.
+2. The raw helpers below are SHAPE-ONLY characterizations (both still
+   free-text/marker-driven by design — the incomplete-awareness lives at the
+   consistency CALL SITE, not in the helpers, so the #195/#243 mirror
+   contract is untouched).
+3. The call-site behavior: an INCOMPLETE verification never vetoes a
+   correction (arm 1) and never drops out of disclosure (arm 2 — a
+   correction into any DISCLOSURE_DROPPED verdict is blocked with an audit
+   record carrying ``incomplete: True``); the incomplete flag survives an
+   allowed correction (pinned).
 """
 
 from __future__ import annotations
@@ -83,19 +81,17 @@ def test_asymmetry_incomplete_intact_path_blocks_downgrade():
     assert _exploit_path(v) is False
 
 
-def test_characterization_incomplete_broken_path_is_the_open_defect():
-    """CHARACTERIZATION (the open defect, pinned so any change is a
-    conscious decision): incomplete:True + BROKEN path + model-authored
-    explanation reads as conclusive today, suppressing a consistency
-    correction. #296's traced constraint: the obvious tightening re-opens
-    the disclosure FN via the consistency update at :918 — the fix must be
-    direction-aware (skip only for non-downgrade updates). Changing this
-    assertion without that analysis is exactly what #296 warns against."""
+def test_characterization_incomplete_broken_path_helper_still_marker_driven():
+    """CHARACTERIZATION (the helper's shape contract, unchanged by the fix):
+    incomplete:True + BROKEN path + model-authored explanation still reads
+    as conclusive to the RAW helper — the incomplete-awareness lives at the
+    consistency call site (which skips the helper for incomplete rows), not
+    in the helper itself, so the #195/#243 mirror contract is untouched."""
     v = {"incomplete": True, "explanation": _MODEL_TEXT,
          "exploit_path": dict(_BROKEN)}
     assert _exploit_path(v) is True, (
-        "KNOWN DEFECT (see #296): an incomplete verification's broken path "
-        "currently suppresses corrections — fix requires direction-awareness")
+        "the raw helper must stay marker-driven (the #195/#243 contract); "
+        "incomplete-awareness belongs to the call site")
 
 
 def _verifier():
--- a/libs/openant-core/tests/test_issue296_conclusive_guards.py
+++ b/libs/openant-core/tests/test_issue296_conclusive_guards.py
@@ -129,3 +129,86 @@ def test_e2e_incomplete_intact_path_downgrade_still_blocked():
     got = next(r for r in out if r["route_key"] == rk)
     assert got.get("finding") == "vulnerable", "the downgrade was applied!"
     assert "consistency_downgrade_blocked" in got
+    assert got["consistency_downgrade_blocked"]["incomplete"] is True
+
+
+# ---------------------------------------------------------------------------
+# #296's direction-aware fix (adjudicated round 3): the finish-path window.
+# _parse_finish_result can mark a verification incomplete WHILE preserving a
+# populated exploit_path and model-authored explanation (agree-missing,
+# self-contradictory finishes) — the string-marker test passes and the
+# incomplete evidence vetoes a CORRECTION. The structured signal exists on
+# the same object: verification["incomplete"].
+# ---------------------------------------------------------------------------
+
+def test_finish_path_incomplete_with_populated_path_allows_correction():
+    """#296's fix: an agree-missing finish (incomplete=True, populated
+    BROKEN exploit_path, model text != the marker) must NOT count as a
+    conclusive exploit path — pre-fix it did (the broken-path conclusive
+    check vetoed), and the unfinished evidence suppressed the correction.
+    Route keys follow the *Msg grouping the pattern detector uses; the
+    proposal (vulnerable -> bypassable) stays inside DISCLOSURE_ELIGIBLE,
+    so the disclosure-drop guard is not the thing under test here. The
+    incomplete flag SURVIVES the correction: a consistency opinion is not
+    a completed verification."""
+    from utilities.finding_verifier import ConsistencyCheckResult
+    v = _verifier()
+    rk = "app/api/handler.go:serveMsg.json"
+    sibling = "app/api/handler.go:helperMsg.json"
+    v._resolve_inconsistency = lambda group, cbr: ConsistencyCheckResult(
+        "handler", "bypassable",
+        [{"route_key": rk, "should_be": "bypassable", "reason": "consistent"}], "x")
+    agree_missing_v = {
+        "incomplete": True,  # the structured signal the guard now reads
+        "correct_finding": "vulnerable",  # the group IS inconsistent (sibling says bypassable)
+        "explanation": _MODEL_TEXT,  # model text — not the marker
+        "exploit_path": dict(_BROKEN),  # populated, BROKEN — the finish-path shape
+    }
+    out = v._check_consistency([
+        {"route_key": rk, "finding": "vulnerable",
+         "verification": agree_missing_v},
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (full process — the finding-visibility PR got the deepest round; orchestrator + fable gate + opus hunt + a live wire run)**

**Re-derived independently:** RED replay on pristine master = the 3 new tests fail exactly as claimed (test 2 RED on master AND on arm-1-alone — the disclosure-drop hole is real); GREEN 7/7; suite green; ruff clean; the #195/#243 mirror guard's helper body **token-identical** base↔PR (0 diff lines); `DISCLOSURE_DROPPED` membership vs real spellings verified (normalization at :1162 + the `_CORRECTABLE_STAGE2_FINDINGS` gate at :1179 — no spelling slips; `error` unreachable by design); the changed-row predicate is exactly `bool(verification.incomplete)` — complete rows bit-identical to master.

**Fable gate: NOT-REFUTED** — the blast-radius table: arm 1 adds only vulnerable↔bypassable on broken-path rows (both disclosure-eligible, both kept in confirmed_findings); arm 2 removes drop-ability for every incomplete shape (strictly more findings retained); the #510 interaction is disjoint by the same predicate (incomplete rows bucket to needs_review before disagreed_inconclusive); the `*Msg` grouping matches the real detector.

**Opus round — one P1 adjudicated and REFUTED, the rest folded:** opus claimed arm 2's disclosure premise was false (the incomplete→"unverified" reporter mapping). Re-derived from source: the reporter's disclosure loop iterates `confirmed_findings` (reporter.py:397-399; verifier.py:488-496), which filters `finding ∈ {vulnerable, bypassable}` — a row corrected to `safe` exits the list UPSTREAM and never reaches that mapping; every downstream filter (dynamic_tester, generator) reads `stage2_verdict`, which only exists for rows inside the list. **Arm 2 protects the actual disclosure entry** — the PR's premise stands. Folded as receipt notes: arm 2's scope covers the four degenerate incomplete producers (no exploit_path) where master allowed drops — conservative, correct-direction, now disclosed; the `protected` spelling untested (same predicate); the #302 direction counters skip incomplete rows (pre-existing); the non-dict verification guard (518-F5, same exposure as master).

**E2E (live wire, glm-5.3, --verify):** all steps clean, cross-artifact consistency **PASS**, $0.47. This run's Stage-2 disagreed-and-corrected both sinks to safe (complete rows, the designed #69 path — the #296 window did not fire; LLM nondeterminism, so the durable mechanism proof is the RED→GREEN replay).

**Gate commit 41ad9b0:** the body claimed a characterization docstring was updated — the diff was purely additive (opus 518-F3). The stale prose now describes the shipped design.

CI 17/17. Merge-ready.

PR #519, opened 2026-09-06
fix(dynamic-test): the runtime policy anchored — target-declared runtime first, stable fallback; the Go pin de-rotted (#412)

### LINKED ISSUE #412: Dynamic-test runtime: Python base image is LLM-chosen with no anchor; Go is prompt-pinned to golang:1.25-alpine — decide the runtime policy
## Background

This converts a deferred decision that currently lives only in PR comments (#351's closure + its correction) into a tracked item — the fragile-record class. Investigation on 2026-08-29 re-derived all of the following from current master:

- The `docker_templates/*.Dockerfile` files are **reference documentation only** — no code reads them (`_TEMPLATES_DIR` feeds only `attacker_server.py`; the executed runtime Dockerfile is the LLM-generated `generation["dockerfile"]` at `docker_executor.py:262`; `resolve_docker_template` is used only as a language-support GATE in `__init__.py:46-48`, never for content).
- Consequently the **real** dynamic-test runtime version is decided per-finding by the generating model — and the system prompt (`test_generator.py`) gives **no base-image guidance for Python at all** while **pinning Go to `golang:1.25-alpine`** (lines 66-68). The attacker sidecar is inlined at `docker_executor.py:286` as `python:3.11-slim` (version-agnostic stdlib only, so no capture-semantics impact).

## The question

What should the dynamic-test runtime policy be?
1. **Status quo** — LLM free choice per test for Python/JS; Go pinned in the prompt. Pros: the model can match the target's runtime when the finding gives it a cue. Cons: nondeterministic, unprompted version choice; the Go pin will rot (1.25 today).
2. **Prompt-anchor all languages** (e.g., "prefer a base image matching the scanned target's runtime; when unknown, prefer current stable LTS"). Deterministic-ish guidance, still adaptable.
3. **Configurable per-scan override** (config key the executor injects into the prompt / overrides the Dockerfile FROM line).

Either way the Go prompt pin deserves a bump/refresh decision (`golang:1.25-alpine` vs the 1.27 template reference) so the two stop disagreeing.

Related: #319 (the real defect in this area — ERROR/NOT_REPRODUCED dynamic tests render as "Verified via dynamic testing" in disclosures), #315, #314.

gadievron:
## What this fixes

Closes #412 with **option 2** (the adjudicated decide-and-close shape): the dynamic-test runtime policy is anchored, the Go pin is de-rotted, and a drift guard makes the rot structurally impossible to re-introduce silently.

**The state being fixed** (re-derived at the sites): the executed runtime Dockerfile is the LLM-generated `generation["dockerfile"]` per finding; the system prompt gave **Python no base-image guidance at all** while pinning Go to `golang:1.25-alpine` — and that pin had drifted **three ways**: prompt `1.25` / template `1.27` / reference doc `1.22`.

## The change

1. **The policy** (prompt, all languages): *"prefer a base image matching the scanned target's **declared** runtime"* — `go.mod`'s `go` directive, `.python-version` / `requires-python`, `engines`, `.tool-versions`, or a version cue stated in the finding — *"a wrong runtime can hide version-dependent exploit behavior or break the repro outright. Only when no declaration is visible, fall back to the current stable release."* The Go example in the prompt is that **fallback**, explicitly not an override of a declared version.
2. **The pin refresh**: `golang:1.27-alpine` everywhere the fallback is spelled — both prompt mentions, `docker_templates/go.Dockerfile`, and the reference doc.
3. **The drift guard** (`tests/test_issue412_runtime_anchor.py`, 3 tests, all RED on current master): the prompt's Go tags and the template's `FROM` tag must **match** and live in the test's known-fallback set — a future bump must touch prompt + template + the set in one conscious change; plus a pin on the policy text itself so it cannot be silently deleted.

## Notes for the reviewer

- `docker_templates/` remain **reference-only** (the issue's own finding: no executor reads them) — no executor change here; only the prompt, the doc, and the guard.
- The **config-key option (3) was ruled out**: config surface for a string the model rewrites per finding, with no consumer — real maintenance, no determinism gain.
- FN scope (bounded in the adjudication): a wrong runtime withholds the dynamic **proof** but does not shrink disclosure eligibility — the collector returns a result (ERROR/INCONCLUSIVE/NOT_REPRODUCED), and the report merge preserves `stage2_verdict`. This PR improves repro fidelity; it does not claim full dependency fidelity (the prompt's "don't pin exact versions unless version-specific" dependency rule still applies).

## Verification

- `pytest tests/test_issue412_runtime_anchor.py` → `3 passed` (RED on pristine master: `3 failed` — every test catches a pre-fix defect: the unknown tag, the prompt/template mismatch, the absent policy).

### DIFF
--- a/libs/openant-core/tests/test_issue412_runtime_anchor.py
+++ b/libs/openant-core/tests/test_issue412_runtime_anchor.py
@@ -55,6 +55,28 @@ def test_runtime_anchor_guidance_present():
     """The policy itself is pinned: target-declared runtime first, stable
     fallback only when nothing is declared."""
     src = (_DYNDIR / "test_generator.py").read_text(encoding="utf-8")
-    assert "BASE IMAGE RUNTIME POLICY" in src
-    assert "DECLARED runtime" in src
-    assert "fall back to the current stable" in src
+    assert "BASE IMAGE RUNTIME POLICY" in src, (
+        "the runtime-policy block was removed from the prompt — #412's "
+        "anchor is gone; restore it or update this pin consciously")
+    assert "DECLARED runtime" in src, (
+        "the target-declared-runtime-first clause is gone from the prompt")
+    assert "fall back to the current stable" in src, (
+        "the stable-fallback clause is gone from the prompt")
+
+
+def test_reference_doc_template_table_matches_the_templates():
+    """The reference doc's template table is a THIRD spelling of the
+    template facts — it drifted before (node:20 vs the template's 24) and
+    is bound here now (the #412 audit's find)."""
+    doc = (_DYNDIR / "DYNAMIC_TESTER_REFERENCE.md").read_text(encoding="utf-8")
+    for name in ("python", "node", "go"):
+        tmpl_first = (_DYNDIR / "docker_templates" / f"{name}.Dockerfile"
+                      ).read_text(encoding="utf-8").splitlines()[0]
+        import re as _re
+        m = _re.search(r"FROM\s+(\S+)", tmpl_first)
+        assert m, f"{name}.Dockerfile's FROM line is malformed: {tmpl_first!r}"
+        image = m.group(1)
+        assert f"`{image}`" in doc, (
+            f"DYNAMIC_TESTER_REFERENCE.md's template table does not name "
+            f"{image} (the {name}.Dockerfile base) — the table drifted "
+            "from the templates it describes (the #412 class)")
--- a/libs/openant-core/tests/test_issue412_runtime_anchor.py
+++ b/libs/openant-core/tests/test_issue412_runtime_anchor.py
@@ -1,0 +1,60 @@
+"""#412: the dynamic-test runtime anchors cannot silently drift again.
+
+The system prompt's Go fallback tag and the docker_templates/ reference
+Dockerfile are two spellings of one fact ("the current Go fallback").
+They drifted three ways once (prompt 1.25 / template 1.27 / doc 1.22)
+because nothing bound them. This test parses both at collection time and
+fails on any mismatch — a future pin bump must touch both (and this test's
+expected fallback list) in one conscious change.
+"""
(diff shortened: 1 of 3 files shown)


gadievron:
**Gate-session receipt (full process; orchestrator + opus hunt + the mechanism-driven e2e)**

**Re-derived independently:** RED on pristine master = 3 failed / GREEN 3/3 on the fix tree; the policy text live in SYSTEM_PROMPT (target-declared first, fallback rationale, no stale 1.25 pin); suite + ruff green.

**E2E (mechanism-driven):** the REAL prompt builder driven on a declared-runtime fixture — SYSTEM_PROMPT carries the policy (PASS×3); **and the load-bearing discovery**: `_build_finding_prompt` interpolates only name/language/application_type + the finding's fenced fields — **the declared runtime never reaches the model mechanically** (go.mod/.python-version/requires-python are invisible; only a version cue inside the finding text is reachable). The anchor is real but text-only. Filed as **#521** with the direction (derive declared runtimes into repo_info) + the undisclosed reverse trade-off (a declared OLD runtime + modern deps → build failure → the dynamic proof is blocked where the old hardcoded-modern fallback would have reproduced; `regenerate_test`'s error-feedback loop is the natural precedence mechanism).

**Opus round folded (gate commit a2a376b):** (1) the drift this PR was filed for was **STILL LIVE one line above the fix** — the reference doc's node row said 20-slim vs the template's 24-slim (the #412 audit's own find); corrected. (2) The drift guard now binds the reference doc's template table to ALL THREE templates' actual FROM lines (a third spelling of the facts; 4/4 tests). (3) Assert messages on the policy pin. Known residual disclosed: the attacker-sidecar `python:3.11-slim` pin in docker_executor.py remains unguarded (carried in #521).

CI 17/17 after the lint fix (an unused variable in the first guard-extension draft — caught by the repo's own Lint leg). Merge-ready.

Issue #319, opened 2026-08-22
Dynamic-test results are merged into reports with no status filter, so an ERROR or NOT_REPRODUCED test renders as "Verified via dynamic testing" in the disclosure


gadievron:
## Summary

`merge_dynamic_results` attaches a `dynamic_testing` block to a finding whenever a dynamic-test
result exists for that finding's ID, **with no filter on the result's status**. Both consumers of
that block — the summary table and the disclosure document — key on its **existence**, not on its
status. A finding whose Docker test **errored out** or was **NOT_REPRODUCED** is therefore rendered
as "Verified via dynamic testing" in the externally-facing disclosure.

The block also carries `"tested": "Docker container, <Month Year>"` unconditionally, so a finding
whose test never ran asserts that a Docker container tested it.

## The merge applies no status filter

`libs/openant-core/report/generator.py:112-121`:

```python
    for finding in pipeline_data.get("findings", []):
        fid = finding.get("id")
        if fid and fid in results_by_id:
            r = results_by_id[fid]
            finding["dynamic_testing"] = {
                "status": r.get("status"),
                "details": r.get("details"),
                "evidence": r.get("evidence", []),
                "tested": f"Docker container, {date_str}",
            }
```

`status` is copied into the block but never branched on. `date_str` comes from the results file's
mtime (`:110`), not from any evidence that a container ran.

## Both consumers key on existence, not status

`libs/openant-core/report/prompts/summary.txt:97`:

> `- "Verified" column: "dynamic" if dynamic_testing exists for that finding, ...`

`libs/openant-core/report/prompts/disclosure.txt:52`:

> `Verified via {"dynamic testing" if dynamic_testing else "attacker simulation (Stage 2)" if stage2_verdict in ("confirmed", ...`

Both are truthiness tests on the block. Neither inspects `dynamic_testing["status"]`.

## Executed at `b5019628`

Real `merge_dynamic_results`, `dotenv` stubbed, fixture in a `mktemp -d`:

```
  Merged 3 dynamic test results from dynamic_test_results.json
VULN-001  present=True  status=ERROR           tested=Docker container, August 2026
VULN-002  present=True  status=NOT_REPRODUCED  tested=Docker container, August 2026
VULN-003  present=True  status=CONFIRMED       tested=Docker container, August 2026
```

All three are indistinguishable to a consumer that tests for the block's presence. Reproduce by
writing a `dynamic_test_results.json` with those three statuses next to a `pipeline_output.json` and
[...]

PR #523, opened 2026-09-06
fix(dynamic-test): null optionals in a complete generation are coerced, not discarded (#522)

### LINKED ISSUE #522: generate_test discards valid dynamic-test generations that emit "docker_compose": null (the _STR_FIELDS validation) — sonnet-family replies degrade to NOT_REPRODUCED
Found by a pre-registered experiment driving the real `generate_test` path (utilities/dynamic_tester/test_generator.py) with the via-sonnet binding:

**The bug**: the post-parse validation
```python
_STR_FIELDS = ("dockerfile", "test_script", "test_filename",
               "requirements", "requirements_filename", "docker_compose")
if any(k in parsed and not isinstance(parsed[k], str) for k in _STR_FIELDS):
    return None
```
rejects any generation in which the model emits a null for an optional field — but the system PROMPT ITSELF declares "docker_compose": string | null ("null if single container"), and instructs Go tests NOT to write go.mod (null requirements are the expected emission). A model obeying the prompt's schema is rejected: `generate_test` returns None.

**The consequence (corrected from the original filing)**: generation-None lands as **status ERROR, un-retried** (result_collector.py: `Test generation failed — LLM did not return valid test code`), inflating the errors bucket and burning the full generation cost — not the NOT_REPRODUCED path as first written (the two in-code comments carry the same error; both fixed in the PR).

**Evidence (exact)**: 18 of 21 scored generations in the run carried a null optional and ALL returned None despite complete, parseable JSON with a correct Dockerfile (`python3 -c` over the experiment's ED-rows.json: total=24, generation_returned_None=18, contaminated=3, scored_OK=3). Pre-existing on master (#256 era).

**Fix direction**: the required trio (dockerfile/test_script/test_filename) keeps strict str validation; the three optionals coerce null → "" (the canonical absent form — what result_collector's `.get(k, "")` and DynamicTestResult already default to); non-str non-null optionals stay rejected (a dict/number is still malformed). Both call sites (generate_test + regenerate_test).

gadievron:
## What this fixes

Closes #522 — a complete, schema-obeying dynamic-test generation was being **discarded** whenever the model emitted a null for an optional field.

**The self-inconsistency** (prompt vs validator): the system prompt declares `"docker_compose": string | null` ("null if single container") and instructs Go tests NOT to write go.mod — so a model enumerating the full schema emits nulls. The validator rejected any non-str value for `docker_compose`/`requirements`/`requirements_filename` → `generate_test` returned None → **status ERROR, un-retried** (the full generation cost burned; the errors bucket inflated). Note: the original issue text said NOT_REPRODUCED — corrected in the issue; the two stale in-code comments carried the same error and are fixed here.

**The evidence (exact)**: 18 of 21 scored generations in the discovery run (the real `generate_test` path, anthropic/claude-sonnet-5) carried a null optional and **all** returned None despite complete, parseable JSON with a correct Dockerfile.

## The fix (both call sites: `generate_test` + `regenerate_test`)

- The **required trio** (`dockerfile`/`test_script`/`test_filename`) keeps strict `str` validation — unchanged.
- The **three optionals** coerce `null → ""` — the canonical absent form (what `result_collector`'s `.get(k, "")` and `DynamicTestResult` already default to; keeps null out of checkpoint JSON and the regenerate prompt).
- **Non-str non-null** optionals (a dict/list/number — e.g. a YAML-object `docker_compose`) stay rejected: still a malformed generation.

## Verification

- **RED→GREEN**: exactly the 8 null-acceptance tests fail on the pristine validator; **21 negative guards** pass on both sides (non-str optionals incl. `dict`/`list`/`int`/`bool` at both sites; the required trio strict incl. `None`).
- Full suite `3850 passed, 33 skipped` (29 tests in the new file); `ruff check .` clean.
- Implementation verified against the adjudicated shape (both sites byte-identical logic; coercion before any consumer; the regenerate prompt's `previous_generation` is always a prior validated return).

## Follow-up noted (not in scope)

A dict-valued `docker_compose` (YAML-as-JSON-object) could be recovered via `yaml.safe_dump` — a real emission shape, but an enhancement; belongs to its own issue if wanted.

### DIFF
--- a/libs/openant-core/tests/test_issue522_null_optionals.py
+++ b/libs/openant-core/tests/test_issue522_null_optionals.py
@@ -5,8 +5,9 @@ tests NOT to write go.mod — so a model that enumerates the full schema
 emits ``"docker_compose": null`` (and, for Go findings, null requirements)
 for a complete, valid generation. The old validator rejected any non-str value for those
 fields, discarding the generation to the ERROR path (result_collector
-records generation-None as status ERROR, un-retried): 18/21 real
-sonnet-family generations in the discovery run were dropped this way.
+records generation-None as status ERROR, un-retried in-run): of 21 scored
+generations in the discovery run, 14 carried a null optional and were
+dropped this way (4 more were truncated replies failing at the parser).
 
 The contract now: the required trio stays strictly str; null optionals
 coerce to the canonical absent form "" (what the executor and
--- a/libs/openant-core/tests/test_issue522_null_optionals.py
+++ b/libs/openant-core/tests/test_issue522_null_optionals.py
@@ -1,0 +1,134 @@
+"""#522: null optionals in a COMPLETE generation are not malformed.
+
+The system prompt declares docker_compose as "string | null" and tells Go
+tests NOT to write go.mod — so a model that enumerates the full schema
+emits ``"docker_compose": null`` (and, for Go findings, null requirements)
+for a complete, valid generation. The old validator rejected any non-str value for those
+fields, discarding the generation to the ERROR path (result_collector
+records generation-None as status ERROR, un-retried): 18/21 real
+sonnet-family generations in the discovery run were dropped this way.
+
+The contract now: the required trio stays strictly str; null optionals
+coerce to the canonical absent form "" (what the executor and
+DynamicTestResult already default to); non-str non-null optionals stay
+rejected (a dict/number is still malformed).
+"""
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+import utilities.dynamic_tester.test_generator as tg  # noqa: E402
+
+
+_FINDING = {
+    "id": "f1", "name": "x", "cwe_id": 22, "cwe_name": "Path Traversal",
+    "location": {"file": "a/b.go", "line": 1},
+    "stage1_verdict": "vulnerable", "stage2_verdict": "vulnerable",
+    "vulnerable_code": "func f() {}", "description": "d", "impact": "i",
+    "steps": "s",
+}
+_REPO = {"name": "r", "language": "go", "application_type": "cli_tool"}
+
+_BASE = {
+    "dockerfile": "FROM golang:1.27-alpine\n",
+    "test_script": "package main\nfunc main() {}\n",
+    "test_filename": "test_exploit.go",
+}
+
+
+def _stub_reply(payload: dict):
+    import json
+    captured = {}
+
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (full process — orchestrator re-derivations + the fable gate + a mechanism-tied e2e)**

**Re-derived independently:** RED replay on pristine master = exactly the **8 null-acceptance tests fail** / the 21 negative guards pass on both sides (the required trio strict incl. `None`; non-str optionals incl. dict/list/int/bool rejected at both sites); GREEN **29/29**; full suite **3797/0**; ruff clean; the diff is exactly the two validator blocks + the test file. The **`""`-vs-absent equivalence verified at every consumer**: all four downstream reads are falsy-guarded (`docker_executor.py:272/:273/:290/:424`), `result_collector`'s and `DynamicTestResult`'s `.get(k, "")` defaults match, the checkpoint `to_dict` path matches, the regenerate prompt embeds identically — and the coercion is what MAKES the two forms equal (a present-`None` would pass straight through `.get` — the pre-existing `test_issue431` fixture demonstrates it). In-place mutation safe (parsed is local; mid-loop None-returns leave a dict nobody holds). Status semantics confirmed: generation-None → ERROR, **un-retried in-run** (the regenerate loop sits after the continue; ERROR checkpoints re-attempt on resume). Rendering untouched (no report path reads the optionals; the #319 CONFIRMED-only contract intact).

**E2E (mechanism-tied):** the REAL `generate_test` driven with a full null-optional reply → returns **a generation** (not None), all three optionals coerced to `""`, the required trio strict, and all four executor consumer-guards skip exactly like absent keys.

**Fable gate: NOT-REFUTED** — with one catch folded (gate commit 3e7c43e): the **18/21 count did not survive re-derivation**. From the discovery-run artifact: 14 of the 18 GENERATION-FAILED rows carry complete JSON + a null optional (this PR's bug); **4 are truncated replies failing at the parser** — before the validator, not recoverable here. The source comments and test docstring now carry the defensible 14-of-21 with the truncated-4 caveat (a count baked into source must re-derive — it didn't).

**Receipt notes:** the PR body's suite count (3850/33) and the gate's (3797/33 post-fold) differ by scope/venv — CI is the arbiter; the 18/21 in the *issue* body stands corrected by the PR's own text but the issue-side figure is historical record (left as-is with the PR as the correction); a `_validate_generation()` shared-helper extraction is the right follow-up for the duplicated 24-line block (this PR already had to fix the same stale comment twice — the drift argument is real).

CI green incl. both windows legs. Merge-ready.

Issue #521, opened 2026-09-06
Dynamic-test declared-runtime policy has no mechanical input channel — and the agree-missing finish can vanish a Stage-1 vulnerable from the artifacts


gadievron:
Two findings from the 2026-09-06 gate round on #517/#518/#519 (both re-derived from source; both pre-existing on master, not regressions from those PRs):

## 1. The declared-runtime policy (#519's anchor) is prompt-text-only

`test_generator.py`'s new policy says "prefer a base image matching the scanned target's **declared** runtime — the `go` directive in go.mod, .python-version / requires-python, engines, .tool-versions, or a version cue stated in the finding". But `_build_finding_prompt` (test_generator.py:135) interpolates only: repo **name**, **language**, **application_type**, and the finding's own fenced fields. `repo_info` never carries the target's declared runtimes — **the model cannot prefer what it never sees**; only the fifth source (a version cue inside the finding text) is actually reachable. Direction: mechanically derive the declared runtimes (go.mod / .python-version / requires-python / engines) into `repo_info` and interpolate one line per language into the prompt.

Related undisclosed trade-off (same round): when a target DECLARES an old runtime (e.g. go 1.21) but the exploit needs a modern dependency, `go mod tidy` can fail to build → the dynamic proof is BLOCKED where the old hardcoded-modern fallback would have reproduced. A precedence rule (declared runtime wins; on build failure regenerate with the stable fallback — `regenerate_test` already feeds errors back) should be stated in the prompt.

## 2. The agree-missing finish can vanish a Stage-1 vulnerable (one layer upstream of #296)

`finding_verifier.py` `_verify_one:890` writes `verification.correct_finding` onto `finding` for **all** incomplete paths. For the agree-missing finish path (`:1486-1489`), `correct_finding` is **model-supplied** — an agree-missing finish with `correct_finding: "safe"` yields `finding="safe", incomplete=True`. That row then:
- leaves `confirmed_findings` (core/verifier.py:492 filters on `finding ∈ {vulnerable, bypassable}`) → never reaches the reporter, **not even as "unverified"** — a Stage-1 vulnerable vanishes from the disclosure artifacts on an *unfinished* verification (the exact direction #296 exists to prevent, one layer upstream);
- counts `needs_review` in metrics (so the metrics/artifact disagreement is observable).

[...]

Issue #286, opened 2026-08-19
Verify checkpoints: errored units are restored as completed on resume, because `_cp_is_error` tests a value nothing writes on the normal path


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

`FindingVerifier`'s resume path decides whether a checkpointed unit needs re-verification with
`_cp_is_error`, which returns True only when the checkpoint has no `verification` dict, or when
`verification["correct_finding"] == "error"`. The verify error path writes neither: it writes
`verification = {"incomplete": True}` and puts the error string on the *result* dict, which the
checkpoint writer does not copy. `correct_finding == "error"` has **no writer on the normal verify
path**. Every errored unit therefore looks like finished work on resume and is never retried.

> ⚠️ **Corrected 2026-08-22.** This paragraph originally read "no writer *anywhere in production
> code*". That is false — `finding_verifier.py:882`/`:921` assigns `correct_finding` from an
> unvalidated LLM field on the Stage-1 consistency path. The value is **unwritten on the path that
> produced these records, not unwritable**. Full correction at the end of this report; the measured
> 223/223 adoption result is unaffected.

## Evidence

`libs/openant-core/utilities/finding_verifier.py:603-630` (excerpt — 6 interior lines elided) (HEAD `b501962`):

```python
        def _cp_is_error(cp_data):
            """A verify checkpoint is errored if verification is missing/empty
            or correct_finding == 'error'."""
            if not cp_data:
                return True
            v = cp_data.get("verification", {})
            if not v:
                return True
            return v.get("correct_finding") == "error"

        # Separate already-done (successful) from to-do (new + errored)
        results_to_verify = []
        _restored_ok = 0
        for r in results:
            key = r.get("unit_id") or r.get("route_key", "unknown")
            cp_data = checkpointed.get(key)
            if cp_data and not _cp_is_error(cp_data):
                # Restore verification data from checkpoint
                ...
                _restored_ok += 1
            else:
                # Either no checkpoint, or an errored one — re-verify
                results_to_verify.append(r)
```

[...]

PR #526, opened 2026-09-07
fix(verify): an incomplete finish never authors a downgrade — the agree-missing severity rule (#521, finding 2)

gadievron:
## What this fixes

Finding 2 of #521: a `finish` call **without `agree`** marked the verification incomplete but still honored the model-supplied `correct_finding` — so a supplied `"safe"` overwrote a Stage-1 `vulnerable`, `_verify_one` wrote it back, and the row **silently left disclosure entirely** (verifier's `confirmed_findings` admits only `vulnerable|bypassable`) while the metrics still counted it `needs_review`: a row in limbo, not a correct drop. (#518 could not cover it: the consistency pass gates on ≥2-row inconsistent groups.)

## The rule (one site, shared by both incomplete-producing branches)

Whether the finish was **agree-missing** or the **FAM-REPORT-2 self-contradiction** (`agree=True` + diverging verdict), an unfinished verification never authors a downgrade:

- `correct_finding = _more_severe(original_finding, normalized-supplied)` — upgrades honored, downgrades withheld; null/blank/garbage supplied falls back to the Stage-1 verdict; the returned verdict normalized.
- The withheld supplied verdict is **recorded verbatim** on the verification (`withheld_correct_finding`), so the triager sees what the unfinished verification proposed — and both withholdings (this rule's and the consistency pass's #518 block) are visible on one row.
- The **complete disagree** path (`agree=False` present) is untouched — the legitimate downgrade channel the disclosure invariants preserve.
- A **taxonomy conformance pin**: `FINDING_VERDICT_ORDER[:2]` IS the eligible set and the rest are all disclosure-dropped — the `_more_severe` authority's load-bearing property, previously an unpinned coincidence (a display-motivated reorder would have re-opened the vanish hole at both call sites).

## Verification

- **RED→GREEN**: 8 behavioral RED on pristine master (the vanish window, the withheld record, the normalization, the null/garbage/error fallbacks, the FAM-downgrade record) + a pre-fix-shape vanish sanity through the real code path.
- The FAM-REPORT-2 suite and the #296 guards unchanged (19 passed) — the complete-disagree channel and the contradiction behavior are pinned intact.
- Full suite `3863 passed, 33 skipped`; `ruff check .` clean.
- Pre-fix checkpoints keep the defect on resume (rows are restored verbatim, not retried — the #286 design); a fresh verify heals them. Disclosed rather than fixed (scope).

## Finding 1 of #521 (the mechanical runtime channel)

Deliberately NOT in this PR — it is a scope decision (narrow the prompt to reachable sources vs. build the manifest-derivation channel) recorded on the issue for the maintainer.

### DIFF
--- a/libs/openant-core/tests/test_issue521_agree_missing_severity.py
+++ b/libs/openant-core/tests/test_issue521_agree_missing_severity.py
@@ -1,0 +1,172 @@
+"""#521 finding 2: an incomplete finish never authors a downgrade.
+
+Agree-missing (`finish` without `agree`) marked the verification incomplete
+but still honored the model-supplied ``correct_finding`` — so a supplied
+``safe`` overwrote a Stage-1 vulnerable, the row left disclosure entirely
+(core/verifier.py's confirmed_findings admits only vulnerable|bypassable),
+while the metrics still counted it needs_review. The rule, shared with
+FAM-REPORT-2's self-contradictory path: on ANY incomplete finish, the
+surfacing verdict is the MORE-SEVERE of (Stage-1, supplied) per
+FINDING_VERDICT_ORDER — upgrades honoured, downgrades withheld and recorded
+(``withheld_correct_finding``).
+"""
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from core.verdict_taxonomy import (  # noqa: E402
+    DISCLOSURE_DROPPED,
+    FINDING_VERDICT_ORDER,
+    SEVERITY_FINDING_VERDICTS,
+)
+from utilities.finding_verifier import FindingVerifier  # noqa: E402
+
+
+def _finish(supplied, original="vulnerable", agree_absent=True):
+    # _parse_finish_result references no self.* — unbound-safe via __new__
+    # (the famreport2 test file's own pattern)
+    v = FindingVerifier.__new__(FindingVerifier)
+    fr = {"correct_finding": supplied, "explanation": "model text"}
+    if not agree_absent:
+        fr["agree"] = False
+    return v._parse_finish_result(fr, original, 2, 100)
+
+
+# --- the core rule ---------------------------------------------------------
+
+def test_agree_missing_supplied_safe_keeps_vulnerable():
+    """RED on master: the vanish window — Stage-1 vulnerable, agree-missing
+    finish supplying 'safe' — must keep 'vulnerable' + incomplete."""
+    v = _finish("safe")
+    assert v.correct_finding == "vulnerable", "the downgrade was authored"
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (full process, finding-visibility depth — orchestrator re-derivations + a two-channel drive + the fable gate, which originated this finding in its #518 round)**

**Re-derived independently:** RED replay on pristine master = 11 failed / 1 pass (the taxonomy pin — expected green there); GREEN **12/12** on the fix tree; the pinned neighbors (FAM-2 + #296 + #509 suites) **32/32 untouched**; full suite **3810/0** post-fold; ruff clean. **The two-channel drive** (the real `_parse_finish_result`): complete-disagree (agree=False + safe) → `correct_finding=safe, incomplete=False, withheld=None` — **the legitimate downgrade channel intact**; agree-missing + safe → `vulnerable, incomplete=True, withheld=safe`, serialized — **and the row ADMITS to `confirmed_findings`** (the three-way agreement restored: disclosure "unverified", metrics needs_review, artifact finding=vulnerable). Fable's own trace confirms the full path (admission → reporter "unverified" → needs_review) plus a bonus: a supplied `"error"` no longer becomes `correct_finding`, closing master's would-be perpetual-retry shape (`_cp_is_error`).

**Fable gate: NOT-REFUTED** — the vanish hole closed exactly as derived; the FAM-REPORT-2 branch share verified per-shape (identical output except the withheld record AND the now-normalized honored-upgrade — strictly safer; every downstream reader lowercases anyway); the `agree: null`/`agree: "false"` sibling hole identified as a **pre-existing follow-up** (a non-bool agree reads as a complete verdict — the same family one step over; `agree_missing = not isinstance(...)` is its fix shape).

**The gate fold (commit 4a3f118):** the "end-to-end vanish sanity through the real code path" test was **VACUOUS** — it hardcoded `finding="vulnerable"` into the row instead of mirroring `_verify_one`'s write-back, so it passed on pristine master and could not detect the vanish it was named for. Now `row["finding"] = v.correct_finding` — **executed both sides: RED on master, 12/12 on the fix tree.**

**Receipt notes (not folded):** falsy non-null garbage (0/False/[]/{}) collapses to the null-branch and isn't recorded verbatim (a non-string-garbage gap in the record contract); the taxonomy pin anchors to `SEVERITY_FINDING_VERDICTS` — one hop short of the literal admission tuples at the three gate sites; `withheld_correct_finding` reaches `results_verified.json` + checkpoints only (consistent with the existing `consistency_downgrade_blocked` surface); the PR body's "8 behavioral RED" is 6 behavioral + the new-field AttributeErrors; the suite count reconciles at 3810/33 with the fold included. Pre-fix checkpoints keep the defect on resume (the #286 verbatim-restore design — disclosed, correct).

CI green on the folded head incl. both windows legs. Merge-ready.

Issue #290, opened 2026-08-19
Verify calls `adapter.complete` with `max_tokens=4096`, bypassing the `simple_text` default raised in #242


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

PR #242 raised `simple_text`'s default `max_tokens` from 8192 to 20000, with a comment naming the
exact failure mode it was fixing: a thinking-era model consumes the output budget on hidden reasoning
and returns a reasoning-only (empty) completion, which the adapter rejects as "no usable content".
Stage 2 verification does not go through `simple_text`. It calls `binding.adapter.complete` directly
with a module constant of 4096, so the raised default never reached `verify` on any route.

## Evidence

`libs/openant-core/utilities/finding_verifier.py:74` (HEAD `b501962`):

```python
MAX_TOKENS_PER_RESPONSE = 4096
```

`utilities/finding_verifier.py:399-411` — the agent loop, which bypasses `simple_text` entirely:

```python
        while iterations < MAX_ITERATIONS:
            iterations += 1

            self._log("debug", f"Iteration {iterations}", iterations=iterations)

            # Adapter handles the rate-limiter wait/report dance internally.
            response = self.binding.adapter.complete(
                model=self.binding.model,
                max_tokens=MAX_TOKENS_PER_RESPONSE,
                system=system_prompt,
                tools=self._tool_defs,
                messages=messages,
            )
```

`libs/openant-core/utilities/llm/helpers.py:41-49` — what PR #242 (`a991416`, merged 2026-08-16,
two days before this run) put in place:

```python
    # Thinking-era default. Claude-5 / Gemini-2.5+ / OpenAI o-series spend
    # output budget on hidden reasoning; 8192 can be fully consumed by
    # reasoning on a large unit, yielding a reasoning-only (empty) completion
    # that the adapter drops -> hard "no usable content" error. 20000 is under
    # the Anthropic non-streaming 10-min ceiling (32000 is rejected with a
    # "Streaming is required" ValueError; 20000 is accepted) and is a CAP, not
[...]

Issue #209, opened 2026-08-03
Report step returns status: success while writing SUMMARY_REPORT.md as 0 bytes


NahumKorda:
**Summary** 

The report phase completes, self-reports success, generates all per-finding disclosure documents — and writes report/SUMMARY_REPORT.md as a zero-byte file. Nothing in the returned status or error list indicates the primary human-readable deliverable is missing. 

**Evidence** 

report.report.json from the run: 

{ "step": "report", "status": "success", 
 "summary": { "formats_generated": ["summary_path", "disclosures_dir"] }, 
 "outputs": { "summary_path": ".../report/SUMMARY_REPORT.md" }, 
 "errors": [] } 
 
summary_path appears in both formats_generated and outputs; errors is empty. On disk the file is 0 bytes. The sibling disclosures_dir was populated correctly (413 documents), so only the summary failed. 

**Why it matters** 

This is a silent success: the tool asserts it produced an artifact it did not produce. It is the same failure class as the recent hardening series (#178–#196) and the two silent consistency-override false negatives closed in #195 — a step reporting success while its output is absent or empty. A consumer that trusts status never discovers the report is missing, and in a pipeline the empty file propagates downstream as if valid. 

**Suggested fix** 

stat the file after writing and fail the step when it is absent or zero-length; only append a path to formats_generated once verified non-empty. A test asserting os.path.getsize(summary_path) > 0 on a fixture scan would catch regressions. 

 
_Observed on one full-pipeline run at production scale: a private TypeScript monorepo (Angular front end, NestJS back end), 1,351 analysis units from 866 source files, run with --verify and no --limit, ~6.5 hours wall clock. The engine was invoked directly (python -m openant scan …, Python 3.13); the Go CLI was not built, so none of this involves the Go↔Python envelope. Stage 1 detection and Stage 2 verification ran on Claude Opus 5; application context, enhancement and reporting on Claude Sonnet 5 — all seven phases routed through OpenRouter (anthropic/claude-opus-5, anthropic/claude-sonnet-5) via a locally-added openrouter provider adapter, on a build based on upstream 2ed78f6. Cost figures are expressed as proportions of the run total; absolute amounts are omitted deliberately, as is the identity of the scanned repository._

gadievron:
Fix opened in #279 (pending review + CI). This issue will close automatically when that PR merges.

gadievron:
Reopen request — the 0-byte half is genuinely fixed; the `status: success` half still ships.

[...]

PR #527, opened 2026-09-07
fix(llm): truncation is never silent — the warn-once helper, the named parse failure, the in-deliverable banners (#512)

### LINKED ISSUE #512: Reasoning models exhaust the fixed output budgets: app-context max_tokens=2000 and report max_tokens=4096 produce fence-only/empty completions on glm-5.3
Found by this window's glm-5.3 full-pipeline e2e (the `--verify` runs):

- `context/application_context.py:794` sets `max_tokens=2000` for the threat-model/app-context call; `report/generator.py:407,:586` set `max_tokens=4096`.
- glm-5.3 (a reasoning model on the chat-completions path) spends the budget on reasoning tokens and returns a fence fragment or an empty completion: the runs show app-context `skipped` with exactly `output_tokens: 2000` ("Failed to parse LLM response as JSON … Response: \`\`\`'"), report `partial` with exactly `4096`, and one `OpenRouterAdapter returned an empty completion`. Recurred in 2 of 3 full runs; the pipeline survives (the #213 fallback context + the report's partial-state handling), but the app-context/report value silently degrades for any reasoning model.

Direction for the maintainers: a per-phase (or per-model-class) output budget, or routing reasoning models through a reasoning-aware path (e.g. OpenRouter's reasoning-effort parameters / the Responses API), or at minimum a loud hint when `output_tokens == max_tokens` (truncation-detection) instead of a JSON-parse failure. Related: the via-glm53 config profile used in the e2e (all phases z-ai/glm-5.3 @ openrouter).

gadievron:
## What this fixes

Closes #512 — reasoning models exhaust fixed output budgets and the failure was **silent**: app-context (`max_tokens=2000`) and the report phases (`max_tokens=4096`) spent the whole budget on hidden reasoning, returned a fence fragment or an empty completion at exactly the cap, and surfaced as a bare `JSONDecodeError` (app-context `skipped`) or a quietly short summary/disclosure — no signal of the cause anywhere.

## The fix (the adjudicated shape, refuted-and-final)

1. **The detector is `stop_reason == "max_tokens"`** — the signal the adapter already normalizes and tests — *not* `output_tokens == max_tokens` (a full reply that happens to equal the cap is not truncation; locked by test).
2. **`simple_completion`** — the typed-result sibling of `simple_text` (same call, same tracking, returns `CompletionResult`) for callers that need `stop_reason`; `simple_text` keeps its exact signature (the #290 `inspect.signature` pin) and return, and now **warns once per (phase, model, cap)** on truncation — lock-guarded (it runs in thread-pool workers), resettable for tests, worded for the providers that map *unknown* stop reasons to `max_tokens`.
3. **App-context names the cause**: the parse-failure `ValueError` carries `stop_reason` and states the truncation explicitly; the private 2000 cap is **dropped** (the default). The cap never protected cost where it mattered — the OpenAI Responses path floors output to 16000 anyway (`_RESPONSES_MIN_OUTPUT_TOKENS`), and Stage-1's per-unit calls already run at the default.
4. **The report phases**: the two private 4096 caps → `DEFAULT_MAX_TOKENS` (the #290 drift class, now pinned by source-scan tests — no numeric `max_tokens=` literal in either file); a **truncated (non-empty) summary/disclosure carries a WARNING banner in the deliverable itself** (stderr alone is lost in CI and quiet runs); the #209 empty-output guard folds `stop_reason` into its `RuntimeError` when the empty text is the reasoning-only truncation.
5. The three existing test monkeypatch sites migrated to the new stub shape.

## Verification

- New battery: `10 passed` — warn-once per key (+ different cap = different signal), end-turn-at-cap emits nothing (the detector lock), the app-context named failure, banner present/absent for summary + disclosure, the empty+truncated `RuntimeError`, both no-numeric-cap source pins.
- The migrated suites (`test_appcontext_llm_unknown_key`, `test_fence_live_sites_escape`, `test_issue290_output_budget`) green — the #290 signature pin intact.
- **RED note (honest)**: on pristine master the new test file fails at collection — `simple_completion` is a new API — so the behavioral RED is the appctx message shape and the banners, which are absent pre-change by construction.

## Scope notes

- **Artifact surfacing**: the once-per-key stderr warning + the in-deliverable banners are the shipped surfaces. Riding the truncation count into `scan.report.json` per-phase (the symlink-skip warning family) is recorded as the follow-up — it needs the tracker to grow a warnings surface, which is its own change.
- **Residual private caps** (recorded, not swept here): `cli.py:1265`, `html_report.py:46`, `stage1_consistency.py:46`, `threat_model_agent.py:40`, `repo_explorer.py:63`, `context_enhancer.py:352`, `llm_reachability.py:451`, `json_corrector.py:157` (deliberately pinned, test-locked), `test_generator.py:278,:357`.
- Per-phase/per-model-class budgets and reasoning-aware routing were ruled **out of fixes-only scope** (capability); this PR removes the *silent* part.

### DIFF
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -33,7 +33,7 @@ from typing import Any
 
 from dotenv import load_dotenv
 from utilities.file_io import open_utf8, read_json, read_repo_file, write_json
-from utilities.llm import PhaseBinding, simple_text
+from utilities.llm import PhaseBinding, simple_completion
 
 load_dotenv()
 
@@ -788,10 +788,21 @@ def generate_application_context(
         f"Generating context with {binding.provider_name}/{binding.model}...",
         file=sys.stderr,
     )
-    response_text = simple_text(
+    # #512: the typed-result sibling (simple_completion) so the parse-
+    # failure error can NAME the truncation when that is the cause; and
+    # the private 2000 cap is dropped (DEFAULT_MAX_TOKENS — the cap was
+    # never protecting cost where it mattered: the OpenAI Responses path
+    # floors output to 16000 anyway, and Stage-1's per-unit calls already
+    # run at this default. A 2000 cap on a reasoning model spends the
+    # whole budget on hidden reasoning and returns a fence fragment —
+    # exactly the failure the parse error below then mislabels).
+    _result = simple_completion(
         binding,
         CONTEXT_GENERATION_PROMPT.format(sources=sources_text),
-        max_tokens=2000,
+    )
+    response_text = "\n".join(
+        block.text for block in _result.content
+        if hasattr(block, "text")
     )
 
     # Extract JSON from response
@@ -811,7 +822,17 @@ def generate_application_context(
     try:
         data = json.loads(json_str)
     except json.JSONDecodeError as e:
-        raise ValueError(f"Failed to parse LLM response as JSON: {e}\nResponse: {response_text}")
+        # #512: name the truncation when that is the cause — a bare
+        # "could not parse" on a max_tokens reply hid the real failure
+        # (the budget went to hidden reasoning, not the answer).
+        trunc_note = ""
+        if _result.stop_reason == "max_tokens":
+            trunc_note = ("; the reply was TRUNCATED at max_tokens — the "
+                          "budget was spent on hidden reasoning, not the answer")
+        raise ValueError(
+            f"Failed to parse LLM response as JSON: {e}\n"
+            f"(stop_reason={_result.stop_reason}{trunc_note})\n"
+            f"Response: {response_text}")
 
     data['source'] = 'llm'
 
--- a/libs/openant-core/report/generator.py
+++ b/libs/openant-core/report/generator.py
@@ -15,6 +15,7 @@ from core.verdict_taxonomy import DISCLOSURE_ELIGIBLE
 from .schema import validate_pipeline_output, ValidationError
 from utilities.file_io import normalize_results, open_utf8, read_json
 from utilities.llm import (
+    DEFAULT_MAX_TOKENS,
     PhaseBinding,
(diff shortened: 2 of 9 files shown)


gadievron:

**Re-derived independently:** the 10-test battery green ×2; the migrated suites + the #290 signature pin 20/20; suite 3689/0; ruff clean; the detector lock verified (`stop_reason=='max_tokens'`, not output==cap); the source pins re-derived (`grep max_tokens=[0-9]` in both files → **0 hits**; `DEFAULT_MAX_TOKENS` present); `simple_text`'s #290-pinned signature intact.

**Fable gate: NOT-REFUTED on code, one overclaim — FOLDED (gate commit d6bea1d):** the body's item 4 claimed the empty-completion case was covered, but the EMPTY reasoning-only truncation raises at the **adapter's empty-content guard** — the only layer that sees the stop reason — with a message naming neither truncation nor the reason. Folded per the gate's own 3-line shape: the anthropic guard now carries `stop_reason={raw_stop!r}`, the openai guard `finish_reason={raw_finish!r}` (google already names the class in prose), + a source-scan pin test. Touched suites 86/86; battery 11/11.

**Receipt notes:** the warn-once state is bounded (finite phases × config-finite models × literal caps); the print-inside-lock diverges from the codebase's warn-once idiom (cosmetic contention); the set is process-global (multi-scan embedders warn once per process — a reset helper exists); **my own framing corrected by the gate**: app-context does NOT run at the provider default — it runs at `DEFAULT_MAX_TOKENS=20000` (a fixed cap; no cost-cap exists anywhere in the tree, pre-existing); the banner cannot be *suppressed* by finding text (prepended before any LLM-controlled bytes), only forged (benign false alarm — the same trust model as the existing headers); the parse-failure `ValueError` still embeds the full response text (up to 10× larger worst case at the default cap).

PR #528, opened 2026-09-07
fix(scanner): a post-persist llm-reachability failure leaves units_count at the disk truth (#516)

### LINKED ISSUE #516: The llm-reachability handler's units_count restore has a narrow post-write divergence window
Found by the astra/evidence-tier adjudication pass on PR #514's dead-store deletion (a NEW derivation, out of that PR's scope, present identically on master):

Between `dataset` being persisted at scanner.py:854 (`write_json(active_dataset_path, dataset, ...)`) and the `result.units_count = pre_llmreach_units_count` restore at :897, the stage body mutates `result.units_count` mid-flow (:885) — if a non-LLM failure fires in that window (e.g., the re-filter raises), the restore at :897 fixes the count, but only because the handler runs. The DELETED dead store (`dataset = None`, PR #514) was never load-bearing here (verified: AST store/load walk, disk-mediated downstream). The residual window: a failure AFTER :854's write but BEFORE :885's mutation would leave `result.units_count` describing the PRE-stage count while the disk dataset describes the POST-stage one — the two invariants the #268 handler guards — and nothing reconciles them on that narrow path.

This is pre-existing on master (identical pre/post #514), narrow (the window is a few lines), and behavior-identical to #268's original finding class. Filed for the maintainer to triage: either extend the #268 test to cover the window, or document that the mid-flow mutation at :885 is the designed restore point.

gadievron:
## What this fixes

Closes #516 — the narrow post-persist divergence window in the llm-reachability handler.

**The window** (re-derived at the fix site): between `write_json(active_dataset_path, dataset)` and the handler's `result.units_count = pre_llmreach_units_count` restore, the stage mutates the count mid-flow (post-re-filter). A failure in the post-write bookkeeping (the `ctx.summary` build indexes the applied-signals summary — a real `KeyError` shape) hit the #268 handler, which **restored the PRE-stage count over a dataset that had already persisted** — leaving `units_count` describing a dataset that no longer exists on disk, while downstream (enhance/analyze) consumes the **disk** (`active_dataset_path`, never the local).

## The fix

- A `dataset_persisted` flag: `False` before the try, set `True` immediately after the atomic write (`write_json` is temp+fsync+`os.replace` — after it returns, disk IS the post-filter state).
- The #268 restore is now **gated on the persist**: restore the pre count only when nothing persisted; leave the post-filter count (== the units on disk) when the dataset landed.
- The skipped step report records `dataset_persisted`, so a degraded-but-persisted stage never reads as "no effects" while downstream consumes the re-filtered dataset.

## Verification

- **RED→GREEN**: the discriminating test injects the failure genuinely *after* the persist (the applied-signals summary missing the key the ctx build indexes — a real `KeyError`, not a mock-side raise): pre-fix the count reported **4 against a 3-unit disk**; post-fix **3 == 3**.
- The #268 pre-write guard suite and the re-filter-loop suite unchanged (11 passed) — the pre-write restore path is pinned intact.
- `units_count` consumers enumerated (scan StepReport, `ScanResult.to_dict`, the CLI envelope's `total_units`) — nothing reads it between the stage and the report, and nothing else changed.

### DIFF
--- a/libs/openant-core/core/scanner.py
+++ b/libs/openant-core/core/scanner.py
@@ -604,6 +604,14 @@ def scan_repository(
             # and that one MUST still abort (bad credentials must not be
             # silently skipped into a "successful" degraded scan).
             pre_llmreach_units_count = result.units_count
+            # #516: whether the re-filtered dataset reached disk. write_json
+            # is ATOMIC (temp + fsync + os.replace) — after it returns, the
+            # disk IS the post-filter dataset and downstream consumes the
+            # DISK (enhance/analyze read active_dataset_path, never this
+            # local). A post-write bookkeeping failure must therefore leave
+            # units_count describing the post-filter state, not restore the
+            # pre-stage count over a persisted dataset.
+            dataset_persisted = False
             try:
                 if dataset is not None:
                     app_ctx_payload = None
@@ -852,6 +860,7 @@ def scan_repository(
                     # Persist final dataset so downstream stages see promoted
                     # entry points, per-unit signals, and the applied filter.
                     write_json(active_dataset_path, dataset, indent=2)
+                    dataset_persisted = True
 
                     ctx.summary = {
                         "units_reviewed": pre_filter_count,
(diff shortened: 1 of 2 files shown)


gadievron:

**Re-derived independently:** RED replay on pristine master = 2 failed (a genuine post-write `KeyError` via the applied-signals summary — the real window, not a mock raise) / GREEN 2/2; the #268 + re-filter suites 9/9 (the pre-write restore path pinned intact); suite 3681/0; ruff clean. **The window closure verified at the atomic level**: `write_json`'s `os.replace` is the last statement of its try (`file_io.py:293`), the flag set immediately after (`scanner.py:863`) — nothing can throw between. `units_count` consumers enumerated and confirmed complete (the scan StepReport, `to_dict`, the CLI envelope's `total_units` — nothing between the stage and the report).

**Fable gate: NOT-REFUTED.** Receipt note: on the summary-build failure path the step report reads `skipped, dataset_persisted: true` without `signals_path` although signals were written (the `ctx.outputs` build is never reached) — disclosed-odd, not wrong.

CI green. Merge-ready.

PR #529, opened 2026-09-07
fix(report): the dev-only Python HTML report pins and SRIs its CDN scripts (#502)

### LINKED ISSUE #502: The Python html report (report/html_report.py) still loads Chart.js from the CDN — the #332 self-containment contract covers only the Go-rendered reports
Found while completing the chart.js 4.4.7→4.5.1 vendored bump (#493, the 4-model panel round):

## The gap

`libs/openant-core/report/html_report.py:425-426` loads its chart dependencies from the CDN:

```
<script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
<script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-datalabels@2"></script>
```

and its own docstring says "Internet connection for Chart.js CDN". The #332 self-containment work (vendored + sha-pinned + go:embed, enforced by `report_vendor_test.go` + `report_issue332_test.go`) covers only the **Go-rendered** reports (`openant report -f html` via the gohtml templates + `openant serve`). The no-CDN guard (`TestRenderedReportHasNoCDNScripts`) scans the Go templates only — this Python surface can regress/floating-load forever and no test notices.

## What it means

- The Python-side HTML report (a different surface from the Go renderer) is not air-gap/offline-capable, loads **floating** chart.js majors (no version pin at all), and carries no SRI.
- Not a new regression — pre-existing since before #332; the vendoring work scoped itself to the Go surface.

## Suggested direction

Either retire the Python html path in favor of the Go renderer (if nothing consumes it distinctly), or vendor the same pinned blobs for it (the #332 recipe applies: blob + sha256 + the same `vendor/` assets or a python-side copy + a no-CDN test that covers this template too). A first step either way: extend the no-CDN-style guard to whatever this surface renders.

(Context: `openant report -f html` uses the **Go** renderer — the Python path is reached via other flows that render `html_report.py`'s output.)

gadievron:
## What this fixes

Closes #502 — the Python HTML report (`report/html_report.py`, the dev-only surface the sink-is-dead boundary guards out of every shipped path) still loaded **floating** chart.js majors from the CDN: no version, no pin, no SRI — while #332's self-containment contract (vendored + sha-pinned + no-CDN guard) covers only the Go-rendered reports.

## The change

- The two `<script>` tags now use the **exact SOURCES.txt URLs** — the same bytes the Go surface vendors (`chart.js@4.5.1/dist/chart.umd.min.js`, `chartjs-plugin-datalabels@2.2.0`) — with **sha256 SRI** (base64 of the vendored blobs, computed from the files, not copied from prose) and **`crossorigin="anonymous"`** (SRI on a cross-origin script without a CORS request yields an opaque response the browser cannot verify — it *blocks* both scripts; the attribute is mandatory).
- **The cross-surface hash test is the sync mechanism**: `test_integrity_matches_the_vendored_go_bytes` asserts the Python pin's SRI equals `base64(sha256())` of the Go renderer's vendored blobs — a Go-side bump turns this file RED until the Python pin follows in the same commit. That is stronger than a renovate notifier (and `renovate.json` is deliberately untouched — every prior edit to it bought an incident; the existing matchStrings wouldn't match this URL form anyway).
- A floating-reference guard: no bare `npm/chart.js"` / `datalabels@2"` forms; every chart `<script>` carries `integrity` + `crossorigin`.
- The vendored-filename-missing case **fails** (that missing file IS the upgrade signal); only a standalone core checkout without the Go tree skips.

## Notes for the reviewer

- The surface remains dev-only and CDN-dependent by design (its docstring already says so — now it says *pinned*, not floating); the retire-vs-inline-vendor question stays open as the product decision it is (the issue's option (a) vs the full #332 recipe for a dead path — ruled wrong-cost for this PR).
- The renderer makes no LLM call itself (that's `main()` → `generate_remediation_guidance`), so the test renders end-to-end with `remediation_html=""`.

## Verification

- New tests: `2 passed` — RED on pristine (the floating refs / no-SRI state fails both).

### DIFF
--- a/libs/openant-core/report/html_report.py
+++ b/libs/openant-core/report/html_report.py
@@ -422,8 +422,8 @@ def generate_html_report(
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Security Analysis Report</title>
-    <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
-    <script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-datalabels@2"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chart.js@4.5.1/dist/chart.umd.min.js" integrity="sha256-SERKgtTty1vsDxll+qzd4Y2cF9swY9BCq62i9wXJ9Uo=" crossorigin="anonymous"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-datalabels@2.2.0/dist/chartjs-plugin-datalabels.min.js" integrity="sha256-IMCPPZxtLvdt9tam8RJ8ABMzn+Mq3SQiInbDmMYwjDg=" crossorigin="anonymous"></script>
     <style>
         :root {{
             --bg-primary: #1a1a2e;
--- a/libs/openant-core/tests/test_report_html_cdn_pin_sri.py
+++ b/libs/openant-core/tests/test_report_html_cdn_pin_sri.py
@@ -1,0 +1,77 @@
+"""#502: the dev-only Python HTML report pins and SRIs its CDN scripts.
+
+The #332 self-containment contract covers the Go-rendered reports; this
+dev-only surface (the sink-is-dead boundary guards it out of every
+shipped path) still loaded FLOATING chart.js majors from the CDN — no
+version, no SRI. The fix: the exact SOURCES.txt URLs (the same bytes the
+Go surface vendors), sha256 SRI (base64 of the vendored blobs), and
+crossorigin=anonymous (without it the browser cannot verify an opaque
+response and blocks the scripts). The cross-surface hash test makes the
+Python pin drift-impossible: when the Go side bumps, this file goes RED
+until the pin (and its hashes) follow in the same commit.
+"""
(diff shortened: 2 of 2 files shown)


gadievron:

**Re-derived independently:** RED on pristine = 2 failed / GREEN 2/2; suite green; ruff clean. **The SRIs re-derived from the vendored blobs myself** — `base64(sha256(chart-4.5.1.umd.min.js))` and `base64(sha256(chartjs-plugin-datalabels-2.2.0.min.js))` — both **MATCH** the emitted integrity attrs exactly (and the SOURCES.txt hex, the same bytes three ways). The `crossorigin=anonymous` claim verified against the W3C SRI spec (integrity requires a CORS-eligible response — jsdelivr sends `ACAO: *`). The skip-guard verified: CI's full checkout runs the hash test (only a standalone-core distribution skips — and it can't receive a Go bump).

**Fable gate: NOT-REFUTED.** Receipt note (low): the floating-ref guard is chart-scoped — a new non-chart CDN script tag would slip it (two external tags today; a universal `src=`-sweep requiring integrity+crossorigin+versioned-path would be the hardened form — left for the maintainer).

CI green. Merge-ready.

PR #530, opened 2026-09-07
fix(reachability): the library-blackout advisory is no longer silent under the band (#520)

### LINKED ISSUE #520: Keep-all net keyed on ANY real seed — consider keying on STRUCTURAL seeds only (the #506 follow-up; changes filtering for every incidental-only corpus)
From the #506 adjudication (closing it as not-planned; this is the ruling's named follow-up — a maintainer-level policy change, recorded for a decision):

**Today**: the empty-seed keep-all net (`core/parser_adapter.py:503-504`, replicated in all 8 language pipelines) fires when `real_entry_point_ids(entry_points, functions)` is empty — and 'real' counts ANY non-synthetic seed, including incidental `input_pattern:` seeds (check-3 regexes: argv, env vars, file-reading idioms — entry_point_detector.py:480-499, :398-404).

**The consequence**: a corpus whose ONLY entry-point cue is incidental (a library whose one helper reads external data) sits on a cliff — zero seeds = keep-everything + a loud warning (safe); one incidental seed = BFS from that seed, silently dropping every unreachable unit from analysis. A single regex widening (the #506 proposal) tips corpora off that cliff; the ≥0.90 blackout advisory is metadata/stderr-only and stays silent in the 0-90% pruning band.

**The proposed change**: key keep-all on STRUCTURAL seeds only (the seeded-by-type/decorator/name checks — the structural reason categories), so incidental seeds can enrich the enhancer's prompt context (entry-path framing, agent.py:236-261) WITHOUT ever disabling the keep-everything safety net. Incidental seeds would still seed BFS when structural seeds exist.

**Why maintainer-only**: it changes filtering behavior for every existing corpus whose seeds are currently incidental-only (those corpora move from BFS-back-to-keep-all), and it changes the meaning of the blackout advisory (structural-empty becomes the trigger). Requires a corpus-behavior review + a retained-unit-set regression across real corpora (in-tree fixtures cannot discriminate — zero file-read idioms in tests/fixtures).

gadievron:
## What this fixes

Closes #520 — not with the proposed re-key, but with the fix both adjudicating seats converged on: **the cliff's real defect is silence, not the BFS.**

**The measured cliff** (from the pre-registered experiment): an incidental-only corpus — every real seed an `input_pattern` match, zero structural seeds — at **87.9% pruning sat UNDER the advisory's 0.90 ratio gate**: the collapse was completely silent. The advisory's own comment names exactly this class ("a handful of incidental seeds yield a high reduction that looks like success while the real public API surface was never analysed") — and then gates it out at 87.9%.

## The change (ADVISORY-ONLY — no unit set changes anywhere)

1. **The ratio gate is dropped**: `blackout_warning` now fires whenever **no seed is structural** — incidental-only *or* excluded-only (the fuzz-only library the original #75 contract was written for) — at **any reduction**, with the wording scaled to what actually happened: a sub-threshold reduction kept most units, so the sub-band message says "the public API surface was not seeded; if this is a library, re-run with `--library-mode`" — it does **not** claim the core was dropped (that claim is earned only above the band).
2. **The one silent case is empty details** — the caller deliberately supplied `extra_entry_points`: chosen seeds are not incidental, and the warning slot belongs to the forward-asymmetry invariant there (this case surfaced in the build as a real interaction: the first version's always-fire crowded out the asymmetry warning; the empty-details discriminator resolves it).
3. **The seed-class counts ride the metadata**: `structural_entry_points` / `incidental_entry_points` in `metadata.reachability_filter` on **both branches** (the keep-all record and the filtered record), via a shared `classify_seeds` helper — the core site, the swift stash, and the rust result cannot drift (the rust pipeline's surface is the returned dict: `_seed_class_counts`).

## Why not the re-key (recorded on the issue)

The keep-all re-key was ruled premature: its findings sign is unmeasured (coverage-vs-framing), the classifier boundary is contested on the motivating corpus itself (`module_level_with_input` is not in `_STRUCTURAL_REASON_CATEGORIES` — the 132-vs-50 gap), the trigger is triplicated (drift risk), and `--library-mode` is the sanctioned union-only, framing-preserving lever. **Revisit condition**: a measured BFS-vs-keep-all-vs-`--library-mode` confirmed-findings comparison on ≥2 incidental-only corpora — only a measured win for keep-all reopens it.

## Verification

- New suite (9 tests): the **sub-band case warns** (the 87.9% cliff, with the no-false-claim wording), low-reduction incidental-only warns, above-band keeps the strong wording, structural suppresses, library-mode/empty/total-blackout unchanged, the classifier's counts + exclusions, and the **no-set-change proof** — the retained sets pinned exactly (the BFS set for incidental-only; keep-all kept all) with the counts asserted on both branches.
- The existing blackout suite (incl. the #134 hybrid negative control and the fuzz-only contract), the forward-asymmetry invariant, and the refilter suites all green.
- Full suite `3894 passed, 33 skipped`; `ruff check .` clean. RED-on-pristine note: the new file's import fails at collection (`classify_seeds` is a new API); the behavioral REDs are the sub-band silence and the absent metadata counts.

### DIFF
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -519,6 +519,10 @@ def apply_reachability_filter(
             "reduction_percentage": 0,
             "warning": warning,
         }
+        # #520: the seed-class counts on BOTH branches — the keep-all record
+        # must not read as "no signal" when synthetic-only seeding was the cause.
+        _rec["structural_entry_points"], _rec["incidental_entry_points"] = (
+            _epd.classify_seeds(detector.entry_point_details))
         # #328 (wave r1, opus+fable): the effective level on the pass-through
         # path is "all" — nothing was pruned — and this is the case where
         # recording it matters MOST: without it the record looks identical to
@@ -559,9 +563,13 @@ def apply_reachability_filter(
         if original_count > 0
         else 0
     )
+    _structural_eps, _incidental_eps = _epd.classify_seeds(
+        detector.entry_point_details)
     dataset.setdefault("metadata", {})["reachability_filter"] = {
         "original_units": original_count,
         "entry_points": len(entry_points),
+        "structural_entry_points": _structural_eps,
+        "incidental_entry_points": _incidental_eps,
         "reachable_units": len(filtered_units),
         "filtered_out": original_count - len(filtered_units),
         "reduction_percentage": reduction_pct,
--- a/libs/openant-core/core/parser_adapter.py
+++ b/libs/openant-core/core/parser_adapter.py
@@ -582,11 +582,24 @@ def apply_reachability_filter(
         file=sys.stderr,
     )
 
-    _blackout = blackout_warning(detector.entry_point_details, original_count,
-                                 len(filtered_units), library_mode=library_mode)
+    # #520 (the fable+astra fold): the seed-quality advisory rides its OWN
+    # key — following the orphan_advisory pattern — so it can never displace
+    # the forward-asymmetry invariant from the reserved ``warning`` slot (the
+    # correctness voice outranks the seed-quality voice at ANY reduction).
+    # extra_seed_count = the ACCEPTED-EFFECTIVE caller-supplied seeds
+    # (deduplicated, resolved to graph nodes) — named in the wording, never
+    # classified.
+    _extra_seeds_effective = sum(
+        1 for s in (extra_entry_points or [])
+        if s in reachable_ids or s in {u.get("id") for u in units}
+    )
+    _blackout = blackout_warning(
+        detector.entry_point_details, original_count,
+        len(filtered_units), library_mode=library_mode,
+        extra_seed_count=_extra_seeds_effective)
     if _blackout:
-        dataset["metadata"]["reachability_filter"]["warning"] = _blackout
-        print(f"  [Warning] {_blackout}", file=sys.stderr)
+        dataset["metadata"]["reachability_filter"]["blackout_advisory"] = _blackout
+        print(f"  [Advisory] {_blackout}", file=sys.stderr)
 
     # Per-unit prune telemetry (ADDITIVE, all-language; advisory — must never crash
     # the filter). Merges classification keys + the pruned_units.json sidecar; a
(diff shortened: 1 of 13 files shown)


gadievron:
**Gate-session receipt (the extra-validation round you asked for: fable + an OpenAI astra seat with two rounds of seat-to-seat cross-feedback + experimentation + impact analysis — the deepest gate of the window, on the reachability surface)**

**The process:** fable REFUTED the PR as-shipped (2 HIGH). astra (openai/gpt-6-astra, user-granted) independently adjudicated fable's record. The two seats then exchanged feedback directly, two rounds: fable conceded the M1 delivery-proof gap in its own alternative, the F2 evidence bar (the LLM CLI scenario plausible-unconstructed), the F3 downgrade, and the intent-language; astra converted from RETURN-TO-AUTHOR to **FOLD-A** with mandatory conditions. **Both seats converged: FOLD-A.** All conditions executed in gate commit 57ae0f9.

**Fable's two HIGHs (both re-derived from source by the orchestrator; astra confirmed):**
1. **The slot displacement** — the sub-band advisory occupied the reserved `warning` slot at ANY reduction, displacing the forward-asymmetry invariant (the correctness voice) via the `if _asym_warning and "warning" not in _rf` guard; the empty-details discriminator was fitted to the no-code asymmetry fixture and missed the incidental-seed variant.
2. **The empty-details mute** — empty details + ≥90% reduction went advisory→SILENT (the old gate fired there), re-creating the #520 class on the LLM path; the body's "empty unchanged" claim was false for this cell.

**The fold (astra's mandatory conditions, all executed):** the advisory rides its own `blackout_advisory` key at **all seven** producing sites (the orphan_advisory pattern); the reserved slot returns to the correctness signals; the reporter forwards the new key onto `reachability_warnings` (the envelope channel — delivery pinned by test); the mute became **wording** (`extra_seed_count` = accepted-effective caller seeds, named as unclassified, never classified — the mixed case names both populations); rust's false comment corrected; the 301 stderr pin updated to the folded contract; astra's two M4 regression tests + the slot-discipline and delivery pins added.

**Experimentation (the real pipeline, an incidental-only library corpus):** pristine master fires the above-band advisory at 93.3%/93.8% (verified) but the sub-band cell was the silent cliff; the folded tree lands the advisory on its own key with the metadata counts present and the reserved slot correctly empty when no correctness signal fired.

**Impact table (astra's, fable-adjudicated):** the S=true rows unchanged; the D=true sub-band row = the intended coverage gain; the D=true r=0 row = noise (closed by the wording scale); **the D=false ≥90% row = the UNSAFE regression the fold restores**; the r=1 and zero-denominator rows = unresolved-boundary, no "unchanged" claim made.

**Evidence:** the 520 suite 13/13; the 301 suite 16/16; the #481 conformance guard clean (it caught the first fold draft's `open()` — the window's own guard again); full suite **3845/0** (CI shape); ruff clean.

Issue #311, opened 2026-08-21
Dynamic-test generation failures bypass the counters, so `_summary.json` reports `errors: 0` and its buckets do not sum to its total


gadievron:
## Summary

When dynamic-test generation fails for a finding, the loop records a result and **saves a
checkpoint**, then `continue`s — bypassing the only site that increments `_completed` / `_errors`.
The phase's `_summary.json` therefore reports `errors: 0` for a run that errored, and its buckets do
not sum to its own total. On the `DetectFallback` path — used only when the Python
`checkpoint-status` invocation fails — that makes an errored run present to the operator as a clean
one.

> **Corrected 2026-08-21 — scope.** This sentence and the first paragraph of *Operator-visible
> effect* were unconditional. They hold **only** on `DetectFallback`. Every command
> (`enhance.go:79`, `scan.go:156`, `analyze.go:144`, `dynamictest.go:68`, `verify.go:86`) calls
> `DetectViaPython`, which builds `Info.Summary` from the **recomputed** counts, so on the normal
> path the label is correct. The counter bypass and the internally inconsistent `_summary.json` are
> real regardless; the mislabelling is not.

## The bypass

`libs/openant-core/utilities/dynamic_tester/__init__.py:248-256`:

```python
        if generation is None:
            print("  Test generation failed.", file=sys.stderr)
            result = collect_result(finding, None, None, generation_cost)
            result.generation_input_tokens = unit_usage["input_tokens"]
            result.generation_output_tokens = unit_usage["output_tokens"]
            results.append(result)
            if checkpoint:
                checkpoint.save(finding_id, result.to_dict())
            continue
```

The counters are incremented only at `:336-338`, inside the block at `:334-339` shown below, after
that `continue`:

```python
        if checkpoint:
            checkpoint.save(finding_id, result.to_dict())
            _completed += 1
            if result.status == "ERROR":
                _errors += 1
            checkpoint.write_summary(total, _completed, _errors, {}, phase="in_progress")
```

So the generation-failure path is checkpointed while counting as neither completed nor errored.
(The unit itself *is* retried on a later resume — `:135-139` treats `status == "ERROR"` checkpoints as
not-done and `:144-146` reports "Retrying N previously errored findings". The defect is in the
counters, not in retry behaviour.) `write_summary` is called with those counters at `:176`,
`:339` and `:371`.

## Observed on a real run

Against a completed scan's `dynamic_test_checkpoints/`, the two available sources disagree:

```
[...]

Issue #216, opened 2026-08-03
A missing `config/models.json` record silently prices a model at zero


NahumKorda:
## Summary

Running a current-generation model whose slug is absent from `config/models.json` produces a full scan in which **every phase reports a cost of zero**. Token counts are recorded correctly; only the monetary total is blind. The model dispatches normally.

## Evidence

An entire pipeline run reported `cost_usd: 0.0000` for all seven phases because the registry contained no record for `anthropic/claude-opus-5` or `anthropic/claude-sonnet-5` under the `openrouter` provider. Adding records restored accurate reporting, reconciling to 0.12% against billed spend on the Claude Sonnet 5 phase.

## Why it matters

The registry's own header comment states the design intent:

> `price` is null for any model that must not be auto-dispatched (retired/unknown) so it never resolves to a silent $0 in cost accounting.

A **missing** record defeats that guard entirely — the model is dispatched and priced at zero, which is the exact outcome the null-price convention was introduced to prevent. The guard covers models the registry knows about and deliberately refuses to price, but not models it has never heard of, and the second case is the common one as new models ship.

## Suggested fix

Two independent parts:

1. Add records for the current-generation slugs (`anthropic/claude-opus-5`, `anthropic/claude-sonnet-5`, and `anthropic/claude-fable-5` were all missing). Rather than embed rates that will drift, they are worth taking from the live catalogue at the time of the change (`GET /api/v1/models` on the gateway exposes per-token rates; the registry wants per-million).
2. **Make an unpriced-but-dispatched model as loud as a null-priced one** — emit a per-run warning and mark the reported cost as incomplete, instead of silently summing zeros.

Separately worth considering: records whose rate is a time-limited introductory price go stale silently. An optional `expires` field would make staleness detectable.

---

[...]

PR #543, opened 2026-09-08
fix(llm-reach): resume parity — per-unit records under the backend-identity gate (#532)

### LINKED ISSUE #532: scan --llm-reachability re-runs the completed reachability pass on every relaunch and overwrites its report — while every other checkpointed stage (enhance/analyze/verify/dynamic-test) restores
scan --llm-reachability re-runs the completed reachability pass on every relaunch and overwrites its report — while every other checkpointed stage (enhance/analyze/verify/dynamic-test) restores

From the 2026-09-07 monitored end-to-end run of the scan pipeline (5,424-unit corpus; at the then-master tree, before #527-#531 landed) and a three-axis verification round (independent raw-artifact re-derivation, source-code trace, adversarial refutation) followed by a two-model adjudication; re-derived from source at current master (ec6b063); pre-existing on master, not a regression from #526-#531.

## The claim

On a full-scan relaunch that restores per-stage checkpoints, the LLM-reachability stage re-runs from scratch:

- `core/scanner.py:564` — `if llm_reachability:` unconditionally re-enters the stage
- `core/scanner.py:645` — `signals = analyze_reachability(...)` is called on every invocation; there is no reuse branch
- `core/scanner.py:655` — the result overwrites `llm_reachability.json`
- `core/llm_reachability.py` — no persistence/reuse of a completed pass: fresh signals/batches per run, `parse_response` counts a malformed batch as dropped and returns `[]`, with no completed-pass cache anywhere

On the same relaunch, the four other checkpointed stages restore eligible work: `utilities/context_enhancer.py` (enhance, single-shot and agentic), `core/analyzer.py` (analyze), `utilities/finding_verifier.py` (verify), and `utilities/dynamic_tester/__init__.py` (dynamic-test) all load `StepCheckpoint` state and skip completed work. LLM-reachability is the only stage in that comparison set that re-bills and re-waits for a stage it already finished.

## Evidence (one run, three full passes)

The run's stderr receipts show the reachability stage completing fully three times (once per launch attempt of the same scan), each time overwriting the previous report:

```
[llm-reachability] Report: .../llm-reachability.report.json (2479.93s, $10.7898)
[llm-reachability] Report: .../llm-reachability.report.json (2322.45s, $10.6599)
[llm-reachability] Report: .../llm-reachability.report.json (2238.06s, $10.5851)
```

The surviving `llm-reachability.report.json` carries only the last pass. The earlier passes' accepted-call cost survives nowhere in the artifacts — the stage's `scan.report.json` cost provenance is overwritten along with the signals, so a relaunched scan cannot show what its total spend actually was.

## Direction

A fingerprinted completed-pass cache, replayed onto the freshly parsed dataset under the current promotion policy:

- Cache key: the pre-LLR dataset projection, application-context content, provider/model/options, prompt/parser version, and batch/cap settings — any change invalidates
- A cache hit must not freeze dropped/failed batches (those stay re-runnable and keep their dropped counts)
- Usage provenance recorded separately from "new calls made" (a replayed pass's report shows restored cost, not zero)
- Interrupted-pass per-batch checkpointing is additional scope

Related (none supplies reuse): #268/#371 (stage-body guard), #294/#386 (drop counting), #516/#528 (units_count restore window).

gadievron:
## Summary

A relaunch re-ran the whole llm-reachability pass (~$10 + ~70 min duplicated on a 5.4K-unit run) while every other LLM stage restored from checkpoints. The stage now uses the checkpoint family's own machinery:

- **per-unit records** `{signals, projection_sha, usage}` — a "reviewed, no signal" outcome IS a record (the majority case; signals alone cannot resume);
- the **backend-identity KEY** gates adoption (model/provider/adapter/base_url + the static template rendered with `app_context=None` — app-context content is excluded: it regenerates non-deterministically every scan, and content keying is the verified spurious-re-pay trap at LLR prices);
- **projection_sha hashes exactly the bytes the prompt sends** (`_unit_for_prompt`'s projection) — a body edit under the same id re-runs; null/non-str fields coerce rather than skip the stage;
- **dropped and exception batches leave no records** — absence IS the retry marker (no frozen coverage gaps, no error records the #311 drift class would miscount); the summary honors the #293 three-state invariant, with an `in_progress` phase written at pass start so the resume/fresh prompt sees a killed run correctly;
- **adopted signals (not promotion outcomes) replay** through `apply_signals` under the CURRENT promotion policy — promote-only semantics preserved (a `medium` recorded under a strict set promotes under a widened one);
- **restored usage lands as prior usage** (`add_prior_usage`; the #216 unpriced-model marker travels with per-record usage and restores across resume);
- `scan.go`'s fresh-start probe list gains `llm_reach` (the most expensive phase; an explicit fresh start can now discard it).

**Disclosed behavior change:** `valid_unit_ids` is per-batch — an out-of-batch signal was applied but never persisted, so adopted state now equals applied state; cross-batch ids were hallucination-only inputs that could wrongly seed the structural re-filter.

**Named residuals:** checkpoint records are forgeable by a directory writer — the same filesystem-bound residual the analyze/verify/enhance checkpoint family has always carried; an app-context change does not invalidate records (the KEY exclusion, same tier as analyze's).

Closes #532.

## Test plan

20 new tests in `tests/test_issue532_llr_resume.py`: adoption/no-repay, edited-unit re-run, foreign/removed records, dropped-batch and exception-batch retry, promote-set replay (recorded under `high`, promotes under `high,medium`), backend-change invalidation, out-of-batch signal isolation, prior-usage restore, the two summary shapes (done + interrupted), OSError tolerance (save + init), null-field coercion, the stale-done hazard, and the projection/prompt golden invariant (str and source-fallback code shapes).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue532_llr_resume.py -q` | 20 passed |
| family regression | `pytest tests/test_llm_reachability.py -q` | 30 passed |
| full suite | `pytest tests/ -q` | 3933 passed, 2 failed — both pre-existing (SDK pin drift; verified failing on the pristine base by reverting this change and re-running) |
| static analysis | ruff / Semgrep / CodeQL | clean / 0 findings / 0 findings in this diff (one pre-existing `empty-except` in scanner.py outside the touched hunks) |
| Go side | `gofmt -l` / `go vet ./cmd/` / `go build ./...` | clean / clean / builds |

End-to-end, live binding (`e2e-raptor-20260907`, a 2-language corpus, base @ master vs this branch, freshly built binaries per side):
- fresh base-vs-fix, python (4 units): identical coverage and promotions (6 signals / 3 promoted / 0 dropped both sides); one signal's confidence differed — the model's own run-over-run variance, no coverage effect;
- fresh base-vs-fix, go (3 units): **byte-identical** signals, kinds, confidences, and the same promotion;
- resume (this branch, same output dir): `units_adopted: 4`, LLR duration **0.0s** with **no new LLM spend**, the prior pass's cost restored as prior usage, the same signals replayed and the same promotions — vs the pre-fix behavior of re-paying the full pass every relaunch.

### DIFF
--- a/apps/openant-cli/cmd/scan.go
+++ b/apps/openant-cli/cmd/scan.go
@@ -151,7 +151,13 @@ func runScan(cmd *cobra.Command, args []string) {
 
 	// Check for interrupted runs in the scan directory
 	if ctx != nil && scanOutput != "" {
-		steps := []string{"enhance", "analyze", "verify"}
+		// #532: llm_reach joins the probe list — its checkpoints are the most
+		// expensive per-pass ($10+ on a 5K-unit corpus), so a silent auto-adopt
+		// on an explicit fresh start would defeat the user's choice. Records
+		// classify as completed (no error arms match) and a missing/unknown
+		// _summary takes PromptResume's legacy arm — same lifecycle as the
+		// other phases.
+		steps := []string{"enhance", "analyze", "verify", "llm_reach"}
 		for _, step := range steps {
 			if cpInfo := checkpoint.DetectViaPython(rt.Path, scanOutput, step); cpInfo != nil {
 				if !checkpoint.PromptResume(cpInfo, step, quiet) {
--- a/libs/openant-core/core/checkpoint.py
+++ b/libs/openant-core/core/checkpoint.py
@@ -1,9 +1,9 @@
 """
 Shared checkpoint utilities for resumable pipeline steps.
 
-Each LLM-heavy step (enhance, analyze, verify) can save per-unit checkpoint
-files so interrupted runs resume where they left off. The checkpoint dir
-lives next to the output file:
+Each LLM-heavy step (enhance, analyze, verify, llm_reach) can save per-unit
+checkpoint files so interrupted runs resume where they left off. The
+checkpoint dir lives next to the output file:
 
     {scan_dir}/enhance_checkpoints/
     {scan_dir}/analyze_checkpoints/
--- a/libs/openant-core/core/llm_reachability.py
+++ b/libs/openant-core/core/llm_reachability.py
@@ -376,6 +376,8 @@ def analyze_reachability(
     max_units: Optional[int] = None,
     on_error: Optional[Callable[[str], None]] = None,
     stats: Optional[Dict[str, int]] = None,
+    checkpoint_path: Optional[str] = None,
+    tracker: Optional[Any] = None,
 ) -> List[ReachabilitySignal]:
     """Run the LLM reachability review stage over a parsed dataset.
 
@@ -428,14 +430,19 @@ def analyze_reachability(
         probe_registry_or_raise(registry)
         binding = registry.get("llm_reach")
 
-    valid_ids = {u.get("id") for u in units if u.get("id")}
-
     # Lazy import so this module stays usable when callers explicitly
     # provide a binding and never want the registry fallback above.
     from utilities.llm import LLMAuthError, simple_text
 
+    # #532 (review-wave fix): the usage machinery must be live in PRODUCTION —
+    # resolve the global tracker when the caller doesn't pass one (the family
+    # pattern; without this, records persist zero usage and the restored-cost
+    # contract is dead outside tests).
+    if tracker is None:
+        from utilities.llm_client import get_global_tracker
+        tracker = get_global_tracker()
+
     signals: List[ReachabilitySignal] = []
-    batches = _chunk(units, batch_size)
     # #294: count parse-level batch drops and the units they carried —
     # a dropped batch is a coverage gap in the most consequential direction
     # (this stage decides which units are analyzed at all), and previously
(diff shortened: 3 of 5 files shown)


gadievron:
**Gate-session receipt (the two-seat round: fable + astra/openai-gpt-6-astra, both independently)**

**Re-derived independently:** the new suite 20/20; the family 30/30; the full suite **3844/0** on a fresh venv (the body's "2 failed pre-existing" verified as stale-venv SDK-drift — the fresh-pinned venv runs clean); ruff clean; the adoption gate verified at source (`:533-536` — a FRESH projection_sha is computed per current unit at every adoption, mismatch = no adoption = the unit re-runs; **the enforcement is live**, not descriptive — this closes astra's decisive question).

**Fable: NOT-REFUTED** — the key construction matches the frozen I2 family rule verbatim; the projection hash over the exact prompt projection; the cross-batch tightening grounded and undisclosed-flow-free; the forgeability the analyze tier or better.

**Astra: NOT-REFUTED** — with its sharpenings: (a) fable's "NO coverage gap possible" too broad (supported narrowly — no *recordless* unit freezes; not a universal coverage guarantee); (b) the MEDIUM confirmed real: **the threat-model/context exclusion conflates the LLM-narrated text (correctly excluded — content keying = the verified re-pay trap) with the deterministic context sources beneath it** — a repo edit that changes the context derivation while leaving a unit's code identical replays stale signals under the changed context (an FN-risk class: the signals gate the re-filter).

**Disposition of the MEDIUM (both seats converged)**: the residual is pre-existing family-wide (the analyze/verify/enhance resume keys exclude the app-context content at the same tier — the PR's own disclosure says "same tier as analyze's"), the PR correctly disclosed it, and the closure (a deterministic-**sources** fingerprint in the key — same-repo/same-fingerprint, no re-pay) is a family-wide design change, not a fold on this PR → **filed as #546**.

**Stacking with #544**: `git merge-tree` exit 0 — clean (fable's distinct-hunk note confirmed; astra's verification demand met).

**The fresh-start probe + the Go side** verified by the e2e the body carries (go byte-identical; resume 0.0s/no-spend — the orchestrator's own earlier runs corroborate the class). CI green incl. both windows legs. Merge-ready.

gadievron:
Counter-evidence on the "hallucination-only" premise in #543's disclosed behavior change — evidence for any future policy reconsideration, no ask.

## The observation

A reviewed full scan of the raptor repository at a pinned commit (2026-09-13, engine at current master) produced 33 `unknown unit_id … skipped` stderr lines (32 distinct ids). Joining the 32 distinct ids against the full parse dataset:

- **20 of the 32 resolve to real dataset units** (they exist in the dataset but were not in the emitting batch — cross-batch references, not hallucinations);
- 12 are absent from the dataset (correctly dropped hallucinations).

Of the 20 real ones, **18 also carry a home-batch signal** in the applied-signals artifact (the unit was named by its own batch elsewhere in the pass), so the skip did not lose the unit's coverage entirely. The residual — real ids with no signal anywhere — is 2 units, and their value **cannot be sized post hoc** because the skip line records neither the signal's `kind` nor its `confidence`: an entry_point/high skip and an external_input/low skip are indistinguishable in the record, yet only the former could have promoted the unit.

## What this does and does not say

- The per-batch design rationale ("a signal for a unit outside the producing batch is ungrounded by construction — the model never saw that unit's code") **stands**: prompts contain application context and source references that can elicit ids from other batches without grounding, and the adopted==applied state invariant is worth preserving. This is not a request to change the policy.
- The "hallucination-only" characterization is **falsified at this pin** — the skip class is a mix of real cross-batch references (20/32) and hallucinations (12/32), not hallucinations alone.
- The diagnostics gap is the actionable residue: the skip line's missing kind/confidence (and a `signals_skipped_unknown_unit` counter) is filed as #602.

Issue #300, opened 2026-08-20
Standalone `openant verify` step report omits `needs_review` and `error_count` that the pipeline records


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

The pipeline's verify step records `needs_review` and `error_count` in its step-report summary. The
standalone `openant verify` command builds the same summary without those two fields, so the command
that a user runs directly is strictly less informative than the same phase run inside `scan`.

## Evidence

Pipeline — `libs/openant-core/core/scanner.py:877-885` (HEAD `b501962`):

```python
                ctx.summary = {
                    "findings_input": verify_result.findings_input,
                    "findings_verified": verify_result.findings_verified,
                    "agreed": verify_result.agreed,
                    "disagreed": verify_result.disagreed,
                    "confirmed_vulnerabilities": verify_result.confirmed_vulnerabilities,
                    "needs_review": verify_result.needs_review,
                    "error_count": verify_result.error_count,
                }
```

Standalone — `libs/openant-core/openant/cli.py:498-504`:

```python
                    vctx.summary = {
                        "findings_input": vresult.findings_input,
                        "findings_verified": vresult.findings_verified,
                        "agreed": vresult.agreed,
                        "disagreed": vresult.disagreed,
                        "confirmed_vulnerabilities": vresult.confirmed_vulnerabilities,
                    }
```

The same omission appears a second time at `openant/cli.py:560-566`.

`VerifyResult.to_dict` (`core/schemas.py:293-304`) does emit both fields, so the stdout envelope from
the standalone command is complete — only the step report it writes to disk is not.

## Why it matters

On the run that prompted this work, those two fields carried `needs_review: 134` and
`error_count: 48` — 182 of 223 findings, 81.6%. A consumer reading the standalone command's step
report sees five counters that sum to far less than `findings_input` and nothing explaining the gap.

This is the same class as #285 (the aggregate report omitting `status`/`errors`), but a different
site: different file, different function, different command. #285's fix cannot reach it, which is why
it is filed separately rather than folded in.

## Suggested fix

[...]

Issue #285, opened 2026-08-19
`scan.report.json` is constructed without `status` or `errors` and never reads them from step reports


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

`_write_scan_report` builds the aggregate `StepReport(step="scan", ...)` with no `status=` and no
`errors=` keyword, and reads only `cost_usd`, `duration_seconds`, `token_usage` and `step` from the
per-step reports it aggregates. `StepReport.status` defaults to `"success"`, so the aggregate report
is unconditionally green. Repo-wide there is exactly one writer of `status="error"` —
`core/step_report.py:53` — and it sits in an `except` block that **re-raises**, so a step that
catches its own exception and records it in `ctx.errors` keeps `status="success"`.

## Evidence

`libs/openant-core/core/scanner.py:1268-1318` (abridged — `summary={ ... }` marks omitted keys) (HEAD `b501962`) — the constructor, abbreviated to the
keywords actually passed:

```python
    scan_report = StepReport(
        step="scan",
        summary={ ... },
        inputs={"repo_path": ...},
        outputs={ ... },
        cost_usd=round(total_cost, 6),
        duration_seconds=round(total_duration, 2),
        token_usage={ ... },
    )
```

No `status=`, no `errors=`. The only reads from `step_reports` are at `core/scanner.py:1259-1266`
(`cost_usd`, `duration_seconds`, `token_usage`) and `:1274` (`sr.get("step")` for
`steps_completed`). Confirm:

```
cd libs/openant-core
awk 'NR>=1253 && NR<=1322' core/scanner.py | grep -n 'status\|errors'
#   -> 33:            "parse_errors": result.parse_errors,      (from ScanResult, not a step report)
```

`libs/openant-core/core/step_report.py:50-57` — the sole writer of a non-default status:

```python
    try:
        yield report
    except Exception as exc:
        report.status = "error"
        report.errors.append(str(exc))
        print(f"[{step}] ERROR: {exc}", file=sys.stderr)
        traceback.print_exc(file=sys.stderr)
[...]

PR #544, opened 2026-09-08
fix(dynamic-test): counted errors make the step partial — the #285/#376 status contract honored via a shared summary helper (#533)

### LINKED ISSUE #533: The dynamic-test step reports status: success while counting errors — its producer writes the errors key, outside the #285/#376 partial-status contract
The dynamic-test step reports status: success while counting errors — its producer writes the errors key, outside the #285/#376 partial-status contract

From the 2026-09-07 monitored end-to-end run of the scan pipeline; re-derived from source at current master (ec6b063); pre-existing on master, not a regression from #526-#531.

## The claim

`core/scanner.py:1255` — the dynamic-test step's producer emits its summary under the key `"errors"` (`"errors": dt_result.errors`). But the step-report status contract established by #376 (closing #285) derives `partial` only from `summary["error_count"]` — `core/step_report.py`'s finally block counts `_summary.get("error_count")` and the `ctx.errors` list, not `summary["errors"]`. The dynamic-test producer's key is outside that contract, so a dynamic-test run with counted generation/build failures reports `status: "success"`.

#376's body covers the step_context contract but never names this producer — this is an uncovered case of the #285/#376 family, verified independently by both adjudication seats.

## Evidence (same run)

`out/dynamic-test.report.json`:
- line 3: `"status": "success"`
- line 18 (inside summary): `"errors": 1`

The corresponding checkpoint `_summary.json` records `errors: 1, error_breakdown: {"build": 1}` — a compose build failure counted but not reflected in the step status.

This run's aggregate scan status was `partial` for an unrelated reason (a verify-stage error), so the false `success` did not propagate to the aggregate this time. An otherwise-clean aggregate would inherit it through `_write_scan_report`'s per-step status aggregation (`core/scanner.py:1534-1569`, the `sr.get("status", "success")` default and `_STATUS_RANK` walk) — that inheritance path is source-derived, not observed in this run.

## Direction

- Emit `error_count` from the dynamic-test producer (or teach `step_context` the counted-error key set), keeping `errors` for existing consumers
- Tests: a failed-dynamic-step status test (a step with counted errors must not report `success`), and an otherwise-clean-aggregate test (one step `success`-with-errors must surface, not stay green)

Related: #209/#347 (report step honesty), #285/#376 (the contract this producer missed), #432/#470 (error_breakdown accounting).

gadievron:
## Summary

The dynamic-test producer wrote its counted failures under `summary["errors"]` (an int), but the `#285`/`#376` partial-status contract reads `summary["error_count"]` (the well-typed int) or the ctx error list — so a dynamic-test run with counted generation/build failures reported `status: "success"` (receipt: a monitored run's `dynamic-test.report.json` carried `"status": "success"` beside `"errors": 1` with a build failure in `error_breakdown`).

The fix, following the `#300` `verify_step_summary` pattern:

- **`dynamic_test_step_summary()`** in `core/schemas.py` is the ONE construction shape — the pipeline site (`core/scanner.py`) and the standalone `openant dynamic-test` site (`openant/cli.py`, the site the issue text missed) both route through it, so the sites cannot drift;
- it adds **`error_count`** beside the retained **`errors`** display key — the status contract is *not* widened to read `summary["errors"]`: the same word is a list of strings one level up on `StepReport`, and keeping one type per level is deliberate;
- a counted-failure step now reads **`partial`**; an otherwise-clean aggregate inherits it through the `_STATUS_RANK` walk; the Go SARIF `stepReportIsFailure` consumer flips `executionSuccessful` on the `error_count` it could already read (the CI-gate effect `#285` exists to deliver);
- the exception/degrade path is unchanged: a propagating error still reads `error`/`skipped` (the derivation's precedence at `core/step_report.py` is untouched).

**Disclosed residual (out of scope, named):** the standalone command's envelope still says `success(...)` and exits 0 on counted errors (only `confirmed > 0` sets exit 1) — the step report and the envelope now disagree where both previously lied success. This is the same envelope shape `verify` and `enhance` already have, tracked as a separate exit-code question. The reconciliation bound is documented on the helper (the `verify` precedent: the counters are a `<=`, the gap being language-SKIPPED rows).

Closes #533.

## Test plan

6 tests in `tests/test_issue533_dt_status_contract.py`:
- the RED case: a counted error through the real `step_context` → on-disk `partial` with `error_count == 1` and `errors` retained;
- zero errors stay `success`;
- the helper's exact key set (the seven-key contract);
- both call sites route through the helper (source tripwire) **and** no third inline construction site exists in either file (exclusivity grep);
- the helper call precedes the degrade `except` in the scanner block (the wiring-order tripwire);
- the aggregate `_write_scan_report` walk: clean steps + a partial dynamic-test ⇒ scan `partial`.

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue533_dt_status_contract.py -q` | 6 passed |
| family regression | `pytest tests/test_issue285_scan_report_status.py tests/test_issue311_dynamic_test_counters.py tests/test_issue431_dyntest_raise_accounting.py tests/test_artifact_serialization_contract.py -q` | 30 passed |
| full suite | `pytest tests/ -q` | 3920 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base by reverting this change) |
| static analysis | ruff / Semgrep / CodeQL | clean / 0 / 0 in this diff (the one new CodeQL hit — a test import style — normalized) |
| domain safety (stage 8) | structural parse base-vs-fix on a real corpus, `OPENANT_CORE_PATH` per tree | `dataset.json` units **byte-identical** (3/3, flags + entry-point reasons), `call_graph.json` identical — the deterministic pipeline core untouched |
| consumer trace | the hand-traced end-to-end (producer → step report → aggregate → sarif) | the only changes: step status success→partial on counted errors, the aggregate inheriting it, SARIF `executionSuccessful` flipping — every other consumer byte-identical |

### DIFF
--- a/libs/openant-core/core/scanner.py
+++ b/libs/openant-core/core/scanner.py
@@ -25,6 +25,7 @@ from pathlib import Path
 
 from core.schemas import (
     ScanResult, AnalysisMetrics, StepReport, verify_step_summary,
+    dynamic_test_step_summary,
 )
 from core.step_report import step_context
 from core import tracking
@@ -1246,14 +1247,7 @@ def scan_repository(
                         repo_path=repo_path,
                     )
 
-                    ctx.summary = {
-                        "findings_tested": dt_result.findings_tested,
-                        "confirmed": dt_result.confirmed,
-                        "not_reproduced": dt_result.not_reproduced,
-                        "blocked": dt_result.blocked,
-                        "inconclusive": dt_result.inconclusive,
-                        "errors": dt_result.errors,
-                    }
+                    ctx.summary = dynamic_test_step_summary(dt_result)
                     ctx.outputs = {
                         "results_json_path": dt_result.results_json_path,
                         "results_md_path": dt_result.results_md_path,
--- a/libs/openant-core/core/schemas.py
+++ b/libs/openant-core/core/schemas.py
@@ -421,6 +421,38 @@ class DynamicTestStepResult:
     def to_dict(self) -> dict:
         return asdict(self)
 
+    def step_summary(self) -> dict:
+        return dynamic_test_step_summary(self)
+
+
+def dynamic_test_step_summary(result: "DynamicTestStepResult") -> dict:
+    """The dynamic-test step-report summary (#533; the #300 verify pattern).
+
+    Shared by every construction site — core/scanner.py (the pipeline) and
+    openant/cli.py's standalone ``openant dynamic-test`` — so the sites
+    cannot drift. ``error_count`` is the #285/#376 partial-status contract's
+    well-typed int (``step_report.py`` derives ``partial`` from it and from
+    the ctx error list, never from ``summary["errors"]``); the ``errors``
+    key is retained as the persisted display contract (an int count here —
+    do NOT widen the status contract to read it: the same word is a list of
+    strings one level up on StepReport itself).
+
+    Reconciliation bound (the verify precedent): ``confirmed + not_reproduced
+    + blocked + inconclusive + errors <= findings_tested`` — the gap is the
+    language-SKIPPED rows (and any restored row whose persisted status falls
+    outside the five counted words); restored checkpoints ARE counted into
+    their buckets, and SKIPPED rows are deliberately not counted as errors.
+    """
+    return {
+        "findings_tested": result.findings_tested,
+        "confirmed": result.confirmed,
+        "not_reproduced": result.not_reproduced,
+        "blocked": result.blocked,
+        "inconclusive": result.inconclusive,
+        "errors": result.errors,
+        "error_count": result.errors,
+    }
+
 
 # ---------------------------------------------------------------------------
 # Step Report — written as {step}.report.json by every pipeline step
--- a/libs/openant-core/openant/cli.py
+++ b/libs/openant-core/openant/cli.py
@@ -779,7 +779,7 @@ def cmd_build_output(args):
 def cmd_dynamic_test(args):
     """Run Docker-isolated dynamic exploit testing."""
     from core.dynamic_tester import run_tests
-    from core.schemas import success, error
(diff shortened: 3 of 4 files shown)


gadievron:
**Gate-session receipt (the two-seat round: fable + astra)**

**Re-derived independently:** the new suite 6/6; the family (issue285 + issue311 + issue431) 16/16; the full suite **3735/0** fresh-venv (the "2 pre-existing" verified as stale-venv SDK-drift); ruff clean; the helper verified at source (the ONE construction shape in schemas.py; both sites route through — the exclusivity grep pinned by test).

**Fable: NOT-REFUTED** — the helper follows the #300 precedent exactly; the SARIF flip trace complete and pinned; the envelope residual the pre-existing verify/enhance shape, **not a new false-clean channel**.

**Astra: NOT-REFUTED** — with its sharpenings: (a) the type-per-level choice is not itself a conflict (different schema paths; no confused consumer identified); (b) fable's envelope-residual bound stands for the SARIF-derived path but "invisible to CI" is too broad — CI can consume the envelope/exit-code instead of SARIF (the residual stays relevant to such callers; simply not shown to be a NEW channel this PR introduces); (c) the aggregate `_STATUS_RANK` walk: the previously-success-now-partial change is the intended correction, not a compatibility defect — with the unverified-propagation demand (exit-code mapping, summary, enum validation) noted; the PR's own test plan covers the aggregate walk and the mixtures (the "cannot mask more severe" test would be the follow-up if the maintainer wants it).

**Disposition**: the disclosed envelope disagreement is correctly bounded (the same shape verify/enhance already carry — the tracked exit-code question named in the body); astra's verification items are the PR's own tests plus the named residual. **Stacking with #543: merge-tree exit 0** — clean, both orders.

CI green. Merge-ready.

PR #545, opened 2026-09-08
fix(report): Stage 2's explanation reaches the disclosure body — the dead hook repopulated, the deterministic banner carries the explanation (#534)

### LINKED ISSUE #534: Disclosure bodies never see Stage 2's explanation — the one hook key has no producer, so a disclosure can assert what the verify stage refuted
Disclosure bodies never see Stage 2's explanation — the one hook key has no producer, so a disclosure can assert what the verify stage refuted

From the 2026-09-07 monitored end-to-end run of the scan pipeline; re-derived from source at current master (ec6b063); pre-existing on master, not a regression from #526-#531. This is an artifact-consistency defect: an OpenAnt-generated document contradicts another artifact of the same run. It makes no claim about the scanned target either way.

## The claim

`core/reporter.py:547-595` builds the disclosure input record from Stage-1 fields (`vuln.description/impact/steps_to_reproduce`, `attack_vector`, `exploit_path.data_flow`). The only Stage-2 text hook is `finding.get("verification_explanation")` (`core/reporter.py:500-501`) — a key **no producer writes**: the verify stage stores its explanation under `verification.explanation` / `verification_note` in `results_verified.json`, never as the top-level `verification_explanation` the reporter reads. `report/generator.py:586-594` dumps that record into `prompts/disclosure.txt`'s `{vulnerability_data}`.

To be precise about what DOES reach the disclosure: the deterministic verdict banner (`_disclosure_verdict_header`, `report/generator.py:623`) reflects Stage 2's verdict, and the verdict itself is in the record. What never reaches the disclosure model is Stage 2's **explanation** — the reasoning, conditions, and corrections that qualify the verdict.

## Evidence (same run)

- A generated disclosure's Steps to Reproduce reads "**Step 2:** Trigger the default-on autodetection." — while the same run's verify-stage explanation for that finding states verbatim that "Stage 1's cited entry path is factually broken: default-on autodetect is gated by `_filter_locally_built` … DROPS all repo-committed binaries" (`results_verified.json` line 152). The disclosure asserts the path Stage 2 said is gated.
- Another disclosure carries unconditional arbitrary-code-execution wording, while the verify stage's explanation for that finding conditions it on a separate native-parser defect (`results_verified.json` line 51).
- The #442 failure label renders correctly in both disclosures' banners — the banner path works; the explanation path is the dead one.

## Direction

- Thread `verification.explanation` (and `verification_note`) into the finding record as named fields the reporter reads
- Instruct `disclosure.txt` that Steps to Reproduce must not contradict the verification explanation and must carry its stated preconditions
- Consider a deterministic "Stage-2 qualifications" block spliced like the code section, so the model cannot drop it

Related: #319/#442 (the verdict label that works), #210 (the deterministic banner), #296/#518 (verdict-preservation discipline).

gadievron:
## Summary

The disclosure record was built from Stage-1 fields only: the one Stage-2 text hook, `finding.get("verification_explanation")`, read a key **no producer writes** — the verify stage stores its explanation under `finding["verification"]["explanation"]`. The disclosure model therefore never saw Stage 2's reasoning, and a disclosure could assert what Stage 2 refuted (receipt: a monitored run's `DISCLOSURE_02` asserted the exact entry path the same run's `results_verified.json` said was gated by the autodetect filter; only the deterministic verdict banner was Stage-2-aware).

The fix:

- the finding record now carries **`verification_explanation`** as its own key, sourced from the verify stage's real location (`verification["explanation"]`), present-only, coerced (`_coerce_to_str` — dict-shaped explanations are a real model-output class); it rides `{vulnerability_data}` into the disclosure prompt;
- **`verification_note`** travels too — read from the top-level `verification_note` the verifier actually writes (the reclassification "Changed from X to Y" text and the incomplete-verdict notes);
- the steps-to-reproduce rebuild path reads the same real location;
- the deterministic **`_disclosure_verdict_header`** (the one block the model cannot drop) now appends a `> **Stage-2 explanation:**` line, blockquote-safe (multi-paragraph explanations stay inside the banner);
- `disclosure.txt` gains one consistency instruction: when the field is present, Summary/Steps must not assert reachability of a path the explanation describes as gated.

**Residual, stated plainly:** the prose can still contradict the banner — the failure mode is downgraded from *silently wrong* to *visibly self-contradictory with the correct half server-stamped*. That is the honest improvement; the model's prose is not deterministic.

Closes #534.

## Test plan

8 tests in `tests/test_issue534_disclosure_stage2.py`: the explanation reaching the record (the #534 regression), the fresh-wins precedence, the top-level note traveling, absent-stays-absent (present-only), the steps-rebuild path, the banner carrying the qualification verbatim, the no-explanation header unchanged (negative), and the multi-paragraph blockquote-safety.

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue534_disclosure_stage2.py -q` | 8 passed |
| family regression | `pytest tests/test_reporter_coercion.py tests/test_disclosure_verdict_header.py tests/test_issue319_dynamic_status.py -q` | 42 passed |
| full suite | `pytest tests/ -q` | 3923 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| static analysis | ruff / Semgrep / CodeQL | clean / 0 / the 3 hits in my two files are pre-existing outside the touched hunks (verified against the prior scan) |
| consumers | the schema is presence-only (no additionalProperties); the payload's exclude list untouched; the summary allowlist untouched; the Go/webui side has no key-set assertion | additive-key safe everywhere |

### DIFF
--- a/libs/openant-core/core/reporter.py
+++ b/libs/openant-core/core/reporter.py
@@ -586,6 +586,13 @@ def build_pipeline_output(
                or finding.get("verification_explanation") else {}),
             **({"verification_note": finding["verification_note"]}
                if finding.get("verification_note") else {}),
+            # #534 gate fold (fable+astra): the consistency pass rewrites
+            # the verdict WITHOUT touching verification["explanation"] — the
+            # banner can stamp an updated status beside the explanation that
+            # argued for the original one. Carry the update present-only so
+            # the banner can attribute the pre-update explanation correctly.
+            **({"consistency_update": finding["consistency_update"]}
+               if isinstance(finding.get("consistency_update"), dict) else {}),
             # #215 (partial repair): the two FINDING-SEMANTIC transit
             # fields — confidence (float 0.0-1.0 per the verdict schema,
             # json_corrector.py:32; NOT analysis_core.py:181's error-shape
--- a/libs/openant-core/core/reporter.py
+++ b/libs/openant-core/core/reporter.py
@@ -497,8 +497,17 @@ def build_pipeline_output(
                 else:
                     flow_str = _coerce_to_str(data_flow)
                 parts.append("Data flow: " + flow_str)
-            if finding.get("verification_explanation"):
-                parts.append("Verification: " + _coerce_to_str(finding["verification_explanation"]))
+            # #534: read Stage 2's explanation from where the verify stage
+            # actually writes it (finding["verification"]["explanation"]),
+            # with the legacy top-level key as fallback — the old read was a
+            # hook to a key NO PRODUCER wrote (the disclosure body could
+            # assert what Stage 2 refuted).
+            _v = finding.get("verification")
+            if not isinstance(_v, dict):
+                _v = {}
+            _v_expl = _v.get("explanation") or finding.get("verification_explanation")
+            if _v_expl:
+                parts.append("Verification: " + _coerce_to_str(_v_expl))
             steps_to_reproduce = "\n\n".join(parts) if parts else None
 
         # Determine stage2 verdict.
@@ -564,6 +573,19 @@ def build_pipeline_output(
             "cwe_name": vuln.get("cwe_name") or finding.get("cwe_name") or full_result.get("cwe_name", "Unknown"),
             "stage1_verdict": finding.get("verdict", finding.get("finding", "vulnerable")),
             "stage2_verdict": stage2_verdict,
+            # #534: Stage 2's text reaches the disclosure prompt as OWN
+            # record keys. Fresh-wins precedence (the verify dict is the
+            # authoritative Stage-2 record — the same precedence as the
+            # steps-rebuild path); coerced (dict-shaped explanations are a
+            # real model-output class); the note is the TOP-LEVEL
+            # verification_note the verifier actually writes; present-only.
+            **({"verification_explanation":
+                    _coerce_to_str(verification.get("explanation")
+                                  or finding.get("verification_explanation"))}
+               if (isinstance(verification, dict) and verification.get("explanation"))
+               or finding.get("verification_explanation") else {}),
+            **({"verification_note": finding["verification_note"]}
+               if finding.get("verification_note") else {}),
             # #215 (partial repair): the two FINDING-SEMANTIC transit
             # fields — confidence (float 0.0-1.0 per the verdict schema,
             # json_corrector.py:32; NOT analysis_core.py:181's error-shape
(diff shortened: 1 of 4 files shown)


gadievron:
**Gate-session receipt (the two-seat round: fable + astra/openai/gpt-6-astra, both independently)**

**Re-derived independently:** the dead-hook claim verified from source (on pristine master ONLY the reporter reads the top-level `verification_explanation` — no producer writes it; the verifier writes `verification["explanation"]` at `:347`); the new suite 8/8; the family 34/34; full suite **3876/0** fresh-venv post-fold; ruff clean.

**Fable: NOT-REFUTED** — the injection surface adjudicated (the prompt path "same class" — description/impact/steps already ride unsanitized; the banner path "a new but well-contained verbatim channel" — the per-line `>` prefix stronger than the existing vulnerable_code fence); the fresh-wins precedence correct; the `verification_note` channel prompt-only (no double render); the steps-rebuild acceptable (it hands the contradiction to the model with the instruction).

**Astra: NOT-REFUTED** — with its sharpenings: (a) MEDIUM 1 downgraded to an explicit design choice ("a documented decision to remain unbounded is acceptable"); (b) MEDIUM 2 retained as "the strongest finding — real, distinct, deserving explicit disclosure; **fold** the consistency context into this PR"; (c) the blockquote containment "conditionally adequate — normalize newlines and verify; not sanitized or injection-proof"; (d) the `vulnerable_code` comparison qualified (a properly constructed code fence suppresses Markdown more strongly than a blockquote — the blanket claim unsupported).

**The fold (gate commit 966295a, all three):**
1. **MEDIUM-2 (the substantive one)**: `consistency_update` now rides the record present-only; the banner renders a `> **Consistency update:** from → to — reason` line; and the explanation's **label attributes the pre-update state** when an update occurred ("Stage-2 explanation (pre-dating the consistency update…)") — the reader always knows which conclusion the reasoning supported.
2. **MEDIUM-1**: resolved by **explicit decision** — the banner is deliberately unbounded (the gate clause is the point of #534; a cap risks cutting it), the decision recorded in the code docstring with the 300-cap-sibling rationale.
3. **LOW-4**: CRLF/bare-CR normalized before the blockquote prefix + the containment tests (CRLF/CR, `>`-leading, links).

**Receipt notes:** the errored-path `verification_note` (raw exception text into the prompt — LOW 3) and the 3× prompt duplication (LOW 5) named as follow-ups per astra's triage. The `vulnerable_code` fence-escape gap (pre-existing) is fable's observation, unchanged by this PR.

PR #548, opened 2026-09-08
fix(dynamic-test): a failed compose build still tears down (#536)

### LINKED ISSUE #536: Dynamic-test compose path: a build failure returns before the down --rmi local teardown, so compose images survive; the outer cleanup removes single-container names only
Dynamic-test compose path: a build failure returns before the down --rmi local teardown, so compose images survive; the outer cleanup removes single-container names only

From the 2026-09-07 monitored end-to-end run of the scan pipeline; re-derived from source at current master (ec6b063); pre-existing on master, not a regression from #526-#531.

## The claim

In `utilities/dynamic_tester/docker_executor.py`, `_run_compose` returns its result on build failure (`code != 0 or timed_out`) **before** reaching the `try/finally` that runs `compose down --volumes --remove-orphans --rmi local` — the teardown only covers the success path upward from there. The outer caller's cleanup is keyed to the single-container naming: `run_single_container`'s `finally` calls `_cleanup_docker(image_tag=f"openant-test-{run_id}-{safe_id}", network)` — which does not match the compose image names at all (`compose_project_name()` yields `openant-{run_id}-{safe_id}`, images `openant-{run_id}-{safe_id}-<service>`).

So a compose build failure leaks the built images and the network. Related family: #315/#424 already gave compose runs per-run project names — the naming part works; this is the uncovered teardown path (the PR that established per-run naming states "Nothing else changes" and does not name the build-failure branch).

## Evidence (same run)

The run's VULN-001 dynamic test hit a compose build failure. A post-run image named `openant-069b737a-vuln-001-attacker` matches the compose naming shape exactly (project name + the `attacker` service the test generator writes). The run's docker event log carries no image names, so attribution is by name shape only — no claim is made about image counts.

## Also carried (smaller): the silent source_file=None resolution

`utilities/dynamic_tester/__init__.py:409-418` — when `source_file` resolves to `None` for a finding whose `location.file` was present (repo_path falsy, non-str path, realpath escape, or not-a-file), nothing records which guard fired; the generated Dockerfile then fails with an opaque `COPY <basename> .` not-found. Emit the resolved-or-rejected reason into the result details.

Caveat, stated plainly: the `repo_path` pass-through omission itself is fixed on master by #531 (`core/scanner.py:1246`), but that fix has not yet been exercised by a dynamic-test run at master; this residual is source-verified, not observed-at-master.

## Direction

- Run the `compose down --volumes --remove-orphans --rmi local` teardown on the build-failure branch (or enclose build/up/logs in the same `finally`)
- Keep the work_dir until teardown has run
- The `source_file=None` path: log which guard fired, into the result details the report surfaces

Related: #315/#424 (per-run project naming — the part that already works), #521/#531 (the repo_path channel).

gadievron:
## Summary

A compose build failure (or timeout) returned early — before the `try/finally` that runs `compose down --volumes --remove-orphans --rmi local` — so the images the partial build had already created leaked (per-run UUID names like `openant-<run>-<id>-<service>` accumulate forever; the receipt was a post-run image matching the compose naming shape exactly). The outer cleanup (`_cleanup_docker`) removes single-container names only, not the compose project's images.

The teardown now runs on the build-failure branch too. `--rmi local` is a no-op for a build that failed before any service built; the success path's teardown is unchanged (exactly one `down`, no double-down — the failure branch returns before the `try`).

Closes #536.

## Test plan

3 tests in `tests/test_issue536_compose_teardown.py`: a failed build tears down with `--rmi local` (exactly 2 commands: build + teardown); a build timeout tears down; the success path tears down exactly once (no double-down).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue536_compose_teardown.py -q` | 3 passed |
| full suite | `pytest tests/ -q` | 3959 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| adversarial review | the combined wave+refute | SHIP; the two test tightenings (exact call counts, the -1 timeout convention) folded |

### DIFF
--- a/libs/openant-core/tests/test_issue536_compose_teardown.py
+++ b/libs/openant-core/tests/test_issue536_compose_teardown.py
@@ -1,0 +1,92 @@
+"""Tests for issue #536 — the compose build-failure teardown gap.
+
+`_run_compose` returns early on a build failure (``code != 0 or timed_out``)
+BEFORE the ``try/finally`` that runs ``compose down --volumes
+--remove-orphans --rmi local`` — so a failed build leaks the images it
+partially built (per-run UUID names accumulate forever). The outer cleanup
+(``_cleanup_docker``) removes single-container names only
+(``openant-test-<run>-<id>`` + the network), not the compose project's
+``openant-<run>-<id>-<service>`` images.
+
+Receipt (the monitored run): a compose build failed for VULN-001; the
+post-run image ``openant-069b737a-vuln-001-attacker`` matched the compose
+naming shape exactly (project + the attacker service) — name-shape
+attribution (the docker event log carries no image names).
+"""
+from __future__ import annotations
+
+from unittest.mock import patch
+
+from utilities.dynamic_tester.docker_executor import _run_compose
+
+
+class _Recorder:
+    def __init__(self, results):
+        self._results = list(results)
+        self.calls: list[list[str]] = []
+
+    def __call__(self, cmd, timeout=None, cwd=None):
+        self.calls.append(list(cmd))
+        return self._results.pop(0)
+
+
+def _seq(*outcomes):
+    """Build the _run_command result sequence.
+
+    outcomes: tuples (stdout, stderr, code, timed_out) in call order.
+    """
+    return [o for o in outcomes]
+
+
+def test_build_failure_still_tears_down():
+    """THE #536 regression: a failed build must still run compose down
+    --rmi local (the images the partial build created must not leak)."""
+    # build fails, then (post-fix) the teardown runs.
+    rec = _Recorder([
+        ("", "failed to solve: exit 1", 1, False),   # docker compose build
+        ("", "", 0, False),                           # docker compose down (teardown)
+        ("", "", 0, False),                           # docker compose logs (defensive)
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (the combined astra round; fable did not need to gate this one — the micro-surface)**

**Re-derived independently:** the 3 tests 3/3; full suite 3773/0 fresh-venv; ruff clean; the early-return-before-try/finally shape verified from the diff.

**Astra: NOT-REFUTED** — "cleanup is appropriate: a build can leave completed intermediate service images; failure is not equivalent to nothing happened" — with the lifecycle caveats named as the checklist: project isolation (the compose project is per-run-UUID-named — the #424 fix — so the pre-existing-container hazard the caveat names is structurally excluded), the subprocess-termination-before-teardown ordering, the failure-precedence (the teardown must not mask the build failure), and the teardown's own boundedness. **None of the hazards is established as occurring in this record** (astra's own words) — the per-run UUID naming answers the ownership caveat; the executor's call-order tests pin the build→teardown sequence.

CI green. Merge-ready.

PR #549, opened 2026-09-08
fix(llm): a rejected completion's returned usage travels on the error and reaches the tracker (#537)

### LINKED ISSUE #537: A rejected completion's returned usage is discarded before it is read; a verify conversation's accumulated usage is lost on exceptional exit
A rejected completion's returned usage is discarded before it is read; a verify conversation's accumulated usage is lost on exceptional exit

From the 2026-09-07 monitored end-to-end run of the scan pipeline; re-derived from source at current master (ec6b063); pre-existing on master, not a regression from #526-#531. The provider-side charge for the failed calls is **unknowable from the artifacts** (the raw responses were destroyed by the harness's relaunch cleanup) and this issue makes no claim about it — the defect claim is about the two source-pinned discard paths.

## The claim

**(a) Raise-before-usage-read at the adapter boundary.** `utilities/llm/providers/openai.py:906-911` raises `LLMResponseError` on an empty completion ("no text or tool calls") **before** the usage read at `:928` (`usage = getattr(response, "usage", None)`). Whatever usage the rejected response carried is discarded un-read. Upstream, `utilities/llm/helpers.py:92` (`simple_completion`) records usage only after `adapter.complete()` returns — so a call that raises never records.

**(b) The verify conversation flushes only on normal exits.** `utilities/finding_verifier.py` calls `adapter.complete()` **directly** (`:486`), bypassing `helpers`' recording; it accumulates per-turn usage locally and reaches `tracker.record_call` only at the outcome branches (`:523, :585, :603, :626, :647`). The outer `except Exception` block records the error and marks the result incomplete but never flushes the accumulation — so a conversation that fails on a later turn loses the accounting for its earlier successful turns too. The same shape exists in `utilities/agentic_enhancer/agent.py:282-305`.

## Evidence (same run)

`results_verified.json` carries a unit errored with `"LLMResponseError: ... returned an empty completion"`, displayed with zero cost. The step's report therefore says $0 for calls whose provider-side usage may or may not have been returned with the rejection — the accounting artifact cannot say, by construction.

## Direction

- Carry `usage`/`usage_details` on the error type at the adapter boundary, so a rejected reply's usage (when present) survives the raise
- Record exactly once in `helpers.py` and in the direct callers' exceptional paths
- Keep the rejection semantics: an empty completion is never accepted as clean — this is about accounting, not acceptance
- Tests: a rejected reply with nonzero usage records it; a successful turn followed by a rejection preserves the earlier turn's accounting

Related (checked, not these): #292/#384 (retry classification), #211/#350, #216/#367 (pricing), #512/#527 (truncation warnings — a different failure class).

gadievron:
## Summary

The empty-completion raise fired BEFORE the usage read, and `simple_completion` recorded usage only after `complete()` returned — so a rejected reply's tokens were discarded un-read (the artifact showed $0 for calls the provider may already have billed), and a verify conversation failing on a later turn lost the accounting for the turns before it.

The fix:

- **`LLMResponseError` carries the rejected reply's usage** (`input_tokens` / `output_tokens` keyword args; zeroes when the failure predates a usage block — a transport error or pre-parse refusal is never billed as a completion by construction). The signature change is additive: every existing raise site (positional message only) is unaffected, including the `LLMRefusalError` subclass.
- Both openai empty-completion guards (the no-choices site and the no-usable-content site — the sibling the review round caught) pass the response's usage onto the error.
- **`simple_completion` records it via the same `record_call` path as a successful call** (pricing lookup included) BEFORE re-raising — the rejection semantics are untouched: an empty completion is never accepted as clean.

**Disclosed bound:** what the provider actually charged is unknowable from the artifact and stays unclaimed; the token counts are what the response carried. The multi-turn verify loop's own flush-on-exit remains a separate surface (this fix covers the call-level accounting at the `simple_completion` boundary — the boundary every phase shares).

Closes #537.

## Test plan

5 tests in `tests/test_issue537_rejected_usage.py`: the error carries the usage; recorded before re-raise (the tracker total never $0 for a billed call); a conversation that succeeds then rejects keeps both turns' usage; an auth error records nothing; and the REAL provider site (a client stub driving `OpenAIAdapter.complete` → the raise carries the usage).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue537_rejected_usage.py -q` | 5 passed |
| full suite | `pytest tests/ -q` | 3961 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| adversarial review | combined wave+refute | the two blockers folded: the sibling no-choices site + the provider-site test; the multi-site sweep verified the refusal subclass's constructor unaffected |

### DIFF
--- a/libs/openant-core/tests/test_issue537_rejected_usage.py
+++ b/libs/openant-core/tests/test_issue537_rejected_usage.py
@@ -1,0 +1,166 @@
+"""Tests for issue #537 — a rejected completion's returned usage is not lost.
+
+The empty-completion raise in providers/openai.py fires BEFORE the usage
+read, and simple_completion records usage only after complete() returns —
+so a rejected reply's usage is discarded un-read, and a verify conversation
+that fails on a later turn loses its earlier turns' accounting on the
+exceptional exit.
+
+The fix: LLMResponseError carries the rejected reply's usage (when the
+provider supplied it), and helpers.simple_completion records it via
+add_prior_usage before re-raising.
+"""
+from __future__ import annotations
+
+import pytest
+
+from utilities.llm import (
+    CompletionResult,
+    LLMAuthError,
+    LLMResponseError,
+    PhaseBinding,
+    TextBlock,
+)
+from utilities.llm_client import TokenTracker
+
+
+class _Usage:
+    def __init__(self, prompt_tokens, completion_tokens):
+        self.prompt_tokens = prompt_tokens
+        self.completion_tokens = completion_tokens
+
+
+class _Resp:
+    def __init__(self, usage, finish="stop"):
+        self.usage = usage
+        self.choices = [type("C", (), {"message": type("M", (), {"content": None}),
+                                       "finish_reason": finish})()]
+
+
+class RejectingAdapter:
+    """Mimics the provider contract: the raise path fires INSIDE complete
+    (the real openai.py raises before returning) — with the usage attached
+    per the #537 fix."""
+    name = "openai"
+    supports_tools = True
+
+    def __init__(self, usage):
+        self._usage = usage
+        self.calls = 0
+
+    def complete(self, *, model, system, messages, max_tokens, tools=None):
+        self.calls += 1
+        raise LLMResponseError(
+            "openai returned an empty completion",
+            input_tokens=getattr(self._usage, "prompt_tokens", 0),
+            output_tokens=getattr(self._usage, "completion_tokens", 0))
+
+    def validate(self, model):
+        pass
+
+
+class OkThenRejectAdapter:
+    name = "openai"
+    supports_tools = True
+
+    def __init__(self, usage):
+        self._usage = usage
+        self.n = 0
+
+    def complete(self, *, model, system, messages, max_tokens, tools=None):
+        self.n += 1
+        if self.n == 1:
+            return CompletionResult(
+                content=[TextBlock("ok")], input_tokens=10, output_tokens=5,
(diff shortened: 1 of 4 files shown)


gadievron:
**Gate-session receipt (the combined astra round)**

**Re-derived independently:** the 5 tests 5/5; full suite 3775/0 fresh-venv; ruff clean; the additive-signature claim spot-verified (the existing raise sites positional-message-only).

**Astra: NOT-REFUTED** — "an unusable completion can still be billable; the right direction" — with its checklist adjudicated: (a) the additive-signature compatibility holds given the premises (message positional; the kwargs defaulted; the base constructor normal); (b) the zeros-vs-unavailable conflation: no numerical double-counting (adding zero tokens adds zero cost) but the provenance distinction lost — "no usage block is not proof of unbilled" — the disclosed bound in the PR body already says exactly this ("what the provider actually charged is unknowable; the token counts are what the response carried"); (c) the real double-count check — whether the same provider response records twice — is the PR's own test 3 (a conversation that succeeds then rejects keeps both turns' usage; distinct attempts each count); (d) the refusal-subclass constructor path (inherits or overrides — must forward): the PR's own multi-site sweep covered the subclass's constructor unaffected.

**Receipt notes (astra's additions, follow-up grade):** the legacy positional-construction test + the refusal-with-usage-and-rethrow test would tighten the suite; the availability-vs-zero distinction is a documentation choice if any downstream report ever needs it.

CI green. Merge-ready.

Issue #294, opened 2026-08-19
llm-reachability: six distinct failure shapes emit one identical log line, and the raw response is discarded


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

When a reachability batch fails to parse, `parse_response` logs a single fixed string and returns an
empty list. At least six materially different failures — including a model refusal and an empty
completion — produce the byte-identical message. The raw response is never retained, the batch index
is not included, and nothing counts the drop. On our run 10 of 434 batches were lost this way and the
cause could not be determined afterwards from any artifact.

## Evidence

`libs/openant-core/core/llm_reachability.py:246-262` (excerpt — the 6-line docstring at `:251-256` elided; all other lines verbatim) (HEAD `b501962`):

```python
def parse_response(
    response_text: str,
    valid_unit_ids: Optional[set] = None,
    on_error: Optional[Callable[[str], None]] = None,
) -> List[ReachabilitySignal]:
    ...
    log = on_error or (lambda msg: print(f"[LLMReach] {msg}", file=sys.stderr))

    data = _extract_json(response_text)
    if not isinstance(data, dict):
        log("malformed response: not a JSON object — skipping batch")
        return []
```

`response_text` is a local parameter and is not written anywhere before the early return.

### Six shapes, one message — executed at HEAD

```python
from core.llm_reachability import parse_response
shapes = {
 "bare array":           '[{"unit_id":"a.py:f","signal":"entry_point"}]',
 "fenced bare array":    '```json\n[{"unit_id":"a.py:f"}]\n```',
 "truncated mid-string": '{"signals": [{"unit_id": "a.py:f", "reason": "this got cut',
 "truncated mid-array":  '{"signals": [{"unit_id": "a.py:f"},',
 "prose refusal":        "I can't help with analyzing this code for security purposes.",
 "empty completion":     "",
}
for name, txt in shapes.items():
    got = []
    parse_response(txt, valid_unit_ids={"a.py:f"}, on_error=got.append)
    print(name, "->", got)
```

All six produce exactly:

```
[...]

PR #550, opened 2026-09-08
fix(llm-reach): the deferred-truncation tracker — the stop reason on every drop line, the private cap retired, the truncation gate (#538)

### LINKED ISSUE #538: [tracker] LLM-reachability: the deferred malformed-batch recovery (#294/#386) and the private max_tokens=4096 cap (#512/#527's recorded residual) are untracked — both recurred on the next target
[tracker] LLM-reachability: the deferred malformed-batch recovery (#294/#386) and the private max_tokens=4096 cap (#512/#527's recorded residual) are untracked — both recurred on the next target

This is a follow-up tracker for two documented deferrals, not a new-defect claim. From the 2026-09-07 monitored end-to-end run of the scan pipeline and its verification round; no open issue tracked either deferral at filing time.

## The two deferrals

- **PR #386** (closing #294) shipped the malformed-batch diagnosis and the `batches_dropped` / `units_not_reviewed` counting. Its body records the deliberate deviation: "suggestion 5 (reuse `json_corrector` before discarding a batch) is deferred." The recovery itself never landed and nothing tracks it.
- **PR #527** (closing #512) swept the private caps it fixed but explicitly lists `llm_reachability.py`'s private `max_tokens=4096` among "Residual private caps (recorded, not swept here)". `core/llm_reachability.py` still hardcodes it at master, against the shared `DEFAULT_MAX_TOKENS`.

## Recurrence (the next target, same day)

A second monitored run (different language, 5,424-unit corpus) hit the same malformed-batch path: stderr shows `[LLMReach] malformed response: truncated or unbalanced JSON — skipping batch [batch 61/217, ...]` with a raw prefix that is well-formed JSON stopping mid-stream, and its report summary carries `batches_dropped: 1, units_not_reviewed: 25`. The raptor run's first attempt recorded 5 failed/malformed batches — 4 "returned an empty completion" (one failure class) + 1 malformed (a different, disjoint class per the two distinct log lines) — cause unknown: the raw responses are gone and stop reasons were not captured at that tree, so no claim is made whether the empty completions were the at-cap shape or provider-side filtering.

Units that land in a dropped/malformed batch get no reachability signals for that pass — they are never reviewed by the stage that decides analyzability. No claim is made about missed findings (structural analysis may still retain those units); the claim is about the review coverage and the untracked deferral.

## Direction (as a tracker)

1. Capture and record the stop reason on the malformed path (the truncated-batch log line should carry what the reply actually did)
2. Lift the private cap to the shared default, or make it phase-configurable
3. On a truncated batch: split-and-retry or a corrector pass, measured before adoption
4. A recovered batch must never overwrite an `units_not_reviewed` count without provenance

Related: #294/#386 (the deferral's ancestry), #512/#527 (the cap residual's ancestry).

gadievron:
## Summary

This tracker follows two documented deferrals to their levers (the #386 malformed-recovery deferral and #527's recorded private-cap residual), both of which recurred on the next target.

**(1) The stop reason is evidence.** The malformed path swaps to `simple_completion` (the `#512` `application_context` precedent — `simple_text` discards the result after joining), threads `result.stop_reason` into `parse_response`, and the classifier's brace heuristic becomes evidence-based: a `max_tokens` reply logs **"truncated at max_tokens"** (the diagnosis the recurrence receipts lacked); an `end_turn` reply with broken braces logs **"unbalanced JSON (not truncated)"**. `batches_truncated` counts the subclass and surfaces in the step summary beside `batches_dropped`. The `#512` one-time truncation warning stops firing for LLR — the per-batch drop line carries the same information.

**(2) The private 4096 cap retired.** The phase shares `DEFAULT_MAX_TOKENS` (the cap-not-floor contract: expected cost unchanged for healthy replies, which stop at `end_turn`; the OpenAI Responses path already floored to 16000). The checkpoint fingerprint is unaffected (`max_tokens` is not a KEY member) — records from batches that succeeded under 4096 stay valid.

**(4)-primitive.** A `max_tokens` reply **never persists records** — even when the salvage parse succeeds on a truncated prefix, the batch drops whole: the tail units must not freeze as "reviewed, no signal" in the checkpoint (the applied==adopted invariant from the resume machinery).

**(3) — the in-pass recovery — named as the measured follow-up:** split-and-retry is preferred over a JSON corrector (a corrector on a truncated reply would persist a partial batch as reviewed, defeating absence-as-retry). The resume machinery (the merged #543) covers relaunch-retry; deterministic in-run truncation is what the cap lift addresses.

Closes #538.

## Test plan

4 new tests (the #538 tracker class in `tests/test_issue532_llr_resume.py`): the truncated drop line + `batches_truncated == 1`; end-turn unbalanced not counted as truncated; the cap is `DEFAULT_MAX_TOKENS` (the fake records the kwarg); a truncated-but-salvageable reply persists nothing. The two `#294` tests migrated off the opaque-binding monkeypatch to the real call path.

## Verification evidence

| check | command | result |
|---|---|---|
| new + family | `pytest tests/test_issue532_llr_resume.py tests/test_issue294_reach_batch_diagnostics.py tests/test_llm_reachability.py -q` | 63 passed |
| full suite | `pytest tests/ -q` | 3959 passed, 2 failed — both pre-existing (SDK pin drift) |
| resume-safety | the checkpoint fingerprint unchanged (the cap is not a KEY member) — all 20 #543-era tests pass unchanged |

### DIFF
--- a/libs/openant-core/core/llm_reachability.py
+++ b/libs/openant-core/core/llm_reachability.py
@@ -282,6 +282,7 @@ def parse_response(
     on_error: Optional[Callable[[str], None]] = None,
     batch_label: Optional[str] = None,
     on_batch_drop: Optional[Callable[[], None]] = None,
+    stop_reason: Optional[str] = None,
 ) -> List[ReachabilitySignal]:
     """Parse a single LLM response into validated ``ReachabilitySignal``s.
 
@@ -304,9 +305,16 @@ def parse_response(
     data = _extract_json(response_text)
     if not isinstance(data, dict):
         shape = _classify_malformed(response_text)
+        # #538: the stop reason upgrades the brace heuristic to evidence —
+        # "truncated at max_tokens" (the model hit the cap) vs "unbalanced
+        # JSON (not truncated)" (end_turn + broken braces = a shape error).
+        truncated = (stop_reason == "max_tokens" and
+                     shape == "truncated or unbalanced JSON")
+        if truncated:
+            shape = "truncated at max_tokens"
         log(f"malformed response: {shape} — skipping batch{label};{snippet}")
         if on_batch_drop is not None:
-            on_batch_drop()
+            on_batch_drop(truncated=truncated)
         return []
 
     raw_signals = data.get("signals")
@@ -314,7 +322,7 @@ def parse_response(
         log(f"malformed response: 'signals' missing or not a list "
             f"(got {type(raw_signals).__name__}) — skipping batch{label};{snippet}")
         if on_batch_drop is not None:
-            on_batch_drop()
+            on_batch_drop(truncated=False)
         return []
 
     out: List[ReachabilitySignal] = []
@@ -432,7 +440,8 @@ def analyze_reachability(
 
     # Lazy import so this module stays usable when callers explicitly
     # provide a binding and never want the registry fallback above.
-    from utilities.llm import LLMAuthError, simple_text
+    from utilities.llm import (LLMAuthError, TextBlock, simple_completion,
+                            DEFAULT_MAX_TOKENS)
 
     # #532 (review-wave fix): the usage machinery must be live in PRODUCTION —
     # resolve the global tracker when the caller doesn't pass one (the family
@@ -450,6 +459,7 @@ def analyze_reachability(
     # in which some units were never reviewed.
     dropped_batches = 0
     units_not_reviewed = 0
+    batches_truncated = 0
 
     # ------------------------------------------------------------------
     # #532: resume/adopt machinery — the checkpoint family's own pattern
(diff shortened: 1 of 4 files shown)


gadievron:
**Gate-session receipt (the two-seat round: fable + astra; fable REFUTED as-shipped; the fold executed)**

**Re-derived independently:** the family 63/63; full suite 3773/0 fresh-venv; ruff clean. **Fable's blocker re-derived from source and CONFIRMED**: the salvage-success case — a max_tokens reply whose prefix parse succeeds — **applied** the parsed signals (the extend) but **skipped the persist** (the truncated_reply guard): applied ≠ adopted, the exact invariant the (4)-primitive exists to hold; and `batches_truncated` was **blind to the one case the gate acts on** (only the on_batch_drop callback incremented it; a salvage success never fires it).

**Astra: REFUTED** ("parse failure and truncation are independent dimensions; a parse-drop callback cannot be the sole truncation counter") — ruling **drop-whole** the coherent implementation: the rejection gate BEFORE the accepted-result mutation; the truncation counted independently; the salvaged content diagnostic-only.

**The fold (gate commit ee632e6)**: the truncation gate moved BEFORE the apply (a max_tokens reply `continue`s past the batch — the salvage prefix never reaches the applied set; applied==adopted restored); the counters deduplicated (the fold increments only when the salvage parse did not already drop — the first draft's 2-vs-1 test failure caught exactly this double-count); **the invariant pinned** by a NEW test (a well-formed truncated reply → zero applied signals, batches_truncated=1, batches_dropped=1). Family 25/25; full 3909/0.

**Receipt notes:** the unmapped-finish-reason interaction (unknown → max_tokens → never-persist) disclosed as the conservative fail-closed direction (the adapters' documented BUG-7/R2-C mapping) with astra's observability ask (the raw reason in the diagnostics) named as the follow-up; the #543-era resume suite green unchanged (the fingerprint non-membership of max_tokens verified).

Issue #435, opened 2026-08-30
analyze resume narration counts error checkpoint rows as 'already done': prints '0 units to process' while one runs, progress counter reaches 3/2


gadievron:
Observed on master @ 6bc3a2eb. Analyze resumed over a checkpoint dir holding two rows: one intact completed SAFE row, one #324-shaped error row (result with neither verdict nor finding). The unit is correctly retried — only the narration is wrong:

```
[Detect] Restored 2 units from checkpoints
[Detect] Mode: parallel (8 workers), 0 units to process (2 already done)
[Detect]  3/2  audit_budget.py:live_escalation_disabled  safe  [ThreadPoolExecutor-0_0]  5.6s  (elapsed 6s, ETA ~0s, $0.0058)
[Detect] Done: 3/2 units in 6s | avg 5.6s/unit | cost $0.03
```

`core/analyzer.py:227` computes `remaining = total - len(checkpointed)` and prints `len(checkpointed)` as "already done" from the RAW restored count. The retry queue at `core/analyzer.py:240-246` is built from the error-filtered count (`_cp_is_error`), so an errored row is counted as done in the narration and retried in the queue at the same time. `ProgressReporter" is also seeded with `completed=len(checkpointed)` (`core/analyzer.py:222`), so the retried unit pushes the counter past the total — the 3/2.

Repro: a checkpoint dir with one complete row plus one neither-verdict-nor-finding row, then `openant analyze <dataset> --checkpoint <dir>`. Any resumed run with errored units shows a remaining-count that understates by the error count and a counter that can exceed its denominator.

Suggested: derive done/remaining and the ProgressReporter seed from the error-filtered count — the same predicate the queue already uses — so "units to process" equals what actually runs.

gadievron:
Fixed in PR #472 (CI green, all 14 checks; leave-for-review — not merged). Closing per the campaign protocol; the PR body carries the full root-cause + evidence.

PR #551, opened 2026-09-08
fix(verify): the resume narration's 'already done' counts restored-complete rows, not the raw checkpoint store (#542)

### LINKED ISSUE #542: verify: the resume narration's "already done" is the raw checkpoint count — the #435 shape PR #472 removed from analyze, live on the verify sibling
verify: the resume narration's "already done" is the raw checkpoint count — the #435 shape PR #472 removed from analyze, live on the verify sibling

From the 2026-09-07 monitored e2e run and its verification rounds; re-derived from source at current master (ec6b063); pre-existing on master.

## The claim

`utilities/finding_verifier.py:834-837` prints the resume Mode line as:

```
[Verify] Mode: parallel (8 workers), {remaining} findings to verify ({len(checkpointed)} already done)
```

where `remaining = len(results_to_verify)` but "already done" is the RAW checkpoint count — which includes the errored rows being queued for retry. That is exactly the pre-#472 shape PR #472 removed from analyze: `remaining = total - len(checkpointed)` → `_done = total - len(units_to_process)` (`core/analyzer.py:250`), whose comment at `:248-249` asserts "finding_verifier.py computes the same quantity the same way" — true for `remaining`, false for `done`.

## Evidence (same run, a resumed pass)

`scan.stderr.log:685-688`:
```
[Verify] Running Stage 2 attacker simulation on 4 findings...
[Verify] Restored 1 findings from checkpoints
[Verify] Retrying 2 previously errored findings
[Verify] Mode: parallel (8 workers), 3 findings to verify (3 already done)
```

4 findings total; 1 restored-complete + 2 errored-being-retried; the Mode line says "3 to verify (3 already done)" — 3+3 over a population of 4. The per-unit ProgressReporter lines in the same log are correct; only the Mode line is wrong. PR #472's file list covers `core/analyzer.py` and its test — the verify path is not named.

## Direction

`done = total - remaining` (or seed it from the same quantity the retry queue derives), test with one restored-complete and one errored row: the Mode line must read "2 to verify (1 already done)" for the fixture above.

Related: #435 (the original), PR #472 (the analyze-side fix), the parity comment at core/analyzer.py:248-249 (its "same way" claim is the reason this sibling was assumed covered).

gadievron:
## Summary

The verify resume narration's Mode line printed `len(checkpointed)` — the raw checkpoint store **including the ERROR rows being re-queued for retry** — so a resumed run narrated "3 findings to verify (3 already done)" over a population of 4 (1 complete + 2 being retried). Exactly the pre-#472 shape PR #472 removed from `analyze` (the `#435` family, its verify sibling).

The fix:

- **"already done"** derives from the restored-complete counts (`_restored_ok + _restored_incomplete`) so done + remaining reconciles to the findings population — the narration, the "Restored N" line, and the progress-reporter seed (2/4, 3/4…) now agree;
- **the sibling (the review round): "Retrying N previously errored"** is now loop-tracked (`_errored_retried` — the cp rows for units actually in results that were error-shaped) instead of the raw store minus restored — a stale/foreign checkpoint row no longer narrates as an "errored retry" that never runs.

Closes #542.

## Test plan

2 tests in `tests/test_issue542_verify_narration.py`: the receipt shape (1 complete + 2 errored + 1 new → "3 findings to verify (1 already done)", never 3 already done); the foreign-checkpoint row narrates no retry.

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue542_verify_narration.py -q` | 2 passed |
| full suite | `pytest tests/ -q` | 3958 passed, 2 failed — both pre-existing (SDK pin drift) |
| adversarial review | combined wave+refute | SHIP + the sibling raw-store count folded (loop-tracked) |

### DIFF
--- a/libs/openant-core/tests/test_issue542_verify_narration.py
+++ b/libs/openant-core/tests/test_issue542_verify_narration.py
@@ -1,0 +1,123 @@
+"""Tests for issue #542 — the verify resume narration's "already done" is
+the raw checkpoint count.
+
+The Mode line printed ``({len(checkpointed)} already done)`` — the raw
+checkpoint store includes the ERROR rows being queued for retry, so a
+resumed run narrated "3 findings to verify (3 already done)" over a
+population of 4 (1 complete + 2 being retried). Exactly the pre-#472 shape
+PR #472 removed from analyze (the #435 family): done must derive from the
+restored-complete counts so done + remaining reconciles to the population.
+"""
+from __future__ import annotations
+
+import io
+from contextlib import redirect_stderr
+from unittest.mock import patch
+
+from utilities.finding_verifier import FindingVerifier
+
+
+def _ckpt_row(kind, finding="vulnerable"):
+    """kind: 'ok' a restored-complete row; 'err' a retryable ERROR row
+    (the shape _cp_is_error detects: correct_finding == 'error')."""
+    if kind == "err":
+        return {"finding": "error",
+                "verification": {"agree": False, "correct_finding": "error",
+                                 "explanation": "boom", "iterations": 1,
+                                 "total_tokens": 10},
+                "error": "LLMResponseError: boom"}
+    return {"finding": finding,
+            "verification": {"agree": True, "correct_finding": finding,
+                              "explanation": "ok", "iterations": 1,
+                              "total_tokens": 10}}
+
+
+def _row(uid):
+    return {"unit_id": uid, "finding": "vulnerable", "verdict": "vulnerable"}
+
+
+class _FakeCheckpoint:
+    def __init__(self, store):
+        self._store = store
+        self.dir = "/tmp/fake"
+
+    @property
+    def exists(self):
+        return bool(self._store)
+
+    def load(self):
+        return self._store
+
+    def write_summary(self, *a, **kw):
+        pass
+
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (proportionate combined round — astra; the micro-surface class)**

**Re-derived independently:** the 2 tests 2/2; the family (the #435 sibling + the #296/#521 neighbors) 26/26; full suite **3772/0** fresh-venv; ruff clean; the diff verified at the restore-loop + the Mode line.

**Astra: NOT-REFUTED** — the loop-tracked retry counter's logic verified airtight (a truthy non-errored cp_data takes the restore branch; truthy-in-the-else must be error-shaped; falsy excluded); the narration reconciliation correct (restored-incomplete belongs in "done" — its outcome was restored); the foreign-checkpoint row correctly outside scope (not in results → the loop never sees it → not counted as a retry — the old raw-store-minus-restored counted it).

**Receipt note:** the consumers of the old `len(checkpointed)` count not independently audited beyond the family suites (astra's caveat — no evidence of a behavioral consumer regression; the family green is the arbiter at this surface class).

CI green. Merge-ready.

PR #553, opened 2026-09-08
feat(scan): the Go scan command exposes the repository metadata flags the Python CLI already accepts (#539)

### LINKED ISSUE #539: The Go scan CLI cannot pass repository metadata the Python CLI already accepts — a bare-path scan's reports render [NOT PROVIDED] with no way to avoid it
The Go scan CLI cannot pass repository metadata the Python CLI already accepts — a bare-path scan's reports render [NOT PROVIDED] with no way to avoid it

From the 2026-09-07/08 e2e verification rounds; re-derived from source at current master (ec6b063); pre-existing on master.

## The claim

The Python scan command accepts `--repo-name`, `--repo-url`, and `--commit-sha` (openant/cli.py:1612-1614, and the same set on the build-output command at :1827-1830) and threads them into the report's repository metadata. The Go `openant scan` command exposes none of these flags (cmd/scan.go's flag set) — instead it passes them to the Python layer only when a project context exists (cmd/scan.go:222-232: `if ctx != nil && ctx.Project != nil`). But a positional-path invocation — the documented primary usage — resolves to a nil project context (cmd/resolve.go:59-62), so the metadata block never fires.

Net effect: for any bare-path scan, every report carries `url: ""`, `commit_sha: null` and the rendered `[NOT PROVIDED]` placeholders, and there is no CLI-level way to supply them. The comment at cmd/scan.go:222-223 ("Pass repository metadata from project context so reports don't show [NOT PROVIDED] placeholders") states the intent — but the only reachable path to it requires `openant init` project registration plus project-mode resolution, which a positional path actively suppresses.

## Evidence

A monitored scan of an external repo at a pinned commit (2026-09-07): the scan was invoked by absolute clone path; the resulting artifacts carry `"url": "", "commit_sha": null` and `[NOT PROVIDED]` in the summary — with the SHA known to the caller the whole time. The same invocation shape through the Python CLI's flags would have carried them.

## Direction

Either expose `--repo-name/--repo-url/--commit-sha` on the Go scan command (pass-through to the existing Python flags — the mechanism at cmd/scan.go:222-232 then works for path scans), or fall back to deriving them when the target is a git checkout (e.g. `git -C <path> rev-parse HEAD` + a configured remote) with the explicit flags taking precedence. Missing metadata should also be distinguishable from "the operator never passed it" in the report.

Related: the report-side rendering of these fields (core/reporter.py repository dict), and #521/#531 (the dynamic-test repo_path channel — a sibling of the same threading).

gadievron:
## Summary

A bare-path scan — the documented primary usage — resolves to a nil project context, so the repository-metadata pass-through block never fired: every report carried `[NOT PROVIDED]` url/commit with no CLI way to supply them (the receipt: a monitored scan of a pinned external clone with the SHA known to the caller the whole time; the Go scan command exposed none of the three flags the Python CLI already accepts).

The fix:

- `--repo-name` / `--repo-url` / `--commit-sha` on the Go `scan` command (and the `diff` wrapper, which shares the flag registration — the same repo, the same metadata semantics), forwarded to the Python flags that already exist;
- the precedence is explicit-flag-wins over the project context (the same shape `build-output` and `report` already use) — the context remains the fallback for project-mode invocations;
- the merge lives in a pure `resolveRepoMetadata` helper with a table test covering the four shapes, **including the nil-context path that is the reported defect**.

**Named follow-up (not this PR):** a bare-path scan of a git checkout still reports `commit_sha: null` unless the user passes the SHA — `init.go`'s `resolveLocalCommit` could supply a third-tier fallback (`flag > project > git rev-parse HEAD`).

Closes #539.

## Test plan

`TestScanRepoMetadataFlagsDefined` (the flag-parity pin, the analyze_flags precedent) + `TestResolveRepoMetadata` (4-row table: nil-ctx flags-through — the defect's own path; ctx-fills-empty; flag-wins; nil-Project-inside-ctx).

## Verification evidence

| check | command | result |
|---|---|---|
| Go tests | `go test ./cmd/ -run 'TestScanRepoMetadata\|TestResolveRepoMetadata' -v` | all pass |
| the cmd package | `go test ./cmd/` | ok |
| build + vet + fmt | `go build ./... && go vet ./cmd/ && gofmt -l cmd/` | clean |
| adversarial review | combined wave+refute | FIX-FIRST 1 (the source-grep test replaced by the extracted-helper table test — the nil-ctx regression now fails the suite); the diffCmd/build-output parity questions verified correct as-is |

### DIFF
--- a/apps/openant-cli/cmd/scan.go
+++ b/apps/openant-cli/cmd/scan.go
@@ -56,6 +56,9 @@ var (
 	scanLLMReachability             bool
 	scanLLMReachabilityMaxCodeBytes int
 	scanLibraryMode                 bool
+	scanRepoName                    string
+	scanRepoURL                     string
+	scanCommitSHA                   string
 )
 
 func init() {
@@ -65,6 +68,25 @@ func init() {
 // registerScanFlags wires the full scan-pipeline flag set onto cmd. Used by
 // scanCmd and by the thin diffCmd wrapper so that both surfaces accept the
 // same knobs.
+
+// resolveRepoMetadata merges the explicit CLI flags with the project
+// context (#539): the flag wins; the project context is the fallback; a
+// bare-path scan (nil context) uses the flags alone.
+func resolveRepoMetadata(name, url, sha string, ctx *projectContext) (string, string, string) {
+	if ctx != nil && ctx.Project != nil {
+		if name == "" && ctx.Project.Name != "" {
+			name = ctx.Project.Name
+		}
+		if url == "" && ctx.Project.RepoURL != "" {
+			url = ctx.Project.RepoURL
+		}
+		if sha == "" && ctx.Project.CommitSHA != "" {
+			sha = ctx.Project.CommitSHA
+		}
+	}
+	return name, url, sha
+}
+
 func registerScanFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&scanOutput, "output", "o", "", "Output directory (default: project scan dir or temp dir)")
 	cmd.Flags().StringVarP(&scanLanguage, "language", "l", "", languages.FlagHelp())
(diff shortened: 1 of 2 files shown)


gadievron:
**Gate-session receipt (proportionate combined round — astra; the micro-feature class)**

**Re-derived independently:** the tests 2/2 (the flag-parity pin + the 4-row resolver table, including the nil-ctx defect path); the full cmd package ok; build+vet clean; the gofmt hit on setup_test.go verified as **pre-existing on master** (a GOAPIFMT local difference, not this PR's diff); the Python flags verified at cli.py:1605-1607; the helper + the flag registration + the forward wiring verified from the diff.

**Astra: NOT-REFUTED** — with the notes: (a) the parity is scan-path parity, not a CLI-wide claim (the report/build-output precedent cited for the precedence shape, not audited for their own metadata surfaces — no remaining asymmetry demonstrated either); (b) the helper purity supported (the parameter-driven merge; the package-level flag vars passed as arguments); (c) the diff-wrapper flag sharing coherent (the same metadata semantics; the --commit-sha-identifies-vs-selects misunderstanding a help-text safeguard, not a rejection ground); (d) the resolveLocalCommit third-tier follow-up sound as a separate lever (the PR supplies the missing input path; the git-derive introduces discovery/failure-handling decisions of its own — the recorded follow-up preserves flag > project > git precedence).

CI green. Merge-ready.

PR #554, opened 2026-09-08
fix(llm-reach): provider-exception batches are counted — batches_failed, the shared units_not_reviewed, and the step's partial status (#541)

### LINKED ISSUE #541: llm-reachability: provider-exception batches are never counted in batches_dropped/units_not_reviewed — the step report can read success/0/0 for a pass that reviewed nothing
llm-reachability: provider-exception batches are never counted in batches_dropped/units_not_reviewed — the step report can read success/0/0 for a pass that reviewed nothing

From the 2026-09-07 monitored e2e run and its verification rounds; re-derived from source at current master (ec6b063); pre-existing on master.

## The claim

In `core/llm_reachability.py`, a batch whose LLM call raises (empty completion, transport error, rate-limit exhaustion) takes the provider-exception branch: `except Exception` builds the failure message, prints it to stderr, and `continue`s — BEFORE the `_count_drop` closure is ever defined or passed to `parse_response` (`:457-463` exit; `:465-476` the parse path where `dropped_batches`/`units_not_reviewed` increment, via `on_batch_drop`). The counters therefore count PARSE-path drops only. The scanner passes no `on_error` (`core/scanner.py:637-643`), so the exception message never reaches `ctx.errors` either — the step summary at `:871-872` can read `status: success, errors: [], batches_dropped: 0, units_not_reviewed: 0` while every unit in the exception-failed batches received no reachability review.

This is the exact report shape PR #386's body names as the harm it fixed ("`status: success, errors: [], units_reviewed: 10845`" while 10 of 434 batches were lost) — but #386's counting is scoped to `on_batch_drop`, the parse-level path. The provider-exception branch is named nowhere in that PR: not fixed, not deferred, not residual — an uncovered sibling path of the same family.

## Evidence

A monitored run's first attempt recorded four provider-exception batches ("returned an empty completion") in its event log — batches 1/216, 2/216, 3/216, 5/216 — before the attempt was killed; the report for that pass was never written, so the report-side consequence is source-derived from the code path, not from a surviving artifact. The same run's final pass, whose report survives, shows the counters working on the parse path (a malformed batch was counted). No surviving report anywhere counts the exception batches.

## Direction

Count the exception batch on the exception path: `units_not_reviewed += len(batch)` plus a `batches_failed` counter distinct from `batches_dropped` (different failure class, different remediation), and surface the failure message into the step's errors via the existing `on_error` hook so it reaches `ctx.errors` and the step report. Tests: an exception batch increments the failed/unreviewed counters; a malformed batch still increments the dropped counters; a mixed pass reports both.

Related: #294/#386 (the parse-path counting this extends), #538 (the deferred-recovery tracker for the malformed class), #537 (the same raise's cost-accounting sibling), #532 (the resume re-run surface).

gadievron:
## Summary

The `#386` counting covered the parse path only (`on_batch_drop` / `_count_drop`): a batch whose LLM call **raised** (empty completion, transport error) exited through the `except-continue` BEFORE the counter — so a step report could read `status: success, batches_dropped: 0, units_not_reviewed: 0` for a pass that reviewed nothing (the receipt: 4 empty-completion batches on the run's first attempt, invisible to every surviving counter).

The fix:

- a **distinct counter** for the exception class (`batches_failed` — a different failure class with different remediation than the parse drops) + the **same `units_not_reviewed` denominator** so the coverage gap is the visible sum;
- the **`#285`/`#376` partial-status contract**: the LLR step summary now carries `error_count = batches_dropped + batches_failed`, so all-batches-failed reads **partial**, never success/0/0 (the adversarial review's catch);
- both counters surface in the scanner's step summary.

Unchanged: the `#538` truncation subclass (`batches_truncated`) is orthogonal; the absence-as-retry semantics (an exception batch still leaves no checkpoint records — it re-runs on resume); the `#293` three-state summary arithmetic (the exception units stay in the incomplete bucket).

Closes #541.

## Test plan

4 tests: the exception batch counts (`batches_failed == 1`, `units_not_reviewed == 2`, the parse counter untouched); a parse drop NOT counted as failed; a mixed pass reports both classes with `units_not_reviewed` as the sum; the step-status derivation — the counters in the summary make `step_context` read partial (never success/0/0).

## Verification evidence

| check | command | result |
|---|---|---|
| new + family | `pytest tests/test_issue532_llr_resume.py -q` | 29 passed |
| full suite | `pytest tests/ -q` | 3995 passed, 2 failed — both pre-existing (SDK pin drift) |
| adversarial review | combined wave+refute | the refute's missed-item folded: the step-status gap (all-batches-failed reading success) closed via `error_count`; the #293 arithmetic and the auth-error class verified unchanged |

### DIFF
--- a/libs/openant-core/core/llm_reachability.py
+++ b/libs/openant-core/core/llm_reachability.py
@@ -460,6 +460,7 @@ def analyze_reachability(
     dropped_batches = 0
     units_not_reviewed = 0
     batches_truncated = 0
+    batches_failed = 0
 
     # ------------------------------------------------------------------
     # #532: resume/adopt machinery — the checkpoint family's own pattern
@@ -618,11 +619,20 @@ def analyze_reachability(
             # so the caller can stop and tell the user the key is bad.
             raise
         except Exception as exc:  # noqa: BLE001 — advisory stage; never crash pipeline
+            # #541: a provider-exception batch is counted in the coverage
+            # truth — the #386 counters covered the parse path only, so a
+            # step report could read success/0/0 for a pass that reviewed
+            # nothing (4 empty-completion batches on the receipt run were
+            # invisible to every surviving counter). A distinct counter
+            # (different failure class, different remediation) + the same
+            # units_not_reviewed so the coverage gap is the visible sum.
             msg = f"batch {i + 1}/{len(batches)} failed: {exc}"
             if on_error:
                 on_error(msg)
             else:
                 print(f"[LLMReach] {msg}", file=sys.stderr)
+            batches_failed += 1
+            units_not_reviewed += len(batch)
             continue
 
         # #532 (disclosed behavior change): valid ids are PER-BATCH, so a
@@ -757,6 +767,9 @@ def analyze_reachability(
         stats["units_not_reviewed"] = units_not_reviewed
         # #538: the truncation subclass — the recurrence's diagnosis lever.
         stats["batches_truncated"] = batches_truncated
+        # #541: the provider-exception class — distinct from the parse
+        # drops, same coverage-truth denominator.
+        stats["batches_failed"] = batches_failed
 
     return signals
 
--- a/libs/openant-core/core/scanner.py
+++ b/libs/openant-core/core/scanner.py
@@ -898,7 +898,15 @@ def scan_repository(
                         # #538: the truncation subclass — the diagnosis
                         # lever for the deferred-recovery tracker.
                         "batches_truncated": reach_stats.get("batches_truncated", 0),
+                        # #541: the provider-exception class — the coverage
+                        # truth the parse-path-only counters missed.
+                        "batches_failed": reach_stats.get("batches_failed", 0),
                         "units_not_reviewed": reach_stats.get("units_not_reviewed", 0),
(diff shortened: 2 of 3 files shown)


gadievron:
**Gate-session receipt (proportionate round — astra; the counting-surface class)**

**Re-derived independently:** the #532 family 29/29 (the 4 new tests included); the family (294 diagnostics + reachability + the #285 status contract) 48/48; full suite **3809/0** fresh-venv; ruff clean; the diff verified (the except-continue increment, the two stats keys, the scanner's `error_count = batches_dropped + batches_failed` wiring).

**Astra: NOT-REFUTED** — (a) the error_count sum is the ONE numeric source (the #285 contract's alternatives-not-additives shape); (b) the LLMAuthError re-raise-before-count correct (the designed abort, not a counted failure; the BaseException class escapes `except Exception` by construction); (c) **no double-count mechanism established** (the truncation `continue` and the provider-failure path are mutually exclusive per iteration — a raised call produces no completion for the truncation path to process); (d) the #293 three-state arithmetic unchanged (the exception units stay incomplete — the counters change observability, not classification).

CI green. Merge-ready.

PR #560, opened 2026-09-08
feat(scan): a bare-path scan of a git checkout stamps the working-tree HEAD — the flag > project > git-detected metadata tier (#557)

### LINKED ISSUE #557: A bare-path scan of a git checkout still reports commit_sha: null — resolveLocalCommit could supply the flag > project > git-detected fallback
A bare-path scan of a git checkout still reports commit_sha: null — resolveLocalCommit could supply the flag > project > git-detected fallback

From the 2026-09-08 #539 round (PR #553's named follow-up); re-derived from source at current master.

## The claim

`openant scan /path/to/git/checkout` (a bare-path scan — the documented primary usage) resolves to a nil project context, so the metadata pass-through fires only for explicitly-passed `--repo-url`/`--commit-sha`. The user must type the 40-char SHA by hand even though the target is a git checkout whose HEAD is one `git rev-parse` away — and `init.go` already has the machinery: `resolveLocalCommit` (`:276`) determines the commit SHA for a LOCAL git repo (used by the `init` flow at `:187`). The scan command never calls it.

`--repo-name` and `--repo-url` could also derive from the checkout's `git remote` when present.

## Evidence

A monitored scan of a pinned external clone with the SHA known to the caller the whole time still produced `"commit_sha": null` in the report (`core/reporter.py` renders `None`, not `""`) — the receipt from the #539 investigation. The `--commit-sha` flag (PR #553) fixes the manual path; this is the auto-detect tier.

## Direction

A three-tier precedence: explicit flag > project context > `git rev-parse HEAD` (and `git remote get-url origin` for the URL) when the target is a git work tree. `resolveLocalCommit`'s shape (it returns `sha, warn, err` — a warn channel for detached/dirty states) is the model; the dirty-worktree case should carry the warn into the report metadata rather than silently asserting a stale SHA.

Related: #539 (the flags this builds on), #553 (the explicit-flag tier).

gadievron:
## Summary

A bare-path scan of a git checkout (the #539 primary usage) still reported `commit_sha: null` — the user had to type the 40-char SHA by hand even though the target's HEAD is one `rev-parse` away (the receipt: a monitored scan of a pinned external clone, the SHA known to the caller throughout).

The fix:

- **`git.HeadSHA`** in `internal/git` (beside `CurrentBranch`; dedupes the private rev-parse copies; resolves **subdirectories** where init's `os.Stat(.git)` gate fails; an unborn repo errors at rev-parse — the empty return is the proof, no zero-guard needed);
- **`resolveRepoMetadata` gains the detected tier** (`flag > project > git-detected`) and a mismatch **warn** to stderr (the codebase's `PrintWarning` pattern): the legitimate CI case (an explicit `--commit-sha` in a synthetic-merge checkout — the #539 flag-wins contract unchanged) and the stale-project footgun are both surfaced, the latter with the remedy (`re-run openant init to refresh the project SHA`);
- **the `nogit` sentinel** never reaches the SHA (it was flowing into `/blob/nogit/…` permalinks and SARIF `revisionId`s via the project tier);
- **detection runs after `resolveScanMode`** (the `--pr` path checks out the PR head in `mode.go`'s `FetchPR` — hoisting detection above would stamp the pre-checkout SHA).

Scoped out (named): dirty-worktree warning (init does not do it either — a divergence would be the opposite of reuse); the URL/remote tier (scp-form remotes and credential-bearing URLs need normalization first); recording HEAD on a possibly-dirty tree is **not** a new assertion — the incremental manifest already stamps `rev-parse HEAD` as `diff.head_sha` with no dirty check.

Closes #557.

## Test plan

The 10-row `resolveRepoMetadata` table (the #539 rows + the detected tier: bare-path detection, non-git empty, the project/flag/CI mismatch warns, the remedy hint, `nogit`) + `TestHeadSHA` (a real repo, a subdirectory, a non-git dir, an unborn repo).

## Verification evidence

| check | command | result |
|---|---|---|
| Go tests | `go test ./cmd/ -run 'TestResolveRepoMetadata\|TestHeadSHA\|TestScanRepoMetadata' -v` | all pass |
| the cmd package | `go test ./cmd/` | ok |
| build + vet + fmt | `go build ./... && go vet ./cmd/` | clean |
| adversarial review | combined wave+refute | the 4 folds: the wrong checkout comment (mode.go, not prepareDiffManifest), the dead unborn guard deleted (rev-parse errors — the zero-id claim was false about git), the missing CI table row, the remedy hint |

### DIFF
--- a/apps/openant-cli/cmd/scan.go
+++ b/apps/openant-cli/cmd/scan.go
@@ -69,10 +69,18 @@ func init() {
 // scanCmd and by the thin diffCmd wrapper so that both surfaces accept the
 // same knobs.
 
-// resolveRepoMetadata merges the explicit CLI flags with the project
-// context (#539): the flag wins; the project context is the fallback; a
-// bare-path scan (nil context) uses the flags alone.
-func resolveRepoMetadata(name, url, sha string, ctx *projectContext) (string, string, string) {
+// resolveRepoMetadata merges the explicit CLI flags, the project context,
+// and the git-detected working-tree HEAD (#539 + #557): the flag wins; the
+// project context is the middle tier; a bare-path scan of a git checkout
+// falls to detection (a nil context is the #539 primary usage; a non-git
+// path detects empty and keeps whatever the earlier tiers supplied).
+// The returned warn is non-empty when the winning SHA disagrees with the
+// detected HEAD (an explicit flag in a CI synthetic-merge checkout is
+// legitimate; a stale project SHA is the footgun — the caller prints it).
+// The "nogit" sentinel (init records it for non-repo paths) is treated as
+// empty so it never flows into permalinks or SARIF revisionIds.
+func resolveRepoMetadata(name, url, sha string, ctx *projectContext, detectedSHA string) (string, string, string, string) {
+	flaggedSHA := sha != ""
 	if ctx != nil && ctx.Project != nil {
 		if name == "" && ctx.Project.Name != "" {
 			name = ctx.Project.Name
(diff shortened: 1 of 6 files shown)


gadievron:
**Gate-session receipt (proportionate round — astra; the micro-feature class, #553's named follow-up)**

**Re-derived independently:** the 3 test groups pass (the 10-row resolver table incl. the detected tier + `TestHeadSHA` with the real-repo/subdirectory/non-git/unborn shapes + the #539 flag-parity pin); the cmd package ok; build+vet clean; the diff verified (the `HeadSHA` helper deduping the rev-parse copies; the three-tier merge; the mismatch warn with the re-run-init remedy; the `nogit` sentinel filtered at the project tier; **the detection-after-`resolveScanMode` ordering** — the `--pr` checkout hazard the review round caught, now pinned by comment at the call site).

**Astra: NOT-REFUTED** — with the sharpenings receipt-noted: (a) the subdirectory resolution correct (`git -C <subdir>` discovers the containing repo — the fix over init's `os.Stat(.git)` gate); the ""-collapse (non-git/unborn/git-missing) acceptable for best-effort enrichment with the caller's uniform absence treatment; (b) **the `nogit` sentinel filtering verified project-tier-only** — an explicit `--commit-sha nogit` from the user remains authoritative (the deliberate-override preserved; the sentinel claim read narrowly: the automatic propagation is gone); (c) the mismatch warn factually right even for the deliberate CI case (the metadata doesn't change the scanned checkout) — the source-aware refinement named as usability follow-up, not a blocker; (d) the `--pr` ordering correct; no `--diff-base` hazard established (the comparisons don't move HEAD); (e) the dirty-worktree scoping sound (the manifest's HEAD-without-cleanliness precedent; the warning a separable improvement).

CI green. Merge-ready.

Issue #212, opened 2026-08-03
Stage 2 loses ~30% of candidates to provider moderation refusals, with no aggregate reporting


NahumKorda:
## Summary

Stage 2 verification returned a verdict for fewer than a third of the candidates it was given. The rest were refused by the inference provider's moderation layer or recorded as needing manual review. Stage 2 consumed roughly **three-quarters of total run cost** while adjudicating **29%** of its input, and no aggregate coverage figure is surfaced anywhere in the console output or the summary report.

## Evidence

Refusal rate by phase — same route (OpenRouter) and same model (Claude Opus 5) in both cases, single run:

| Phase | Units | Refused | Rate |
|---|---|---|---|
| Stage 1 detection | 1,349 | 2 | **0.15%** |
| Stage 2 verification | 467 | 139 | **29.8%** |

Of 467 Stage 1 candidates, Stage 2 produced a verdict for 135 (82 upheld, 53 disagreed, 1 overturned); 139 errored and 192 were flagged as needing manual review — **331 unadjudicated**.

The Stage 2 rate was stable across the phase, sampled every 50 candidates: 32.0%, 33.7%, 33.3%, 31.3%, 33.0%, 30.8%, 30.7%. It is a property of the phase, not a transient. Refusals arrive from OpenRouter as `finish_reason='content_filter'` and are correctly mapped to a typed refusal error by the adapter. Note both phases used the *same* model and route, so the difference is attributable to prompt content rather than to model or provider choice.

The asymmetry has a plausible mechanism: Stage 1 asks the model to *classify* code, whereas Stage 2 asks it to *construct a working exploit* — which is what content moderation exists to block. One trivial utility function was refused deterministically on three separate occasions across two runs, so specific inputs reliably trip it rather than it being random sampling.

## Why it matters

This is the finding I would prioritise, because it undercuts the core value proposition rather than a peripheral feature. "What survives is real" becomes "what survives is real, and roughly a third was never tested" — and a user has to read the JSON to discover which. `verify.report.json` exposes `error_count` but nothing frames it as a coverage percentage, and the console summary omits it entirely.

## Suggested fix

1. **Report Stage 2 coverage as a headline figure** — e.g. "328/467 adjudicated (70%); 139 refused by provider moderation." This is cheap and would have changed how we read our own results.
2. **Retry a refused unit once on a configured fallback binding.** Moderation behaviour differs between routes and vendors, so a second attempt elsewhere often succeeds.
[...]

PR #563, opened 2026-09-09
fix(llm): the empty-completion raise's cause clause branches on the finish reason it already carries (#561)

### LINKED ISSUE #561: llm adapters: the empty-completion raise's cause clause is constant ("filtered or malformed") regardless of the finish reason it now carries
llm adapters: the empty-completion raise's cause clause is constant ("filtered or malformed") regardless of the finish reason it now carries — for finish_reason='length' the chat path contradicts the module's own Responses-path wording

Re-derived from source at current master; pre-existing.

## The claim

`utilities/llm/providers/openai.py:912-917` raises `"{adapter} returned an empty completion (no text or tool calls; finish_reason={raw_finish!r}); the request may have been filtered or the response was malformed"`. Since #527 the message carries `raw_finish`, but the cause clause does not branch on it. When `raw_finish == "length"` the reply was not filtered or malformed — the output budget was consumed before any visible content (the reasoning-model shape #512 root-caused), and this module already says so on its other path: `openai.py:810-814` (`_responses_to_unified`) raises "no usable content … may have been truncated (reasoning consumed the budget) or filtered". `anthropic.py:388-392` has the same constant clause beside `stop_reason={raw_stop!r}`.

The string is persisted verbatim into `verify_checkpoints/*.json` (`verification_note`, `error`) and `results_verified.json`, so an operator reading a `finish_reason='length'` row is pointed at moderation/gateway — the #212 lesson (the "moderation layer" phrasing sent that reporter chasing a route change) in a new coat.

## Direction

Branch the clause on the raw finish — `length` (chat) / `max_tokens` (anthropic) → "the output budget was consumed before any visible content (reasoning models spend it on hidden reasoning)"; every other value → keep "filtered or malformed". Keep the `finish_reason=` interpolation and the #537 usage fields. One test per branch on the existing empty-completion fixtures.

## Not evidence for this issue

The 2026-09-07 rust-meilisearch verify receipt (`OpenRouterAdapter returned an empty completion (no text or tool calls); the request may have been filtered or the response was malformed`, tree fd82c57, pre-#527) carries no finish_reason and its checkpoint records usage 0/0 (pre-#537); its class is unknowable. Motivating observation only — the issue stands on source.

Related: #512/#527 (added the interpolation), #537 (usage on the raise), #212 (never assert who filtered without evidence).

gadievron:
## Summary

The chat path interpolated `finish_reason={raw_finish!r}` (#527) but the cause clause stayed constant ("may have been filtered or the response was malformed") regardless — for `finish_reason='length'` the reply was not filtered or malformed: the output budget was consumed before any visible content (the #512 reasoning-model shape), and the module's own Responses path already says so for its truncation arm ("may have been truncated (reasoning consumed the budget) or filtered"). The string is persisted verbatim into verify checkpoints and `results_verified.json`, so an operator reading a length-stop row was pointed at moderation/gateway — the #212 misdirection class in a new coat.

The fix: the openai chat path and the anthropic path **branch the cause clause on the raw finish/stop reason** — `length` / `max_tokens` → "the output budget was consumed before any visible content (reasoning models spend it on hidden reasoning)"; every other value keeps the honest non-assertion (#212: never say WHO filtered). The #537 usage fields and the #527 interpolation are unchanged. The no-choices sibling site keeps the constant clause deliberately (with no choices there is no finish reason to branch on). Bedrock/Ollama/OpenRouter inherit the fix through the shared translators; Gemini's existing combined wording ("may have been truncated (a thinking model consumed the token budget) or filtered/malformed") is the honest non-assertion already — untouched.

Closes #561.

## Test plan

5 tests: the openai chat raise branches (length → budget; stop → filter wording); the anthropic raise branches (max_tokens → budget; end_turn → filter); and the classifier contract — both new budget-wording strings stay retryable on the string branch, the typed `empty_completion` dict stays retryable (the contract this diff could have broken, pinned at the review's request).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue561_cause_clause.py -q` | 5 passed |
| full suite | `pytest tests/ -q` | 3997 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| consumers | the three string-matching readers re-derived: `is_retryable_error` / `_build_error_info` key on "empty completion"/"no usable content" (both survive in the new strings); the refusal guard untouched | classification unchanged |
| adversarial review | combined wave+refute | SHIP; the classifier-contract fold added; the retry-semantics follow-up (a budget exhaustion retried as transient) and the anthropic #537 usage asymmetry named as separate follow-ups |

### DIFF
--- a/libs/openant-core/tests/test_issue561_cause_clause.py
+++ b/libs/openant-core/tests/test_issue561_cause_clause.py
@@ -1,0 +1,159 @@
+"""Tests for issue #561 — the empty-completion raise's cause clause
+branches on the finish reason it already carries.
+
+The chat path's raise interpolates `finish_reason={raw_finish!r}` (#527)
+but the cause clause stays constant ("may have been filtered or the
+response was malformed") regardless — for `finish_reason='length'` the
+reply was not filtered or malformed (the output budget was consumed
+before any visible content; the reasoning-model shape #512 root-caused),
+and the module's own Responses path already says so ("may have been
+truncated (reasoning consumed the budget) or filtered"). The string is
+persisted verbatim into verify checkpoints and results_verified.json, so
+an operator reading a length-stop row is pointed at moderation/gateway —
+the #212 misdirection class.
+
+The fix: the chat adapter branches the clause on the finish reason
+(`length` → the budget wording); the anthropic adapter's identical
+constant clause branches on `stop_reason == "max_tokens"`.
+"""
+from __future__ import annotations
+
+import pytest
+
+from utilities.llm import LLMResponseError
+from utilities.llm.providers.openai import OpenAIAdapter
+from utilities.llm.providers.anthropic import AnthropicAdapter
+
+
+class _Usage:
+    def __init__(self, p=10, c=5):
+        self.prompt_tokens = p
+        self.completion_tokens = c
+
+
+class _ChatResp:
+    def __init__(self, finish, usage=None):
+        self.usage = usage
+        self.choices = [type("C", (), {
+            "message": type("M", (), {"content": None,
+                                      "tool_calls": None})(),
+            "finish_reason": finish})()]
+
+
+class _AnthResp:
+    def __init__(self, stop):
+        self.stop_reason = stop
+        self.content = []
+
+
+def _openai_client(finish):
+    class _C:
+        class chat:
+            class completions:
+                @staticmethod
+                def create(**kw):
+                    return _ChatResp(finish)
+    return _C()
+
+
+class TestOpenAICauseClause:
+    def test_length_stop_names_the_budget(self, monkeypatch):
+        """finish_reason='length' → the cause clause names the consumed
+        output budget, never 'filtered or malformed'."""
+        adapter = OpenAIAdapter(api_key="sk-test",
+                                _client=_openai_client("length"))
+        from utilities.llm import Message, TextBlock
+        with pytest.raises(LLMResponseError) as ei:
+            adapter.complete(
+                model="gpt-4o-mini", system=None,
+                messages=[Message(role="user", content=[TextBlock("x")])],
+                max_tokens=100)
+        msg = str(ei.value)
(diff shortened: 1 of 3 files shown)


gadievron:
**Gate-session receipt (proportionate round — astra; the #527/#537/#212 intersection)**

**Re-derived independently:** the 5 tests 5/5; the family (#512 truncation + #537 usage + #212 refusal + the SDK-contract floor) 72/72; full suite **3811/0** fresh-venv; ruff clean. **The consumers re-derived by construction myself**: both new full raise strings (the budget wording on each adapter's vocabulary) still hit the retryability substring readers (`rate_limiter.py:267/:309` — "no usable content"/"empty completion" both survive; the "refused the request" guard untouched) — the classification unchanged, as claimed.

**Astra: NOT-REFUTED** — with one decisive demand I resolved from source: **the branching predicate uses the RAW provider values, not the normalized ones** (`raw_finish = getattr(choice, "finish_reason", None) or "stop"` at openai.py:884 / `raw_stop` at anthropic.py:368 — both sites 30+ lines ABOVE the branch; the BUG-7 unknown→max_tokens normalization happens LATER at the `stop_reason=` construction, :936-946). So an UNKNOWN finish reason keeps its raw value at the branch, fails the `== "length"` check, and gets the honest non-assertion — **no normalized-lie** (the conservative mapping never reaches the cause clause).

**Receipt notes:** the checkpoint-persistence question (old constant-clause vs new branched-clause strings coexisting) — no exact-wording reader demonstrated (the retryability substrings + the refusal marker preserved; the field treated as opaque text by the surviving consumers); the no-choices sibling correct (no finish reason to branch on); the absent-finish-reason shape defaults to "stop" → the filter wording (correct: the non-assertion, not the budget diagnosis); both named follow-ups (the retry-semantics policy question; the anthropic #537 usage asymmetry) correctly scoped out.

CI green. Merge-ready.

PR #566, opened 2026-09-09
feat(scan): the repo-url metadata tier — the origin remote derived and NORMALIZED (scp→https, credentials never persist) (#562)

### LINKED ISSUE #562: Deriving --repo-url from the git remote needs normalization first — scp-form remotes produce broken report permalinks and non-URI SARIF repositoryUri values; credential-bearing remotes leak into reports
Deriving `--repo-url` from the git remote needs normalization first — scp-form remotes produce broken report permalinks and non-URI SARIF repositoryUri values; credential-bearing remotes leak into reports

From the 2026-09-08 #557 round (the need-check split the URL tier out of the metadata PR); re-derived from source at current master.

## The claim

`--repo-url` is currently caller-supplied only (or project-context-supplied). Deriving it from the scanned checkout's `git remote get-url origin` would close the same gap #557 closed for the SHA — but the naive derivation is unsafe:

- **scp-form remotes** (`git@github.com:org/repo.git`): the report's `FileURL` only strips a `.git` suffix (`internal/report/types.go:166-168`), so a scp-form URL produces `git@github.com:org/repo/blob/<sha>/f` — a broken permalink; and SARIF's `repositoryUri` (`internal/report/sarif.go:73`) expects a URI — a non-URI value can break Code Scanning upload.
- **Credential-bearing remotes** (`https://user:TOKEN@host/...`): stamped verbatim into the HTML report and the SARIF output. A security tool leaking a token into its own report is a real defect class.
- Multiple remotes (which one?), no `origin` (a local-only repo → no URL, fine), local-path remotes.

## Direction

The normalization spec (from the #557 need-check): ssh→https normalization (`git@host:org/repo.git` → `https://host/org/repo`), userinfo stripping (never persist credentials), `org/repo` extraction (which could also feed `--repo-name`). Accept only `http(s)` URLs with empty userinfo; anything else after normalization → skip silently (the current honest absence). A parser test matrix over the remote shapes (scp, https+creds, multiple, none, local-path).

Related: #557 (the SHA tier this completes), #539 (the flags), PR #553 (the explicit-flag tier).

gadievron:
## Summary

Two halves, per the need-check ruling (the fix before the feat):

**The normalizer (the fix — a bug reachable today).** `openant init` on an scp-form remote (`git@host:org/repo.git`) records the raw string; the report's `FileURL` only strips a `.git` suffix, producing broken permalinks, and SARIF's `repositoryUri` receives a non-URI — and a credential-bearing remote (`https://user:TOKEN@host/...`) is stamped verbatim into the HTML report and the SARIF output. `git.NormalizeRemote` (pure):

- scp-form → https with the `scpHost` last-`@` parsing rule (`git@evil@127.0.0.1:path` resolves to `127.0.0.1`);
- `ssh://` → https with **the ssh port dropped** (it is sshd's, not the browse UI's); **userinfo stripped always** — the same guarantee extends to the warn lines (normalized forms only);
- http(s) preserved, host lowercased in both branches, query/fragment dropped, `.git` trimmed; the scheme branch parses via `net/url`;
- `git://`, local paths, Windows drives, anything unparseable → `""` (the honest absence).

**The tiers.** The project tier normalizes defensively at **both** consumers (scan's `resolveRepoMetadata` and build-output's `--repo-url` pass-through). The detection tier (`git.RemoteURL` beside `HeadSHA`) fills a never-set flag or an empty project tier **only** — an explicit flag that fails normalization keeps the honest absence, never a silent fall-through to detection behind the user's back. A flag-vs-origin mismatch warns in normalized forms.

**Named residual (out of scope):** the render boundary itself (`FileURL`/SARIF/`server.go`) still passes through whatever an old `pipeline_output.json` carries — a render-boundary defense wants the normalizer in a dependency-light package (`internal/git` pulls exec); the follow-up is noted.

Closes #562.

## Test plan

20-row normalizer matrix (scp, smuggled userinfo, ssh+port, tokens, `.git` suffixes, trailing slashes, uppercase scheme, http preserved, no-path, query-dropped, mixed-case host, Windows drive, `git://`, local path, empty) + the tier-order pins (detection fills, flag-wins incl. the unparseable-flag honest-absence shape).

## Verification evidence

| check | command | result |
|---|---|---|
| Go tests | `go test ./cmd/ -run 'TestNormalizeRemote\|TestResolveRepoMetadata' -v` | all pass |
| the whole tree | `go test ./...` | ok |
| build + vet + fmt | `go build ./... && go vet ./cmd/ ./internal/git/` | clean |
| adversarial review | combined wave+refute | 7 findings, ALL folded: the ssh-port leak (+ the test row flipped), the credential leak in the warn, the unparseable-flag silent fall-through (the #557 invariant restored), the buildoutput hazard (the same tier, now closed), the host-case mismatch, the http rewrite, the IPv6/Windows/query edges (net/url) |

### DIFF
--- a/apps/openant-cli/cmd/buildoutput.go
+++ b/apps/openant-cli/cmd/buildoutput.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"os"
 
+	"github.com/knostic/open-ant-cli/internal/git"
 	"github.com/knostic/open-ant-cli/internal/output"
 	"github.com/knostic/open-ant-cli/internal/python"
 	"github.com/spf13/cobra"
@@ -83,7 +84,14 @@ func runBuildOutput(cmd *cobra.Command, args []string) {
 		pyArgs = append(pyArgs, "--repo-name", buildOutputRepoName)
 	}
 	if buildOutputRepoURL != "" {
-		pyArgs = append(pyArgs, "--repo-url", buildOutputRepoURL)
+		// #562: normalize defensively (an init-recorded scp-form or
+		// credential-bearing remote must never reach the report permalinks
+		// or the SARIF repositoryUri verbatim — the same tier fix as the
+		// scan path's resolveRepoMetadata).
+		normalized := git.NormalizeRemote(buildOutputRepoURL)
+		if normalized != "" {
+			pyArgs = append(pyArgs, "--repo-url", normalized)
+		}
 	}
 	if buildOutputLanguage != "" {
 		pyArgs = append(pyArgs, "--language", buildOutputLanguage)
--- a/apps/openant-cli/cmd/scan.go
+++ b/apps/openant-cli/cmd/scan.go
@@ -80,6 +80,17 @@ func init() {
 // The "nogit" sentinel (init records it for non-repo paths) is treated as
 // empty so it never flows into permalinks or SARIF revisionIds.
 func resolveRepoMetadata(name, url, sha string, ctx *projectContext, detectedSHA string) (string, string, string, string) {
+	return resolveRepoMetadataFull(name, url, sha, ctx, detectedSHA, "")
+}
+
+// resolveRepoMetadataFull is resolveRepoMetadata with the #562 URL tier: a
+// bare-path scan of a git checkout derives its --repo-url from the origin
+// remote, NORMALIZED (scp→https, credentials stripped — a raw remote never
+// persists). The detected URL is the LAST resort (an explicit flag or a
+// project URL wins; the project tier also normalizes defensively — an init
+// recorded scp-form or credential-bearing remote would otherwise reach the
+// report permalinks and the SARIF repositoryUri verbatim).
+func resolveRepoMetadataFull(name, url, sha string, ctx *projectContext, detectedSHA, detectedURL string) (string, string, string, string) {
 	flaggedSHA := sha != ""
 	if ctx != nil && ctx.Project != nil {
 		if name == "" && ctx.Project.Name != "" {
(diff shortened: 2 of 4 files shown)


gadievron:
**Gate-session receipt (astra REFUTED the headline; the fold executed after the live probe)**

**Re-derived independently:** the 32 subtests (the 20-row matrix + the tier pins) green; the whole tree ok; build+vet clean; the normalizer verified from source.

**Astra's refutation — CONFIRMED LIVE by the orchestrator on the tree (probe tests run against the real `NormalizeRemote`):**
1. **The password-bearing scp shape LEAKED**: `user:pass@host:path` → `https://user/pass@host:path` — the credential text AND the raw host:path surviving into the report/SARIF surface (the mechanism: `scpHostRegion` splits at the first colon; the path region carries the password). **Folded (gate commit f5f6b23)**: a legitimate scp path (`org/repo.git`) never contains `@` — the normalizer now rejects any scp form whose path region carries `@` → the honest absence. The fold pin: the password shape → `""`; the legitimate scp and the last-`@` rule (`git@evil@127.0.0.1:path` → `127.0.0.1`) unchanged.
2. **The path-segment credential survives** (`https://user:pass@host/TOKEN/repo.git` → `https://host/TOKEN/repo`): the userinfo stripped correctly; a credential IN THE PATH has no general discriminator. **The scope corrected on the record**: the guarantee is userinfo + query/fragment removal, not every-possible-credential removal — the body's own narrower wording was right; any broader reading is now explicitly disclaimed here.

**Astra's other adjudications (receipt-noted):** the scheme-branch userinfo strip verified sound (https+port+path, ssh+userinfo); the last-`@` scp rule is git's own semantic (the `git@evil@127.0.0.1` example resolves to the ATTACKER's host per git's rule — not a normalization bug); the warn lines verified from source to interpolate `NormalizeRemote()` output only (the raw flag never reaches the log); the tier contract (the unparseable flag keeps the honest absence, no fall-through — the #557 invariant) supported by the tier pins; the `git://` absence correct per the honest-absence doctrine.

Post-fold: 33 subtests + the whole tree + build/vet/fmt clean. CI green on the folded head. Merge-ready.

Issue #292, opened 2026-08-19
`is_retryable_error` honours "no usable content" only on the string path, so `enhance` never retries the one error class that actually occurs


gadievron:
> **Evidence provenance (added 2026-08-22).** The run-derived figures in this issue come from a
> **private scan that a reader cannot reproduce** (`raptor-e2e-20260818`, a 9.85-hour run against
> a Python target). They are reported for completeness, not as independently checkable evidence.
> Every **code** claim below is pinned to public source at `b5019628` and *is* checkable; where a
> conclusion rests on the private numbers rather than on the code, treat it as my report of what
> that run showed. This label was missing when the issue was filed.

## Summary

`is_retryable_error` documents an empty-completion `LLMResponseError` ("... returned no usable
content ...") as retryable, and the string branch implements that. The dict branch does not: it
classifies the same exception as `api_status`, looks for a `status_code` that an empty-completion
raise never carries, and falls through to `False`. `analyze` passes a string and retries it;
`enhance` passes a dict and does not. On our run this was the **only** error class that occurred.

Secondary, same function: the string branch matches HTTP status codes by substring over the whole
error message, so any message containing `500`/`502`…`529` as a substring is classified retryable.

## Evidence

`libs/openant-core/utilities/rate_limiter.py:213-228` (excerpt — 6 interior lines elided) (HEAD `b501962`) — the documented contract:

```python
def is_retryable_error(error_info: dict | str | None) -> bool:
    """
    Check if an error is retryable (transient network/server issues).

    Retryable errors include:
    ...
    - no usable content: an adapter's "... returned no usable content ..."
      LLMResponseError (empty completion) — a transient thinking-truncation /
      malformed reply, across anthropic/google/openai. Deterministic refusals
      (LLMRefusalError, "refused the request") are NOT matched and stay
      non-retryable.
```

`rate_limiter.py:239-253` (excerpt — 3 interior lines elided, 3 of them comments) — the dict branch, which has no clause for it:

```python
    if isinstance(error_info, dict):
        error_type = error_info.get("type", "")
        if error_type in ("rate_limit", "connection", "timeout", "parse_error"):
            return True
        # Retry server errors (5xx), but not client errors (4xx)
        if error_type == "api_status":
            status_code = error_info.get("status_code", 0)
            return status_code >= 500
        return False
```

`utilities/context_enhancer.py:84-134` (`_build_error_info`) maps the exception to
[...]

PR #571, opened 2026-09-10
fix(analyze): a budget-exhausted empty completion retries once at a raised, SDK-safe cap (#569, choice c)

### LINKED ISSUE #569: A budget-exhaustion empty completion (finish_reason='length') is still retried as transient — the message now states the deterministic cause, but the classifier treats it as recoverable
A budget-exhaustion empty completion (finish_reason='length') is still retried as transient — the message now states the cause, but the classifier treats it as recoverable

From the 2026-09-09 #561 review round (the semantic tension surfaced — not created — by the cause-clause fix); re-derived from source at current master. A design question for the maintainers, filed so the decision has a home.

## The tension

`rate_limiter.py:306-316` classifies an empty-completion `LLMResponseError` as **retryable** ("typically a thinking-block truncation or a malformed/overloaded reply, which re-generating usually recovers" — the #292 decision, deliberately bounded to the single detection retry pass). Since #561, the message for a `finish_reason='length'` empty completion states **"the output budget was consumed before any visible content (reasoning models spend it on hidden reasoning)"** — a deterministic same-cap retry will very plausibly reproduce it. The message and the classification now visibly disagree: one says "the budget ran out", the other says "retry, it usually recovers".

The #292 rationale predates the split: it lumped all empty completions as "typically a thinking-block truncation" — but #561 separated the truncation class (budget exhaustion, deterministic at the same cap) from the filtered/malformed class (potentially transient). The classifier's single term now spans both.

## The decision to make

- (a) Keep as-is: the retry is bounded (one pass), and #292's trade ("a small billing under-report for verdict recovery") may still be worth it even for deterministic truncations — a re-roll occasionally lands under the cap.
- (b) Split: a `finish_reason='length'`/`stop_reason='max_tokens'` empty completion classifies as **non-retryable** (surface as ERROR, needs_review) — the classifier's message check could read the finish reason the message now interpolates.
- (c) The middle: retry once at a higher cap (the #512 approach on the retry path only).

The discriminating datum is in the message (or on the error object since #537's carriage), so (b) is mechanically cheap; the question is which semantic the maintainers want.

Related: #292 (the retry decision this revisits), #561 (the cause split that surfaced the tension), #512/#527 (the cap work), #537 (the error object's fields).

gadievron:
## Summary

The #292 decision classified every empty-completion error as same-cap-retryable — a rationale that predates #561's cause split. Since #561 the message for a length-stop empty states the **deterministic** cause ("the output budget was consumed before any visible content"), and the classifier retried it at the same cap anyway: a billed coin flip against a cause the message itself names.

The fix (the choice-(c) ruling — both review seats converged):

- `is_budget_exhausted_error` discriminates the class by the **budget wording across all adapters** (the openai chat + anthropic "output budget was consumed"; gemini-thinking "consumed the token budget"; the o-series Responses path "reasoning consumed the budget") — nothing is declassified; only the retry's **cap** differs;
- the analyze retry loop splits: the budget class retries once at `BUDGET_RETRY_MAX_TOKENS = min(2× default, 21000)` — raised above the default but **under the Anthropic non-streaming ceiling (~21,333**; the SDK rejects higher with "Streaming is required", which would make the retry a guaranteed failure on that adapter — the adversarial review's blocker, folded); every other retryable keeps the #292 same-cap re-roll. The retry log line names the cap used; the per-index decision is a testable production helper;
- the cap threads through `_process_unit → analysis_core.analyze_unit → simple_text` (`None` = the unchanged default, byte-identical);
- the contradictory stale CONC-C2 notes unified (post-#537/#564 the rejected reply's usage rides the error — a retry is billed *and* recorded).

**Named deferred:** `context_enhancer.py`'s retry loop is the same class (its single-shot call pins 4096) — a follow-up; a true 2× cap needs a provider-layer streaming/timeout change (the SDK ceiling).

Closes #569.

## Test plan

12 tests: the discriminator (the length/stop/transient/dict shapes + the three adapters' parity wordings + the filtered class not budget); the cap chain exists; the SDK-ceiling pin (above the default, at/below 21000); the production `budget_retry_cap` decision; the end-to-end split (a length-empty unit's retry carries the raised cap, a stop-empty unit's carries None).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue569_budget_retry.py -q` | 12 passed |
| full suite | `pytest tests/ -q` | 4029 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| adversarial review | combined wave+refute | BLOCK → the 5 findings folded: the Anthropic SDK ceiling blocker (the cap clamped to 21000, the constant renamed to say so), the dead test (the production helper extracted + the real drive), the parity gap (gemini/Responses markers), the per-iteration set, the contradictory stale note |

### DIFF
--- a/libs/openant-core/core/analysis_core.py
+++ b/libs/openant-core/core/analysis_core.py
@@ -309,7 +309,8 @@ def analyze_unit(
     use_multifile: bool = False,
     json_corrector: JSONCorrector = None,
     context_reviewer: ContextReviewer = None,
-    app_context: "ApplicationContext" = None
+    app_context: "ApplicationContext" = None,
+    max_tokens: int | None = None
 ) -> dict:
     """
     Analyze a single code unit.
@@ -395,7 +396,12 @@ def analyze_unit(
     # Call the configured analyze-phase model with the threat-model system prompt.
     start_time = datetime.now()
     system_prompt = get_stage1_system_prompt(app_context=app_context)
-    response = simple_text(binding, prompt, system=system_prompt)
+    # #569 (choice c): the budget-retry path passes a raised cap — the
+    # deterministic length-empty class gets one retry that attacks the
+    # cause (the budget) instead of a same-cap coin flip.
+    response = simple_text(binding, prompt, system=system_prompt,
+                           max_tokens=max_tokens) if max_tokens else \
+        simple_text(binding, prompt, system=system_prompt)
     elapsed = (datetime.now() - start_time).total_seconds()
 
     # Parse response
--- a/libs/openant-core/core/analyzer.py
+++ b/libs/openant-core/core/analyzer.py
@@ -36,7 +36,25 @@ from utilities.llm import (
 )
 from utilities.file_io import read_json, write_json
 from utilities.json_corrector import JSONCorrector
-from utilities.rate_limiter import get_rate_limiter, is_retryable_error
+from utilities.llm import DEFAULT_MAX_TOKENS
+from utilities.rate_limiter import (
+    get_rate_limiter,
+    is_budget_exhausted_error,
+    is_retryable_error,
+)
+
+# #569 (choice c): the budget-retry cap — raised above the analyze
+# default but under the Anthropic non-streaming ceiling (the SDK rejects
+# >~21,333 with "Streaming is required" — see helpers.py:29-31; the
+# adapters call non-streaming). The deterministic length-empty class gets
+# ONE retry that attacks the cause (the budget) within that envelope.
+BUDGET_RETRY_MAX_TOKENS = min(DEFAULT_MAX_TOKENS * 2, 21000)
+
+def budget_retry_cap(index: int, budget_set: set) -> int | None:
+    """#569: the per-index retry cap — the raised cap for the budget class,
+    None (the unchanged default) for every other retryable."""
+    return BUDGET_RETRY_MAX_TOKENS if index in budget_set else None
+
 
 # These live in core/ because core is shipped and experiment.py is not: importing
 # them from the research harness made `import core.analyzer` fail in any installed
@@ -119,7 +137,8 @@ def _apply_limit(units, limit):
     return prioritized[:limit]
 
 
-def _process_unit(binding: PhaseBinding, unit, index, json_corrector, app_context):
+def _process_unit(binding: PhaseBinding, unit, index, json_corrector, app_context,
+                  max_tokens=None):
     """Process a single unit for Stage 1 detection.
 
     Returns a dict with all result data. Does not mutate shared state.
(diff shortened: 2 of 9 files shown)


gadievron (commented):
**Verification review — recommend-merge.** (all evidence pinned to head `570ddf9`; re-verify on any new head)

**The fix**: `is_budget_exhausted_error` (rate_limiter.py) discriminates the deterministic budget-exhaustion empty completions by the adapter wording; `run_analysis`'s retry pass gives those units ONE retry at `BUDGET_RETRY_MAX_TOKENS = min(2×DEFAULT, 21000)` (under the Anthropic non-streaming ceiling), threaded through `_process_unit` → `analyze_unit` → `simple_text`. Every other retryable keeps the #292 same-cap rationale.

**The review round, fixed on this branch (`570ddf9`)**:
- The gemini and OpenAI-Responses empty-completion raises emitted the budget wording UNCONDITIONALLY, so the filtered/malformed class also took the raised cap on two of four adapters. Both producers now branch on the finish signal (google: `raw_finish` in the MAX_TOKENS forms; Responses: `stop_reason == "max_tokens"`), mirroring the openai-chat #561 pattern — the truncated wordings carry the marker, the filtered wordings do not. Producer-side pins added in both adapter suites (driving the real raises, positive and negative).
- The "end-to-end" retry-loop test re-implemented the loop and never invoked `run_analysis` — deleting the wiring left every test green. It now drives the REAL `run_analysis` retry pass (stubbed `_run_detection` + a `_process_unit` capture) and asserts the raised cap threads only into the budget-class unit. Must-trip verified: with the wiring line deleted the test fails; restored it passes.

**Verification** (run at the pinned head): targeted slices (issue569 + google/openai adapters + retry-empty-completion) — 107 passed; full suite — 4037 passed, 33 skipped, 0 FAIL; CI at `570ddf9`: tests + gitleaks success. Cross-PR: the combined tree of all six open PRs passed the full suite (3912, pre-delta); no retry stacking with #575 (the raised cap is analyze-phase-only; LLR failures go to `batches_failed` → absence-as-retry).

**Known trade-offs (documented, accepted by the review round)**: the raised cap is +5% over the 20000 default — efficacy-limited against deep reasoning exhaustion (an Anthropic-envelope constraint; a per-adapter ceiling would raise it for the other providers); the `min()` clamps non-Anthropic providers to Anthropic's ceiling.

**Follow-ups (named, none blocking)**:
- The Responses empty-content marker gate could key on the raw `incomplete_details.reason == "max_output_tokens"` instead of the derived `stop_reason` — the unknown-incomplete/unknown-status relabels also set `max_tokens`, so those exotic shapes carry the budget wording as a false causal statement (retry direction safe; message honesty only).
- The same deterministic budget class remains same-cap in verify/enhance and LLR (scoped out; the cheap LLR composition is to split the batch when `_attempt`'s exception arm sees `is_budget_exhausted_error`).
- The resume path re-runs budget-error checkpoint rows at the default cap before the retry pass (pre-existing shape).
- `test_retry_empty_completion.py`'s `_EMPTY_GEMINI` literal is stale vs the current producer wordings (non-load-bearing — retryability rides on the shared substrings); `_EMPTY_OPENAI_RESPONSES_FILTERED` is defined but not asserted.

PR #572, opened 2026-09-10
fix(context+checkpoints): the deterministic-derivation identity enters the resume keys — source_sha256 stamped by all three context arms, folded into the gated phases' fingerprints (#546)

### LINKED ISSUE #546: The checkpoint family's resume keys exclude the app-context content — a deterministic-context-sources fingerprint would close the stale-replay gap without the re-pay trap
Filed from the two-seat adversarial gate on PR #543 (fable + astra, both found it; the design's own disclosure names the residual).

## The gap

The llm-reach resume key (`fingerprint_for_binding` + the static template rendered with `app_context=None`, llm_reachability.py:510-516) and the projection hash (`unit_type + trimmed code`, :498-506) both exclude the app-context payload — correctly for the *narrated* text (LLM non-determinism → content-keying = the verified spurious-re-pay trap at LLR prices; the backend_identity ~17k-token lesson). But the app-context is **derived from deterministic repo sources** (`gather_context_sources`) and fed into every batch's prompt (`scanner.py:648 → analyze_reachability → _build_app_context_block`).

**The stale-replay shape**: a repo edit that changes the context derivation (e.g. a new entry-point-shaped file changing the app-type) while leaving an existing unit's own code identical → the unit's projection_sha matches, the key matches, the **stale signals replay under the changed analysis context**. The signals gate the re-filter, so this is an FN-risk class (a stale seed under a new context).

**The pre-existing scope**: the same exclusion exists in the analyze/verify/enhance resume keys (the same tier — the #543 disclosure says "same tier as analyze's"). This is a FAMILY-wide design question.

## The closure (fable + astra's converged recommendation)

Add a **deterministic-context-sources fingerprint** to the checkpoint family's identity key: hash `gather_context_sources`' output (the raw deterministic inputs), NOT the narrated text. Same-repo → same fingerprint → no spurious re-pay (the re-pay trap is avoided because the sources are deterministic); changed-repo-shape → a different fingerprint → the stale records archive and the pass re-runs. This can land as one change to `fingerprint_for_binding`'s call sites across the four stages, each with its own regression test.

## Acceptance criteria

- a new context source (a file that changes the app-type derivation) invalidates the resume keys across the family;
- a same-repo re-run (with the narrated text differing run-over-run) does NOT invalidate (the fingerprint is over the sources, not the narration);
- each stage's existing backend-identity tests stay green.

gadievron:
## Summary

The checkpoint family's resume keys excluded the app-context payload — correctly for the **narrated** text (the LLM non-determinism trap), but the context is derived from **deterministic repo sources**: a repo edit changing the derivation while leaving a unit's code identical replayed stale signals under the changed analysis context (they gate the re-filter — the FN-risk class the two-seat gate on PR #543 named).

The fix (the stamp-the-artifact architecture — no consumer re-derivation, no `repo_path` plumbing, no production/consumption race):

- **`context_sources_digest`**: the canonical sha over the gathered CONTEXT_FILES entries + the two degradation markers — deliberately **excluding** `[directory_structure]` (it lists the in-repo output dir; hashing it would self-invalidate on every resume that writes an artifact) and `[detected_patterns]` (rglob-order, cap-windowed — not canonical); those stay a named residual;
- **all three arms stamp `source_sha256` on the artifact**: the threat-model arm (the file's bytes — pre-existing), the manual-override arm (the override content), the LLM arm (the sources digest) — a **different narration over the same sources keeps the stamp** (no spurious re-pay);
- the gated phases fold it into their identity keys via `extra_key` (`ctx_sources_sha256`): **analyze** (the load-before-fingerprint reorder — the prior order computed the fingerprint before the context loaded, keying on `None` forever), **verify** (the standalone `--app-context` case), **llm-reach** (the dir-level fingerprint; the per-prompt narration stays excluded). A pre-fix artifact yields `None` → a one-time archive on first contact;
- `FINGERPRINT_SCHEME_VERSION` → **2** + the exclusion clause rewritten: the *narration* never enters; the *derivation identity* does.

Closes #546.

## Test plan

9 tests: the digest (same-files/same, content-change differs, new-context-file differs, **output-dir artifacts do not invalidate** — the self-invalidation hazard, untracked-py-under-results unchanged); the producer stamps (a source landing between calls differs; a different narration over the same sources keeps the stamp); the analyze fingerprint folds the sha (None vs sha differ; the pre-fix artifact loads yielding None).

## Verification evidence

| check | command | result |
|---|---|---|
| new tests | `pytest tests/test_issue546_ctx_fingerprint.py -q` | 9 passed |
| the family suites | `pytest tests/test_issue293_three_state_counters.py tests/test_issue532_llr_resume.py tests/test_llm_reachability.py tests/test_backend_identity.py -q` | 91 passed |
| full suite | `pytest tests/ -q` | 4026 passed, 2 failed — both pre-existing (SDK pin drift, verified on the pristine base) |
| adversarial design review | the fable need-check | the full-`gather_context_sources` hash rejected (the `[directory_structure]` self-invalidation + the `[detected_patterns]` order hazard); the consumer-side re-derivation rejected (the `repo_path` crux); the three-arm coverage gap (threat-model + manual-override arms never call `gather_context_sources`) closed by the stamp-the-artifact design |

### DIFF
--- a/libs/openant-core/context/application_context.py
+++ b/libs/openant-core/context/application_context.py
@@ -151,8 +151,13 @@ class ApplicationContext:
     # file is handed to the pipeline — the precondition for comparing them.
     # Provenance of a repo-supplied threat model, for scan-artifact visibility.
     # Both are additive/defaulted so save_context(asdict)/load_context(**data)
-    # round-trip unchanged. sha256 is over the raw file bytes; permissive_warnings
-    # is warn_permissive_threat_model's output, which was previously discarded.
+    # round-trip unchanged. #546 generalizes source_sha256 to the
+    # DETERMINISTIC-DERIVATION identity of whichever arm produced the
+    # context: the threat-model file's raw bytes (that arm), the override
+    # file's content (the manual arm), or the gathered CONTEXT sources'
+    # digest (the LLM arm) — the checkpoint family's resume keys fold it;
+    # permissive_warnings is warn_permissive_threat_model's output, which
+    # was previously discarded.
     source_sha256: str | None = None
     permissive_warnings: list = None
     threat_model_version: int | None = None
@@ -389,6 +394,27 @@ def gather_context_sources(repo_path: Path) -> dict[str, str]:
     return sources
 
 
+def context_sources_digest(sources: dict[str, str]) -> str:
+    """#546: the deterministic-derivation identity over the gathered
+    CONTEXT sources — the fingerprint the checkpoint family folds into its
+    resume keys (via the artifact's source_sha256).
+
+    Hashes ONLY the repo-derived CONTEXT_FILES entries and the two
+    degradation markers — NOT ``[directory_structure]`` (it lists the
+    in-repo output dir, so hashing it would self-invalidate on every
+    resume that writes a new artifact) and NOT ``[detected_patterns]``
+    (an rglob-order, cap-windowed list — not canonical). Those two stay a
+    named residual: a rename that flips the LLM's classification without
+    touching the hashed sources does not invalidate.
+    """
+    import hashlib
+    entries = {k: v for k, v in sources.items()
+               if not k.startswith("[directory_structure]")
+               and not k.startswith("[detected_patterns]")}
+    payload = json.dumps(entries, sort_keys=True, separators=(",", ":"))
+    return hashlib.sha256(payload.encode("utf-8")).hexdigest()
+
+
 def get_directory_structure(repo_path: Path, max_depth: int = 2) -> str:
     """Get directory tree for pattern recognition.
 
@@ -580,6 +606,16 @@ def _application_context_from_override(data: Any, filename: str) -> ApplicationC
         )
     data["override_warnings"] = override_warnings
     data["override_filename"] = filename
+    # #546: the override arm's deterministic input IS the override file —
+    # sha over the raw content (the family folds it via the artifact).
+    if not data.get("source_sha256"):
+        import hashlib as _h
+        try:
+            _raw = json.dumps(data, sort_keys=True, separators=(",", ":"))
+            data["source_sha256"] = _h.sha256(
+                _raw.encode("utf-8")).hexdigest()
+        except (TypeError, ValueError):
+            data["source_sha256"] = None
     known = {f.name for f in fields(ApplicationContext)}
     unknown = [k for k in data if k not in known]
     if unknown:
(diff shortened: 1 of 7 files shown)


gadievron (commented):
**Verification review — recommend-merge.** (all evidence pinned to head `3f6813b`; re-verify on any new head)

**The fix**: `context_sources_digest` — the canonical sha over the gathered CONTEXT sources (excluding `[directory_structure]`, which self-invalidates, and `[detected_patterns]`, which is rglob-order-dependent — a named residual); all three context arms stamp `source_sha256`; the three identity-gate sites (analyze, verify, LLR) fold `ctx_sources_sha256` into their fingerprints; `FINGERPRINT_SCHEME_VERSION` 1→2; the app context now loads BEFORE the analyze fingerprint. The narration stays excluded (the verified ~17k-token re-pay clause is amended, not violated) — only the deterministic derivation invalidates.

**The review round, fixed on this branch (`3f6813b`)**: the LLM arm and the override arm HONORED a supplied `source_sha256` (`if not data.get(...)`) — a scanned repo could steer the model's JSON (the sources text is attacker-influenceable) or author an OPENANT.json that pins a constant sha, freezing the checkpoint identity against future source edits — re-opening exactly the stale replay #546 exists to kill. Both arms now compute UNCONDITIONALLY (LLM: the sources digest; override: sha over the serialized override data). Negative tests pin it: two different supplied constants + the same sources yield the same DERIVED digest (neither constant adopted).

**Verification** (run at the pinned head): targeted slices (issue546 + issue293) — 24 passed; full suite — 4030 passed, 33 skipped, 0 FAIL; CI at `3f6813b`: tests + gitleaks success. Cross-PR: the combined tree of all six open PRs passed the full suite; the #572×#575 same-file (`llm_reachability.py`, disjoint regions) merge is clean; this is the only PR bumping the scheme version.

**Known cost (for the approver)**: the scheme bump 1→2 invalidates every pre-existing analyze/verify/LLR sidecar — a one-time re-pay for all users, including no-context users whose old checkpoints were valid (policy-conformant per backend_identity's own bump rule; not strictly correctness-required, since a post-fix scan's key already differs).

**Follow-ups (named, none blocking)**: the verifier/LLR fold sites have no dedicated wiring tests (only `_analyze_fingerprint` is pinned); the LLR site reads the artifact dict while analyze/verify read the dataclass attribute (type asymmetry — safe today since the scanner passes the raw dict); a YAML override with a non-JSON-serializable scalar silently degrades `source_sha256` to `None` and skips the fold for that run (`json.dumps(..., default=str)` would keep it derived); unhashed renames that flip `detected_patterns` classification remain the named residual.

PR #575, opened 2026-09-10
feat(llm-reach): the in-pass split-and-retry for dropped batches — a malformed batch re-issued once as two halves, the recovery carrying its own provenance (#558)

### LINKED ISSUE #558: llm-reachability: the in-pass recovery for malformed/failed batches is unmeasured — split-and-retry is the candidate; a JSON corrector on a truncated reply must be rejected
llm-reachability: the in-pass recovery for malformed/failed batches is unmeasured — split-and-retry is the candidate; a JSON corrector on a truncated reply must be rejected

From the #538 tracker's direction (3) — named as the "measured follow-up" in PR #550's body and the tracker's close; this issue is its home so the deferral chain ends in a tracked place.

## The claim

A malformed or provider-exception batch's units receive no reachability signals for that pass (they are counted — #541 landed `batches_failed`; the #538 round landed the evidence-based stop-reason diagnosis + the private-cap lift). The resume machinery (PR #543) re-runs them on relaunch, but an in-run failure finishes the pass with the gap, and a relaunch under the same conditions may replay the same deterministic failure. An in-pass recovery would re-attempt the batch within the same run.

## The measured-before-adoption requirement

- The cap lift (#550) should already have removed most of the truncation class — the residual malformed population (prose refusals, bare arrays, `signals` non-list) is a DIFFERENT population; measure the residual rate on a real corpus first. If the rate is ~0, this issue closes as satisfied-by-#550 with the measurement cited.
- **Split-and-retry is the candidate**: re-issue the failed batch as two halves (re-run at the same tier), measurable as recovered-units/extra-calls. A recovered batch must never overwrite `units_not_reviewed` without provenance (#538's direction (4) — carried into any implementation).
- **A JSON corrector is the WRONG tool for the truncation class and must be rejected**: the existing `JSONCorrector` is verdict-shaped (`json_corrector.py:209-328`, returns a vuln-shaped error dict on failure) and, more fundamentally, a corrector on a truncated reply can only salvage the PREFIX — yielding a partial batch whose tail units would persist as "reviewed, no signal" in the checkpoint (PR #543's records), converting a visible drop into a FROZEN coverage gap that absence-as-retry never revisits. The applied==adopted invariant forbids it.

## Direction

1. Measure the residual malformed rate under `DEFAULT_MAX_TOKENS` on a real ≥2-language corpus (the #550 PR's body names the measurement).
2. If non-trivial: implement split-and-retry (halve + re-issue; the recovery counter distinct from dropped/failed) with the provenance guard.
3. Never the corrector for the truncation class.

Related: #538 (the tracker this completes), #541 (the exception counters), PR #543 (the resume machinery), #550 (the cap lift + the diagnosis), #294/#386 (the counting ancestry).

gadievron:
## Summary

The post-lift measurement (this issue's step 1, run 2026-09-09, receipts in the issue comment): **4 malformed batches in ~95 at the `DEFAULT_MAX_TOKENS` cap — all the end_turn broken-JSON finish class, 0 max_tokens** — the cap lift eliminated the truncation-at-cap class; the residual is the model's own structurally-broken finishes, a fresh-roll class. The close-out rule failed → the in-pass recovery proceeds.

The implementation (this issue's direction 3 — **split-and-retry, never the JSON corrector**: it cannot recover signals the model never emitted, and on the truncation class it would freeze partial batches as reviewed, defeating absence-as-retry):

- a dropped or truncated batch of **≥2 units re-issues once as two halves** — smaller outputs (less likely to break mid-structure) and a fresh roll; bounded to **one split level, no recursion** — a still-dropped half stays dropped (its units re-run on resume via absence-as-retry);
- **the recovery carries its own provenance** (direction 4): `batches_split_recovered`/`batches_split_lost` — distinct counters, surfaced in the step summary; the original drop's counters are *revised* by the attempt's **exact applied deltas** (the adversarial review's catch: an inferred-shape subtraction drove `batches_truncated` negative on the max_tokens+prose shape);
- **the usage window spans the whole recovery** (the review's second catch: per-attempt tracking reset left the records carrying only the last half's spend — the window now starts once per original batch, the recovered units' records carrying the original + halves' true cost);
- the recovered halves' units persist checkpoint records; the still-dropped leave none. A provider-exception batch (`failed`) is deliberately **not** split — not output-size-shaped; the #541 counters + the resume own it.

Closes #558.

## Test plan

7 new tests (the #558 class): full recovery via halves (the counters revised to 0, the provenance counted); the recovered units persist; a 1-unit batch never splits; the mixed half-recovered/half-lost shape (the coverage truth exact); the max_tokens+prose no-brace shape never drives a counter negative (the deltas pin); the usage spans the recovery; plus the 3 old-behavior pins updated (a still-dropped batch needs the halves to also fail).

## Verification evidence

| check | command | result |
|---|---|---|
| new + family | `pytest tests/test_issue532_llr_resume.py tests/test_issue294_reach_batch_diagnostics.py tests/test_llm_reachability.py -q` | 74 passed |
| full suite | `pytest tests/ -q` | 4022 passed, 2 failed — both pre-existing (SDK pin drift) |
| adversarial review | combined wave+refute | 5 findings, all folded: the negative-counter catch (the deltas mechanism), the usage-window catch, the provenance surfacing, the recovery stderr line, the failed-class note |
| the measurement | the issue comment's receipts (4 malformed in ~95 at the 20000 cap, 0 max_tokens) | the decision rule's answer, posted |

### DIFF
--- a/libs/openant-core/core/llm_reachability.py
+++ b/libs/openant-core/core/llm_reachability.py
@@ -461,6 +461,10 @@ def analyze_reachability(
     units_not_reviewed = 0
     batches_truncated = 0
     batches_failed = 0
+    # #558: the split-and-retry provenance counters (direction 4 — the
+    # recovery never silently overwrites the coverage counts).
+    batches_split_recovered = 0
+    batches_split_lost = 0
 
     # ------------------------------------------------------------------
     # #532: resume/adopt machinery — the checkpoint family's own pattern
@@ -597,15 +601,22 @@ def analyze_reachability(
 
     batches = _chunk(units_to_run, batch_size)
     persisted = 0
-    for i, batch in enumerate(batches):
+
+    # #558: one call+parse attempt for a (sub-)batch. Returns
+    # (signals, outcome) where outcome ∈ {"ok", "failed", "dropped",
+    # "truncated"}; the counters are mutated here so the split-and-retry
+    # below can revise them (a recovered original subtracts what its
+    # drop counted; the halves count their own outcomes).
+    def _attempt(sub_batch, label) -> tuple:
+        """Returns (signals, outcome, deltas) — deltas are THIS attempt's
+        applied counter changes ({dropped, units, truncated}), so the
+        split-and-retry subtracts exactly what was applied (never an
+        inferred shape — the refutation's negative-counter catch)."""
+        nonlocal dropped_batches, units_not_reviewed, batches_truncated, \
+            batches_failed
         prompt = build_prompt(
-            batch, app_context=app_context, max_code_bytes=max_code_bytes
+            sub_batch, app_context=app_context, max_code_bytes=max_code_bytes
         )
-        if tracker is not None:
-            try:
-                tracker.start_unit_tracking()
-            except Exception:  # noqa: BLE001
-                pass
         try:
             result = simple_completion(binding, prompt,
                                        max_tokens=DEFAULT_MAX_TOKENS,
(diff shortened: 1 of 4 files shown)


gadievron (commented):
**Verification review — recommend-merge.** (all evidence pinned to head `e291bac`; re-verify on any new head)

**The fix**: `_attempt` returns (signals, outcome, deltas) where deltas are THIS attempt's applied counter changes; a dropped or truncated batch of ≥2 units is re-issued ONCE as two halves with the original drop's counters subtracted EXACTLY (a prior refutation drove `batches_truncated` negative when the shape was inferred — this closes that class); ONE split level, no recursion; provider-exception (`failed`) batches are deliberately not split; the recovery carries its own provenance counters (`batches_split_recovered`/`batches_split_lost`) surfaced in scanner stats; the usage-tracking window spans the original + halves so per-unit cost carries the recovery; per-unit records persist for OK units only — dropped halves re-run on resume via absence-as-retry.

**Verification** (run at the pinned head): RED was assertion-level — 6 tests failed against the base tree for the stated reasons (including `test_truncated_no_brace_never_negative`); GREEN — 44 passed in the two touched suites; full suite — 3889 passed, 168 skipped, 0 FAIL; CI at `e291bac`: success. The review round (six independent reviewers) traced every counter path — split, partial-recovery, both-halves-lost, odd sizes, the max_tokens/no-brace shape — and found no defect; the counter arithmetic was re-derived case-by-case against the tests.

**Cross-PR**: the combined tree of all six open PRs passed the full suite; the same-file merge with #572 (disjoint regions of `llm_reachability.py`) is clean; no retry stacking with #571 (LLR's `_attempt` failures are never split at a raised cap — they go to `batches_failed`).

**Known semantics (disclosed)**: `batches_dropped` now counts *attempts* (a fully-lost split = 2) while `len(batches)` remains the source-batch denominator — consumers gate on >0 only, but the counter's unit changed; a partial recovery (one OK half, one lost) counts `batches_split_recovered` — a counter-only reader over-reads coverage.

**Follow-ups (named, none blocking)**: no test drops an odd-size (≥3) batch (mid = (len+1)//2 is correct by inspection but untested); the split also fires on prose-refusal drops while the comment argues only output-size-shaped failures should split — either state the isolation rationale or gate on shape; the cheap composition with #571: have `_attempt`'s exception arm return "truncated" when it sees `is_budget_exhausted_error`, so the split covers the empty-length-stop class in LLR too.

Issue #540, opened 2026-09-07
The vendored tailwind Play-CDN blob is frozen at 3.4.17 — execute the prebuilt-CSS / v4-migration decision


gadievron:
Filed from the adversarial closure re-audit (fable + astra, both seats: all five renovate closures stand; this is the one durable follow-up they named — correcting the earlier claim that #412 covered this, which was **false**: #412 is the dynamic-test Docker runtime policy, unrelated to tailwind).

## The state

- The vendored tailwind blob (`apps/openant-cli/internal/report/vendor/tailwindcss-3.4.17.js`, 407KB Play-CDN JIT compiler) is **frozen**: cdn.tailwindcss.com serves the Play-CDN build only up to 3.4.17 (live-verified 2026-09-05 and re-verified 2026-09-08 — 3.4.18/3.4.19 return `console.error` stubs naming 3.4.17 as the last). The 3.4.18/3.4.19 GitHub releases contain **no security fixes** — nothing security-relevant is frozen out today.
- The npm tarballs for 3.4.19 (and v4) carry **no browser/Play-CDN script** — the Play-CDN JIT is a separate build published only at that CDN. #494/#495 were correctly closed as invalid on the vendored path.
- `renovate.json` caps the tailwind line at `allowedVersions <=3.4.17` (#511), making the notifier line dead by design. The cap deliberately doesn't track upstream changes or reconsider the freeze — **this issue is that tracker**.

## The decision to make (the maintainer's)

1. **Prebuilt CSS via the tailwind CLI** — the v3.4.18/3.4.19 releases ship standalone CLI binaries; a build step produces 10-30KB CSS vs the 407KB compiler-in-browser (the SOURCES.txt:17-20 recorded direction). Executable on the v3-LTS line today without committing to v4. Trade-off: a build step + a second generated artifact to pin.
2. **v4 migration** — v4's browser story is `@tailwind/css/browser` (a different artifact; the report's `Chart.register(ChartDataLabels)` usage needs re-verification against v4's API). #495 was correctly closed as an automated-notifier artifact, not as a v4 decision.
3. **Status quo** — the frozen blob keeps working (the report is self-contained and the vendor tests pin it); revisit only if a security fix lands upstream or the CDN state changes.

Acceptance criteria for any path: the vendor sha256 test, the breakout guard, and the report self-containment (#332 contract) hold; the drift-guard family (#412/#511-class tests) covers any new artifact source.

gadievron:
[...]

PR #578, opened 2026-09-11
fix(ui): vendored ui scripts under the #332 discipline — version-carrying names, the sha lock, renovate notifiers, DOMPurify 3.4.15

### LINKED ISSUE #577: The ui/vendor directory has none of the vendored-artifact governance — DOMPurify (the sole XSS sanitizer for untrusted markdown, with no CSP backstop) is frozen by omission at 3.4.13, two patch releases behind
The ui/vendor directory has none of the vendored-artifact governance — DOMPurify (the sole XSS sanitizer for untrusted markdown, with no CSP backstop) is frozen by omission at 3.4.13, two patch releases behind

Filed from the 2026-09-09 before-#540 sweep; every fact re-derived from the tree.

## The gap

`apps/openant-cli/ui/vendor/` holds two third-party browser-runtime blobs with **zero** of the #332 vendored-artifact discipline that `internal/report/vendor/` carries:

- **no provenance record** (no `SOURCES.txt` — no source URL, no dist-integrity anchor);
- **no sha lock** (no test pins the bytes; grep for `marked|purify` across `*_test.go` → 0 hits);
- **no notifier** (the filenames carry no version — `marked.min.js`, `purify.min.js` — so the `renovate.json` regex-manager pattern that covers `report/vendor/` structurally cannot match them; and `**/vendor/**` is renovate-ignored via `SOURCES.txt:7-9`'s ignore line — that line lists only the report path, so `ui/vendor/` is simply never seen).

The versions, re-derived from the license headers and npm at write time:
- **DOMPurify 3.4.13** (`ui/vendor/purify.min.js:1`) vs latest **3.4.15** — and DOMPurify patch releases are routinely XSS-bypass fixes;
- **marked v12.0.2** (`ui/vendor/marked.min.js:2`) vs latest **18.0.12** — six majors behind.

## Why this is higher-stakes than the report vendor

`ui/summary.html:52-67` and `ui/disclosure.html:54-56`: the server embeds these scripts (`ui/embed.go:6`), serves them (`server.go:619-635`), and renders **attacker-influenced LLM/repo-derived markdown** through `marked.parse` → `DOMPurify.sanitize` → `innerHTML` (`summary.html:67`). DOMPurify is the **sole XSS boundary** on that path — and `internal/server/` sets **no Content-Security-Policy header anywhere** (grep: 0 hits), so there is no defense-in-depth backstop. A published sanitizer bypass would leave this surface open until someone happens to notice, and nothing in the repo would signal the drift.

## Direction (the #332 recipe, the variant for the filename problem)

1. Rename to version-carrying filenames (`marked-18.0.12.min.js`, `purify-3.4.15.min.js`) so the existing renovate regex-manager pattern applies; update the embed/serve/consume references together;
2. Add a `ui/vendor/SOURCES.txt` (source URLs, dist-integrity anchors) + a sha-lock test (the `report_vendor_test.go:20-27` shape);
3. **Bump DOMPurify to 3.4.15 in the same commit** (the bypass-fix class: staying behind is the exposure);
4. marked's six-major gap is a re-vendor decision (the API is stable for our usage; bump when the governance lands, not as an emergency).

Related: #540 (the same class's other pending decision — whichever recipe lands there defines this one's), #332 (the vendored-artifact discipline), #502/#529 (the SRI/pin family), #412 (the drift-guard family).

gadievron:
## What

`apps/openant-cli/ui/vendor/` (the web UI's two vendored browser-runtime scripts) had none of the vendored-artifact governance that `internal/report/vendor/` carries (#332): no provenance record, no sha lock, no update notifier — and the versionless filenames (`marked.min.js`, `purify.min.js`) made the notifier structurally impossible. DOMPurify — the sole XSS sanitizer for the untrusted-markdown render path of both web-UI report pages (`marked.parse → DOMPurify.sanitize → innerHTML`, including the second sink in disclosure's repository-URL rewrite) — sat two patch releases behind.

This change:

1. renames to version-carrying filenames: `marked-12.0.2.min.js` (byte-identical; verified against the official 12.0.2 dist) and `dompurify-3.4.15.min.js` (currency bump; the tarball's sha512 re-derived and matched the registry `dist.integrity` at write time),
2. adds `ui/vendor/SOURCES.txt` (provenance: source URL, npm tarball anchor, dist integrity, sha256, verification date) — deliberately not embedded, asserted by test,
3. locks the bytes: `ui/vendor_test.go` (sha pins + versioned-name invariants + banner==filename + embed/disk set-equality) and `internal/server/assets_test.go` (serving: every embedded script 200 + byte-identity + content-type + nosniff; the versionless names 404; the handleAsset allowlist ↔ embedded-set consistency, both directions),
4. adds two renovate customManagers (marked, dompurify) so the version is visible to the notifier — static `packageNameTemplate` + value-only matchStrings per the in-file #476/#485 invariant (the reference files carry the version; `**/vendor/**` is renovate-ignored by the preset, exactly like the report side),
5. extends `.gitattributes` (the Windows autocrlf vs sha-lock hazard, same rule as the report blobs),
6. pins the sanitizer policy — per key AND per value: the MD_SANITIZE block is cross-template byte-equal, its key set is pinned to exactly the five policy keys (so no `ADD_TAGS`/`ADD_ATTR`/`ALLOWED_URI_REGEXP`-style key can reopen a channel the value pins never see), the FORBID floor (incl. `script`, and DOMPurify's forbid-beats-allow semantics verified against the vendored blob) AND the ALLOW ceiling (no dangerous tag may enter ALLOWED_TAGS; no URL-style attribute or `on*` handler may enter ALLOWED_ATTR; `ALLOW_DATA_ATTR` stays false), both sanitize sinks including the rewrite call's policy argument, and the page invariants (exactly the vendored script set, in load order, positioned before the first inline consumer; no external-origin script tag in any tag form; no module scripts, https imports, external stylesheets/@imports, or cross-origin fetches; and a total scheme-literal ban — any `http(s)://` URL in any page beyond the one scan-form placeholder fails, closing the reference by ANY channel, not just the named ones).

Versioned filenames also make bumps cache-safe: the assets are served with `Cache-Control: public, max-age=86400`, and a rename means newly rendered pages never pick a stale blob out of the 24h cache (already-open pages keep their in-memory copy until reload).

## The DOMPurify 3.4.13 → 3.4.15 bump, honestly framed

This is a currency/hygiene bump of the sole sanitizer, not a claim that a known exploit is closed: 3.4.14/3.4.15 carry no advisory ids; the bypass conditions upstream names (risky tags allow-listed; XML-content clobbering) are not met by this app's inert MD_SANITIZE allowlist; and the 2026-series advisories were fixed by 3.4.5/3.4.13 — the previously vendored version was already the fixed one (verified: the old blob is byte-identical to the official 3.4.13 dist). 3.4.15 does add removal semantics vs 3.4.13 (LITERAL_TEXT_ELEMENTS force-removal; linkage-attribute stripping) — additive and in the safer direction under this allowlist (all affected tags are already non-allowlisted); the observable delta is that KEEP_CONTENT no longer preserves raw-text-element text carrying its own end tag.

## Verification

- **RED→GREEN**: the new tests were written first and failed on pristine master (`9ce6069`) — 5 failing (the sha ReadFile on the missing versioned file; the versionless-name walk; the banner mismatch; the old HTML srcs; the old-name route 200) — all green after the change. Working tree == committed tree, re-tested at every amend.
- **Full suite**: `go test ./... -count=1` — 12/12 packages ok. `go vet ./...` clean; gofmt clean over the changed files. (Page reads in the tests are CRLF-normalized: text files convert on Windows checkouts and `go:embed` embeds the converted bytes.)
- **Guard teeth**: 23 mutations, each run RED then reverted — external script tags in double-quoted, single-quoted, unquoted-upper-case, and `defer`-variant forms; the rewrite call dropping its policy argument; tags moved after/inside the consumer; a module script; an external stylesheet link; an `@import`; a cross-origin fetch; `'script'` in ALLOWED_TAGS; an `on*` handler in ALLOWED_ATTR; `ALLOW_DATA_ATTR` flipped; a stale allowlist literal; the embedded-but-unlisted direction; a SOURCES.txt sha edit; a quoted-`>` attribute evasion; a stray file in `ui/vendor`. Tolerance receipts (stay green by design): a CRLF-converted page; an editor dotfile.
- **Serving path through the real binary**: `openant serve` → both new assets 200 with `application/javascript` + `nosniff` + the cache header, served bytes sha-identical to the embedded, test-pinned bytes; the old names 404.
- **Browser-level spot check** (a one-time runtime receipt, recorded as PR evidence — not a CI lane): the exact vendored 3.4.15 bytes + the exact template config under jsdom — benign markdown renders (table/bold/code); 7 hostile payload classes sanitized (img-onerror, script, svg-onload, `javascript:`-link, form, style, iframe).
- **Renovate extraction** (the repo's own `--dry-run=extract` methodology, CLI 37.440.7): both new deps extract at their pinned versions from all five reference files; pristine master extracts neither; the three existing managers unchanged.
- **Static analysis**: CodeQL (full go-queries suite) base-vs-fix — 0 alerts base, 0 fix, 0 new in changed files; semgrep clean over the changed surface.

## Documented limitations

- No CSP backstop exists (the sanitizer is the sole boundary; a CSP would need nonces/hashes for the templates' inline scripts/styles — a design change, out of scope here; recorded in `ui/vendor/SOURCES.txt`).
- marked 12.0.2 → 18.x is deferred to its own PR (a six-major re-vendor; no published advisory covers 12.0.2 — that is the current record, not a safety proof; the new notifier surfaces the delta; deliberately no `allowedVersions` cap).
- The upstream-tarball equality was verified at write time; the in-tree sha lock is the tamper guard, not an upstream re-verification. The textual pins (the golden policy block, the identifier counts, the external-reference bans) guard drift — mistakes, refactors, ignorance; adversarial construction by an editor with write access is the code-review control, with the jsdom receipt as the one-time runtime evidence.
- The vendored-blob renovate charset `[0-9.]+` accepts non-semver shapes — matching the existing report-side managers (house style; notifier-only).
- `moon.yml`'s test/build inputs don't include embedded assets (pre-existing, shared with the report side; version bumps touch `embed.go`, which is an input, so they invalidate — noted for a future chore).

## Coordination

`renovate.json` is also touched by the pending tailwind prebuilt-CSS work (#540, a separate branch not yet a PR); the edits are additive on both sides (different manager stanzas) and merge trivially in either order.

Closes #577.


gadievron (commented):
**Verification review — recommend-merge.** (head `c7a86e3`; all evidence pinned to it — re-verify on any new head)

**The fix**: the vendored ui scripts gain the #332 discipline — version-carrying names (`marked-12.0.2.min.js`, `dompurify-3.4.15.min.js`, retiring the versionless names that made an update notifier structurally impossible), a `vendor/SOURCES.txt` provenance record, a sha256 lock enforced by `ui/vendor_test.go`, `/assets` serving tests (`server/assets_test.go`: every embedded script served byte-identically with the JS content-type + nosniff; the versionless names 404, never aliased), two renovate customManager notifiers, `.gitattributes -text -diff` for the blobs, and DOMPurify 3.4.13 → 3.4.15 — the sole XSS sanitizer for both report pages' untrusted-markdown render path — plus `'script'` added to `MD_SANITIZE`'s `FORBID_TAGS` in both pages.

**Verification** (re-derived from raw at the pinned head):
- **Provenance chain**: the npm registry's `dist.integrity` (re-derived live) matches the downloaded tarballs' sha512 for both packages; the tarballs' inner-file sha256 equals the committed bytes, `SOURCES.txt`, and the pin table (`marked 15fabce5…`, `dompurify f263b05…`). The historical claims verified too: the base `purify.min.js` was byte-identical to official 3.4.13, and the base `marked.min.js` to official 12.0.2 — a true rename, not a re-vendor.
- **Tests**: all 9 governance tests pass (the sha locks, banner↔filename, disk↔embed set equality, the no-external-scripts guard incl. protocol-relative forms and a script-open inventory, versioned-and-ordered-before-consumer pins, `SOURCES.txt`↔pin-table agreement, the `MD_SANITIZE` golden block + key-set + identifier-count pins covering both sinks' invocations, and the `/assets` serving/404/allowlist-equality tests). Full Go suite: 12 packages, 0 FAIL.
- **Renovate extract receipt**: a mechanical simulation of both customManagers yields exactly `marked@12.0.2` and `dompurify@3.4.15`, from the fileMatch'd reference files only (value-only matchStrings + static package templates — the #485 invariant).
- **Advisory triage**: every published DOMPurify advisory is patched by ≤3.4.13 — none affects 3.4.15. 3.4.14's named bypass preconditions (risky tags allow-listed) are not met by this app's inert allowlist; 3.4.15 is the XML-clobbering hardening.
- **jsdom render probe** (the actual vendored bytes + the pages' verbatim policy): 16/16 — script/img-onerror/`javascript:`/`data:`/svg/math/style/iframe/form/input vectors, data-* attributes, and clobbering ids all stripped; benign markdown shapes preserved.
- semgrep on the changed Go files: 0 findings. CI at the head: tests + gitleaks success.

**Corrections to the review record** (from this review wave):
- The disclosure repository-URL rewrite is **not** dead-never — it is **misfire-only**: marked converts the common `**Repository:**` form to `<strong>` before the regex runs, but code-span content (`` `**Repository:** https://x` ``) preserves the literal asterisks and the regex fires there. The consequence is benign (an allowlisted anchor, re-sanitized with the same policy; the capture excludes quotes/angles/spaces) — no capability the markdown link syntax doesn't already grant. Any future cleanup must treat it as behavior-changing and move the pinned rewrite assertions (`vendor_test.go`'s `MD_SANITIZE`/DOMPurify counts + the rewrite-tail pin) in the same commit.
- `SOURCES.txt` now records the sha lock's blind spot (added in `c7a86e3`): the lock detects uncoordinated drift only — a coordinated blob+sha+pin swap passes every test; the guard is the reviewer's out-of-band cross-check on every bump (tarball sha512 vs registry integrity, inner-file sha256 vs committed bytes). The module-script tag guard was also made case/spacing-insensitive.

**Follow-ups (named, none blocking)**: no CSP on the server (a recorded design change — DOMPurify is the single XSS boundary; a CSP would need nonces for the inline scripts); the marked 12→18.x re-vendor (no cap by decision; the 18.x OOM advisory range does not cover 12.0.2; expect a build-red-by-design renovate branch — that is the notifier working); the rewrite misfire cleanup; the trailing `sourceMappingURL` 404 (cosmetic, kept for byte fidelity).

PR #580, opened 2026-09-12
fix(ui): the disclosure repository-URL rewrite removed — dead on its own page, the single-sink invariant pinned (#578 follow-up C)

gadievron:
## What

The #578 review round named the disclosure repository-URL rewrite as dead code. This removes it and pins the single-sink invariant.

The rewrite regexed the literal `**Repository:**` against post-marked innerHTML — but that line is the **summary** prompt's (`report/prompts/summary.txt:10`); the disclosure generator never emits it. On the summary page marked renders it as `<strong>Repository:</strong>` + a native GFM autolink before the regex ever ran. Its only live behavior was a misfire: inside code spans it injected an allowlisted anchor into inert `<code>` content. The second `DOMPurify.sanitize` guarded only the rewrite's own mutation — dead weight over the primary sink.

## Changes

- `ui/disclosure.html`: the rewrite block removed; the primary render is the sole `innerHTML` sink per page.
- `ui/vendor_test.go`: the disclosure `MD_SANITIZE` count 3→2 and `DOMPurify` code-count 2→1; both rewrite pins deleted; a **new single-sink pin** — `innerHTML` appears exactly once per markdown page, so the second-sink class cannot return unnoticed.

## Behavior delta (disclosed)

Repo links now open same-tab — the rewrite's `target="_blank" rel="noopener"` is gone; marked's native autolink emits a plain `<a href>`. Native autolink was already the live behavior for the common form; nothing user-visible changes except the misfire stops.

Depends-on: #578 (stacked). Verification: `go test ./ui/ ./internal/server/` green at the pinned head; jsdom post-removal probe — the common form still clickable via native autolink, code spans stay literal, explicit links unchanged.

### DIFF
--- a/apps/openant-cli/ui/disclosure.html
+++ b/apps/openant-cli/ui/disclosure.html
@@ -71,17 +71,6 @@ const MD_SANITIZE = {
 };
 const markdown = {{.MarkdownJSON}};
 document.getElementById('content').innerHTML = DOMPurify.sanitize(marked.parse(markdown), MD_SANITIZE);
-
-// Make repository URLs in the rendered output clickable links. The result is
-// re-sanitized with the same inert allowlist: rewriting innerHTML can reintroduce
-// markup, and the URL class excludes quotes/angle-brackets so a crafted URL
-// cannot break out of the href attribute.
-document.querySelectorAll('#content p, #content li').forEach(el => {
-  el.innerHTML = DOMPurify.sanitize(el.innerHTML.replace(
-    /(\*\*Repository:\*\*\s*)(https?:\/\/[^\s<>"']+)/g,
-    '$1<a href="$2" target="_blank" rel="noopener">$2</a>'
-  ), MD_SANITIZE);
-});
 </script>
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


gadievron (commented):

**Independently verified at the pinned head**: the rewrite's target line is the *summary* prompt's (`report/prompts/summary.txt:10`) — the disclosure page never emits it, so the block never matched on the page it ran on; marked's GFM autolink (a default asserted from both vendored blobs' bytes) already makes the common form clickable. The removal's only live-behavior loss was the code-span misfire (an allowlisted anchor injected into inert `<code>` content — no capability markdown links don't already grant).

The same-commit pin moves all hold: disclosure `MD_SANITIZE` 3→2, `DOMPurify` code-count 2→1, both rewrite pins deleted, the second-sink comments replaced — and the new single-sink pin (`innerHTML == 1` per markdown page) prevents the class from returning unnoticed. Six reviewers, zero Critical/Significant findings.

Behavior delta (disclosed in the body): repo links open same-tab now.

The stack: this PR sits on #578 — merge that first, then this, then #581, then #582.

PR #581, opened 2026-09-12
fix(ui): the marked re-vendor 12.0.2 -> 18.0.12 — the full #332 recipe, the advisory cleared (#578 follow-up B)

gadievron:
## What

The marked re-vendor the `SOURCES.txt` record deferred: 12.0.2 → **18.0.12**, the full #332 recipe in one commit.

## Changes

The vendored artifact is the npm tarball's `lib/marked.umd.js` — **the 18.x tarball ships no root `.min.js`** (removed in v16.0.0), and jsdelivr's synthesized `.umd.min.js` was deliberately NOT vendored (it strips the version banner and is not a registry artifact). The `-<semver>.min.js` name convention is kept (the version-suffix + renovate conventions require it; the file IS upstream-minified); the path mapping is recorded in `SOURCES.txt`.

The one-commit recipe: blob + `SOURCES.txt` (marked section rewritten, the 12.0.2 sha256 deleted per the stale-record pin) + the pin table + the `go:embed` list + the `handleAsset` cases + both pages' script tags + `marked-12.0.2.min.js` added to the retired-name 404 test.

## Provenance (re-derived at write time)

Registry `dist.integrity` == the downloaded tarball's sha512; the inner-file sha256 == the committed bytes. Advisory gate: CVE-2026-41680 / GHSA-6v9c-7cg6-27q7 (OOM; reviewed range [18.0.0, 18.0.2)) — **18.0.12 clears it**. The 12.0.2 sha-lock paragraph that deferred this decision is replaced.

## Verification (out-of-band, recorded in SOURCES.txt)

The 12→18 render differential, two oracles (raw `marked.parse` AND post-DOMPurify output), 23 corpus items including a real scan summary: **20/23 identical post-sanitize; 3 named deltas** — the v15 entity-escaping change (an `&amp;` inside a markdown URL now renders as the literal sequence; cosmetic, no new capability) and two v18 trailing-whitespace trims. Zero new tag/attribute shapes in the emission envelope. The defaults objects are identical (gfm/breaks/pedantic/async asserted from both blobs' bytes); `marked.parse` returns a string (sync) on both. The 16-vector sanitize battery at 18: **16/16** (the quote-breakout shape percent-encoded inertly — identical to 12's behavior).

Depends-on: the rewrite-cleanup PR (stacked below). Evidence: `go test ./ui/ ./internal/server/` green at the pinned head — the five-way rename net re-derived at 18.0.12.


gadievron (commented):

**Provenance re-derived from raw** (per `SOURCES.txt`'s own reviewer-obligation note): the npm registry's `dist.integrity` matches the downloaded tarball's sha512; the inner-file sha256 matches the committed bytes; the vendored artifact is the tarball's `lib/marked.umd.js` (the 18.x tarball ships no root `.min.js` since v16 — the blob carries the `marked v18.0.12` banner; jsdelivr's synthesized min was correctly avoided). **Advisory gate**: CVE-2026-41680 / GHSA-6v9c-7cg6-27q7 (OOM; [18.0.0, 18.0.2)) — 18.0.12 clears it. One reviewer independently widened the eval/`new Function`/`Function(` grep beyond the original check — 0 shapes.

**The differential** (two oracles, 23 items incl. a real scan summary): 20/23 identical post-sanitize; 3 named deltas — the v15 entity-escaping change (an `&amp;` inside a markdown URL; cosmetic) and two v18 trailing-whitespace trims. Zero new tag/attribute shapes in the emission envelope; the defaults identical on both blobs; sync string return on both; the 16-vector battery at 18: 16/16.

**The recipe**: one commit — blob + SOURCES.txt (the 12.0.2 sha deleted per the stale-record pin) + pin table + embed + handleAsset + both pages' tags + the retired-name 404. Six reviewers, zero Critical/Significant.

Named follow-up: a golden render fixture as the next bump's in-repo oracle (the differential is out-of-band today, per the recorded reviewer-obligation note).

The stack: merge #578 → #580 → this → #582.

PR #582, opened 2026-09-12
feat(ui): route-scoped Content-Security-Policy — per-response nonces on the template pages, the no-network policy on /report (#578 follow-up A)

gadievron:
## What

The CSP gap the `SOURCES.txt` record carried since #577 closes — route-scoped, per-response nonces.

## The design

- **The four UI template pages** (index, scan, summary, disclosure): a per-response 32-hex nonce (crypto/rand — hex avoids html/template's `+`-escaping of base64), threaded middleware → context → data struct → `{{.Nonce}}` on every inline `<style>` and `<script>` open. Policy: `default-src 'none'; script-src 'self' 'nonce-N'; style-src 'self' 'nonce-N'; style-src-attr 'unsafe-inline'; connect-src 'self'; form-action 'self'; base-uri 'none'; object-src 'none'; frame-ancestors 'none'`. `Cache-Control: no-store` (a nonce is single-use). The index page's two inline `onclick` handlers (blocked by nonce script-src) are refactored to a delegated `addEventListener` on `#scans-list` with `data-id`/`data-repo` attributes.
- **`/report/{id}`**: a pre-generated static file whose inline scripts/styles cannot carry a serve-time nonce — it gets the **no-network policy** instead: `script-src/style-src 'unsafe-inline'` (the page is the trusted origin) but `connect-src 'none'; form-action 'none'` — exfil/remote loads blocked from a page rendering LLM-authored remediation HTML. Checked at write time: the report's three vendored blobs contain **zero** eval/`new Function` shapes (grep receipts in `csp_test.go`), so no `'unsafe-eval'` is needed.
- **`/assets`**: no CSP (not a document).
- **`style-src-attr 'unsafe-inline'`**: the static `style=` attributes (index/scan/disclosure) — contained: `MD_SANITIZE` forbids `style` on the markdown path; the fail-safe direction on older browsers is cosmetic blocking.

## Honest scope

DOMPurify stays the **primary** content boundary; the CSP converts the worst residual — script execution via a sanitizer bypass — into markup-level abuse (allowlisted-tag defacement / same-origin phishing), which no header can fix because the page IS the trusted origin. `SOURCES.txt` and `vendor_test.go`'s provenance records are amended to this sanitizer model.

## Verification

`TestCSPPoliciesPerRoute`: the per-route policies (the nonce-policy shape + **header/body nonce equality** on `/` and `/scan/{id}`; rotation across responses; `no-store`; the no-network `/report` policy nonce-free; `/assets` CSP-free with nosniff intact; no inline event handlers). `vendor_test.go`'s script-inventory made nonce-aware — zero bare `<script>` opens enforced. The summary/disclosure route classification pinned on their 404 paths (the middleware sets CSP pre-handler). Known limitation (recorded): jsdom does not enforce CSP — the header/nonce/policy shape is httptest-proven; a real-browser run is the residual gap.

Depends-on: the marked-18 PR (stacked below). Evidence: `go test ./...` — 12 packages, 0 FAIL; semgrep (p/golang) on the changed files — 0 findings.

**Follow-ups (named)**: the `isUIPage` route-table duplication (a per-route-wrapper at registration would fuse route+policy); the `MarkdownJSON` `</script>`-breakout fixture test (json.Marshal HTML-escapes today — pin it); a golden render fixture as the next marked bump's in-repo oracle; `target`/`rel` in `MD_SANITIZE`'s allowlist have no producer post-cleanup.


gadievron (commented):

**The CSP contract, verified at the pinned head**: the four template pages carry per-response 32-hex nonces (crypto/rand; hex chosen so html/template's attribute escaper cannot desync the header/body comparison), threaded middleware → context → data struct → `{{.Nonce}}` on every inline `<style>`/`<script>` open, with `no-store`. `/report/{id}` carries the no-network policy (nonce-free by design — a pre-generated static file); `/assets` carries none. The report's three vendored blobs were checked for eval/`new Function`/bare-`Function(`/`window.eval` shapes by two independent reviewers: **zero matches** — no `'unsafe-eval'` needed, and a dedicated subtest pins the policy shape the absence justifies.

**The review round's folds** (all confirmed discharged by the two request-changes authors): the stale "no CSP anywhere" provenance records (`SOURCES.txt` + `vendor_test.go`) rewritten to the current sanitizer model — DOMPurify stays the **primary** content boundary, the CSP is the execution-class backstop; the CSP tests extended to assert header/body nonce equality + no-inline-handlers on `/` and `/scan/{id}` (both render 200; the test parses those two templates — the summary/disclosure *route classification* is pinned on their 404 paths since the middleware sets CSP pre-handler); the dead `<script>` anchor replaced by `<script nonce=`.

**Verification**: `go test ./...` — 12 packages, 0 FAIL; semgrep — 0; six reviewers → request-changes (provenance + eval risk) → both disproven/folded → confirmed discharged.

**Known limitation (recorded)**: jsdom does not enforce CSP — the header/nonce/policy shape is httptest-proven; a real-browser run (all four pages + `/report` under headless Chromium, console checked) is the residual gap.

**Follow-ups (named)**: the `isUIPage` route-table duplication (a per-route wrapper at registration would fuse route+policy — a future HTML route missing from `isUIPage` ships CSP-less, a visible gap the tests now expose); the `MarkdownJSON` `</script>`-breakout fixture (json.Marshal HTML-escapes today — pin it with a test); the golden render fixture; `target`/`rel` in `MD_SANITIZE`'s allowlist have no producer post-#580.

The stack: merge #578 → #580 → #581 → this.

